PR #34, opened 2020-10-13
Handle GIL Issue by measuring non-uniform time intervals

kngoyal:
This PR handles 
- GIL issue by recording the last measured time when the Background Scheduler called the measure function in place of using a constant value.
- Initialize gpu_ids to a default null value in the initializer.

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -49,7 +49,7 @@ class BaseEmissionsTracker(ABC):
         self._project_name: str = project_name
         self._measure_power_secs: int = measure_power_secs
         self._start_time: Optional[float] = None
-        self._last_measured_time: Optional[float] = None
+        self._last_measured_time: float = time.time()
         self._output_dir: str = output_dir
         self._total_energy: Energy = Energy.from_energy(kwh=0)
         self._scheduler = BackgroundScheduler()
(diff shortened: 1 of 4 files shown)


codecarbon/emissions_tracker.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         self._project_name: str = project_name
         self._measure_power_secs: int = measure_power_secs
         self._start_time: Optional[float] = None
+        self._last_measured_time: Optional[float] = None
```
Lothiraldan, line 52:
```suggestion
        self._last_measured_time: float = time.time()
```

See below for the reason behind this code change.

codecarbon/emissions_tracker.py
```diff
@@ -164,8 +165,9 @@ def _measure_power(self) -> None:
         """
         self._total_energy += Energy.from_power_and_time(
             power=self._hardware.total_power,
-            time=Time.from_seconds(self._measure_power_secs),
+            time=Time.from_seconds(time.time()-self._last_measured_time),
```
Lothiraldan, line 168: Mypy is complaining that `self._last_measured_time` could be None as its type hints is `Optional[float]`. Even if it's unlikely that `_measure_power` would be called before `start` is called, if you want to play the safe card, check out the code suggestion here: https://github.com/mlco2/code-carbon/pull/34/files#diff-99ecda6f3fccf943e43b5939305579d7cef3734928186e5395b14c0eda681307R52

Lothiraldan (approved):
LGTM

vict0rsch (approved):

PR #38, opened 2020-11-02
Country name is optional

adijo:


codecarbon/emissions_tracker.py
```diff
@@ -182,7 +182,7 @@ class OfflineEmissionsTracker(BaseEmissionsTracker):
     def __init__(
         self,
         country_iso_code: str,
-        country_name: str,
+        country_name: str = None,
```
kngoyal, line 185: country_name: Optional[str] = None,

kngoyal (approved):

PR #61, opened 2020-11-13
Remove old co2-tracker-utils dependency

Lothiraldan:


meta.yaml
```diff
@@ -19,7 +19,6 @@ requirements:
   run:
     - python>3.6
     - apscheduler
-    - co2-tracker-utils
```
kngoyal, line 22: remove from setup.py too ?

Lothiraldan: Good catch!

setup.py
```diff
@@ -5,14 +5,14 @@
...
     "APScheduler",
-    "co2-tracker-utils",
     "dash",
     "dash_bootstrap_components",
     "dataclasses",
     "fire",
     "pandas",
     "pynvml",
     "requests",
+    "typing>=3.7.4; python_version<'3.5'",
```
kngoyal, line 15: why python version<3.5 ?
In the docs, the recommended version is 3.6

Lothiraldan: Ok good catch, let's remove it

kngoyal (approved):

PR #63, opened 2020-11-16
Intel CPU Power Tracking via Power Gadget on Mac/Windows

kngoyal:
Tried following options from online literature to fetch CPU Power usage:
- PyJoules - unable to pip install
- PyRAPL - unable to pip install
- reading RAPL files at `/sys/class/powercap/intel-rapl/*` on Linux, however, unable to find these files on Azure and AWS VMs running Intel Processors.

Installed [Intel Power Gadget](https://software.intel.com/content/www/us/en/develop/articles/intel-power-gadget.html) on Mac. Generates current CPU power consumption across GT, IA, and DRAM and net Processor details via command line and is logged into a file, from which the `Processor Power_{i}(Watt)` values are fetched, to serve as current total Power drawn in Watts.


codecarbon/core/cpu.py
```diff
@@ -0,0 +1,103 @@
...
+            raise Exception("Platform not supported by Intel Power Gadget")
+
+    def _log_values(self):
+        """
+        Logs output from Intel Power Gadget command line to a file
+        """
+        if self._system.startswith("linux"):
+            os.system(
+                f"{self._cli} -d {self._duration} -e {self._resolution} > {self._log_file_path}"
+            )
```
Lothiraldan, line 71: os.system is not recommended for new usage, we should likely use https://docs.python.org/3/library/subprocess.html#subprocess.call instead

codecarbon/core/cpu.py
```diff
@@ -0,0 +1,103 @@
...
+        try:
+            cpu_data = pd.read_csv(self._log_file_path).dropna()
+            for col_name in cpu_data.columns:
+                if col_name in ["System Time", "Elapsed Time (sec)", "RDTSC"]:
+                    continue
+                if "Cumulative" in col_name:
+                    cpu_details[col_name] = cpu_data[col_name].iloc[-1]
+                else:
+                    cpu_details[col_name] = cpu_data[col_name].mean()
+        except FileNotFoundError:
```
Lothiraldan, line 98: This might fails also if the file is empty or malformed.

codecarbon/external/hardware.py
```diff
@@ -67,6 +67,23 @@ def from_utils(cls, gpu_ids: Optional[List] = None) -> "GPU":
 
 @dataclass
 class CPU(BaseHardware):
-    @property
+    def __init__(self, output_dir: str):
+        self._output_dir = output_dir
```
Lothiraldan, line 71: This attribute doesn't seemed used, is that expected?

codecarbon/core/cpu.py
```diff
@@ -0,0 +1,103 @@
+import os
```
Lothiraldan, line 1: Could we add a small module docstring?

codecarbon/core/cpu.py
```diff
@@ -0,0 +1,103 @@
...
+        self._system = sys.platform.lower()
+        self._duration = duration
+        self._resolution = resolution
+        self._cli = None
+
+        if self._system.startswith("linux"):
+            if shutil.which(IntelPowerGadget._linux_exec):
+                self._cli = IntelPowerGadget._linux_exec
+            else:
+                raise Exception(
```
Lothiraldan, line 42: It might be better to raise a more specialized exception, maybe `FileNotFoundError` in case it cannot find the command line executable

Lothiraldan (changes requested):
How do you install power_gadget on Linux? I want to test.

Could you also paste the install errors for PyJoules and PyRAPL, they should be straightforward to install.

codecarbon/emissions_tracker.py
```diff
@@ -55,12 +55,18 @@ def __init__(
...
-        self._is_gpu_available = is_gpu_details_available()
-        self._hardware = GPU.from_utils(
-            gpu_ids
-        )  # TODO: Change once CPU support is available
+        self._hardware = []
+
+        if gpu.is_gpu_details_available():
+            logger.debug("Tracking Nvidia GPU")
+            self._hardware.append(GPU.from_utils(gpu_ids))
+        elif cpu.is_powergadget_available():
```
vict0rsch, line 63: Shouldn't this be a new `if` instead of an `elif` so that we can track both GPU and CPU?

vict0rsch (changes requested):

vict0rsch:
Thanks a lot for this PR @kngoyal we're almost there!

codecarbon/emissions_tracker.py
kngoyal: good catch !

codecarbon/core/cpu.py
kngoyal: catching general exception, in case it is unable to fetch proper details. It will report no emissions, as before.

codecarbon/external/hardware.py
kngoyal: should be used in the subsequent object attribute, which was using the default value before.

codecarbon/core/cpu.py
kngoyal: done

codecarbon/emissions_tracker.py
kngoyal: done

kngoyal:
> How do you install power_gadget on Linux? I want to test.
> 
> Could you also paste the install errors for PyJoules and PyRAPL, they should be straightforward to install.

From the official website, it is only available for MacOS and Windows.

vict0rsch:
So the alternative is to use code from energy-usage. I can do that this afternoon @kngoyal

codecarbon/core/cpu.py
```diff
@@ -0,0 +1,99 @@
...
+                    continue
+                if "Cumulative" in col_name:
+                    cpu_details[col_name] = cpu_data[col_name].iloc[-1]
+                else:
+                    cpu_details[col_name] = cpu_data[col_name].mean()
+        except Exception as e:
+            logger.debug(
+                f"CODECARBON Unable to read Intel Power Gadget logged file at {self._log_file_path}\n \
+                Exception occurred {e}"
+            )
```
Lothiraldan, line 97: Let's log the traceback as well:
```suggestion
            logger.debug(
                f"CODECARBON Unable to read Intel Power Gadget logged file at {self._log_file_path}\n \
                Exception occurred {e}", exc_info=True
            )
```

codecarbon/core/cpu.py
```diff
@@ -0,0 +1,99 @@
...
+def is_powergadget_available():
+    try:
+        IntelPowerGadget()
+        return True
+    except Exception as e:
+        logger.debug(
+            f"Exception occurred while instantiating IntelPowerGadget : {e}",
+            exc_info=True,
+        )
+        print(e)
```
Lothiraldan, line 26: Let's not print the exception, the logging should do. We do not want the linux users to get this every time..
```suggestion
```

codecarbon/external/hardware.py
```diff
@@ -67,6 +67,23 @@ def from_utils(cls, gpu_ids: Optional[List] = None) -> "GPU":
...
+        Get CPU power from Intel Power Gadget
+        :return: power in kW
+        """
+        all_cpu_details: Dict = self._intel_power_gadget.get_cpu_details()
+
+        power = 0
+        for metric, value in all_cpu_details.items():
+            if re.match("^Processor Power_\d+\(Watt\)$", metric):
+                power += value
+        logger.info(f"CODECARBON CPU Power Consumption : {power}")
```
Lothiraldan, line 85: I'm guessing that this message will be logged everytime the energy is measure, right? If that's the case, we likely don't want the log message to be printed by default.

```suggestion
        logger.debug(f"CODECARBON CPU Power Consumption : {power}")
```

vict0rsch:
```
import os
import re
import time
from pathlib import Path

BASE = "/sys/class/powercap/"


class RAPLFile:
    def __init__(self, name, path):
        self.name = name
        self.path = path
        self.recent = 0

    def set_recent(val):
        self.recent = val

    def __repr__(self):
        return "RAPLFile({}, {}, {})".format(self.name, self.path, self.recent)
        
    
def reformat(name, multiple_cpus):
    """ Renames the RAPL files for better readability/understanding """
    if 'package' in name:
        if multiple_cpus:
            name = "CPU" + name[-1] # renaming it to CPU-x
        else:
            name = "Package"
    if name == 'core':
        name = "CPU"
    elif name == 'uncore':
        name = "GPU"
    elif name == 'dram':
        name = name.upper()
    return name

def get_files():
    """ Gets all the RAPL files with their names on the machine
        Returns:
            filenames (list): list of RAPLFiles
    """
    # Removing the intel-rapl folder that has no info
    if not Path(BASE).exists():
        return [], False
    files = list(filter(lambda x: ':' in x, os.listdir(BASE)))
    names = {}
    cpu_count = 0
    multiple_cpus = False
    for file in files:
        if (re.fullmatch("intel-rapl:.", file)):
            cpu_count += 1

    if cpu_count > 1:
        multiple_cpus = True

    for file in files:
        path = BASE + '/' + file + '/name'
        with open(path) as f:
            name = f.read()[:-1]
            renamed = reformat(name, multiple_cpus)
        names[renamed] = BASE + file + '/energy_uj'

    filenames = []
    for name, path in names.items():
        name = RAPLFile(name, path)
        filenames.append(name)

    return filenames, multiple_cpus

def read(file):
    """ Opens file and reads energy measurement """
    if file == "":
        return 0
    with open(file, 'r') as f:
        return to_joules(int(f.read()))


def to_joules(ujoules):
    """ Converts from microjoules to joules """
    return ujoules*10**(-6)


def start(raplfile):
    measurement = read(raplfile.path)
    raplfile.recent = measurement
    return raplfile

def end(raplfile, delay):
    measurement = read(raplfile.path)
    raplfile.recent = (measurement - raplfile.recent) / delay
    return raplfile

def joules_to_kwh(j):
    return j * 2.7778e-7

def measure_files(files, delay = 1):
    """ Measures the energy output of all packages which should give total power usage
    Parameters:
        files (list): list of RAPLFiles
        delay (int): RAPL file reading rate in ms
    Returns:
        files (list): list of RAPLfiles with updated measurements
    """

    files = list(map(start, files))
    time.sleep(delay)
    files = list(map(lambda x: end(x, delay), files)) # need lambda to pass in delay
    return sum(joules_to_kwh(f.recent) for f in files)

def compute_cpu_power(delay=1):
    rapl_files, _ = get_files()
    if not rapl_files:
        return -1
    return measure_files(rapl_files, delay=delay)
```

`compute_cpu_power(delay=1)` will return the cpu power measured on all available cpus through the RAPL interface. Returns -1 if RAPL not available

Is that fine to integrate @kngoyal ?

Adapted from `energy-usage.utils`

codecarbon/external/hardware.py
kngoyal: yes, that is essentially the goal

kngoyal:
> So the alternative is to use code from energy-usage. I can do that this afternoon @kngoyal

Thanks @vict0rsch

kngoyal:
> ```python
> import os
> import re
> import time
> from pathlib import Path
> 
> BASE = "/sys/class/powercap/"
> 
> 
> class RAPLFile:
>     def __init__(self, name, path):
>         self.name = name
>         self.path = path
>         self.recent = 0
> 
>     def set_recent(val):
>         self.recent = val
> 
>     def __repr__(self):
>         return "RAPLFile({}, {}, {})".format(self.name, self.path, self.recent)
>         
>     
> def reformat(name, multiple_cpus):
>     """ Renames the RAPL files for better readability/understanding """
>     if 'package' in name:
>         if multiple_cpus:
>             name = "CPU" + name[-1] # renaming it to CPU-x
>         else:
>             name = "Package"
>     if name == 'core':
>         name = "CPU"
>     elif name == 'uncore':
>         name = "GPU"
>     elif name == 'dram':
>         name = name.upper()
>     return name
> 
> def get_files():
>     """ Gets all the RAPL files with their names on the machine
>         Returns:
>             filenames (list): list of RAPLFiles
>     """
>     # Removing the intel-rapl folder that has no info
>     if not Path(BASE).exists():
>         return [], False
>     files = list(filter(lambda x: ':' in x, os.listdir(BASE)))
>     names = {}
>     cpu_count = 0
>     multiple_cpus = False
>     for file in files:
>         if (re.fullmatch("intel-rapl:.", file)):
>             cpu_count += 1
> 
>     if cpu_count > 1:
>         multiple_cpus = True
> 
>     for file in files:
>         path = BASE + '/' + file + '/name'
>         with open(path) as f:
>             name = f.read()[:-1]
>             renamed = reformat(name, multiple_cpus)
>         names[renamed] = BASE + file + '/energy_uj'
> 
>     filenames = []
>     for name, path in names.items():
>         name = RAPLFile(name, path)
>         filenames.append(name)
> 
>     return filenames, multiple_cpus
> 
> def read(file):
>     """ Opens file and reads energy measurement """
>     if file == "":
>         return 0
>     with open(file, 'r') as f:
>         return to_joules(int(f.read()))
> 
> 
> def to_joules(ujoules):
>     """ Converts from microjoules to joules """
>     return ujoules*10**(-6)
> 
> 
> def start(raplfile):
>     measurement = read(raplfile.path)
>     raplfile.recent = measurement
>     return raplfile
> 
> def end(raplfile, delay):
>     measurement = read(raplfile.path)
>     raplfile.recent = (measurement - raplfile.recent) / delay
>     return raplfile
> 
> def joules_to_kwh(j):
> 	"""
> 	Converts joules to kwh
> 	"""
>     return j * 2.7778e-7
> 
> def measure_files(files, delay = 1):
>     """ Measures the energy output of all packages which should give total power usage
>     Parameters:
>         files (list): list of RAPLFiles
>         delay (int): RAPL file reading rate in ms
>     Returns:
>         files (list): list of RAPLfiles with updated measurements
>     """
> 
>     files = list(map(start, files))
>     time.sleep(delay)
>     files = list(map(lambda x: end(x, delay), files)) # need lambda to pass in delay
>     return sum(joules_to_kwh(f.recent) for f in files)
> 
> def compute_cpu_power(delay=1):
>     rapl_files, _ = get_files()
>     if not rapl_files:
>         return -1
>     return measure_files(rapl_files, delay=delay)
> ```
> 
> `compute_cpu_power(delay=1)` will return the cpu power measured on all available cpus through the RAPL interface. Returns -1 if RAPL not available. Unit is kWh.
> 
> If files are computed beforehand in the `__init__` then `measure_files` should be used instead
> 
> Is that fine to integrate @kngoyal ?
> 
> Adapted from `energy-usage.utils`

This should be done in another PR, as this is a disjoint module for RAPL interface.

vict0rsch:
I agree I just thought you'd know better than myself where to put it :)

README.md
```diff
@@ -60,6 +63,17 @@ conda install -c codecarbon -c conda-forge codecarbon
 `codecarbon` is now installed in your the local environment
 
 
+# Infrastructure Support
+Currently the package supports following hardware infrastructure.
+
+### GPU
+- Tracks Nvidia GPUs power consumption using pynvml library.
```
Lothiraldan, line 70: Might be worth adding that no additional installation is required from the user, as pynvml is tracked as a package dependency.

Lothiraldan:
I've tested this PR on Windows and I've identified several issues:
* The PowerLog executable is named `PowerLog3.0.exe`.
* It is not in the PATH by default.
* If the user don't have permission to write the csv file in the given directory, PowerLog don't show an error or return with a different return code. I'm not sure if we check that the user have write permission to the output dir.

After fixing those issues locally, it seems to works as it should.

kngoyal:
> I've tested this PR on Windows and I've identified several issues:
> 
> * The PowerLog executable is named `PowerLog3.0.exe`.
> * It is not in the PATH by default.
> * If the user don't have permission to write the csv file in the given directory, PowerLog don't show an error or return with a different return code. I'm not sure if we check that the user have write permission to the output dir.
> 
> After fixing those issues locally, it seems to works as it should.

Thanks @Lothiraldan ! Good to know that it works on Windows as well.
In terms of local changes, if there are code changes, can you please suggest them or push those changes.

- As you described, I believe this would change, `_windows_exec = "PowerLog.exe"` to `_windows_exec = "PowerLog3.0.exe"`.

- What is the default path it resides at? We can add it to the documentation, to set the PATH accordingly.

- We can add to the documentation that a user specifies the `output_dir`  to the one they have write permissions to.

Let me know if you have additional suggestions.

Lothiraldan:
> > I've tested this PR on Windows and I've identified several issues:
> > 
> > * The PowerLog executable is named `PowerLog3.0.exe`.
> > * It is not in the PATH by default.
> > * If the user don't have permission to write the csv file in the given directory, PowerLog don't show an error or return with a different return code. I'm not sure if we check that the user have write permission to the output dir.
> > 
> > After fixing those issues locally, it seems to works as it should.
> 
> Thanks @Lothiraldan ! Good to know that it works on Windows as well.
> In terms of local changes, if there are code changes, can you please suggest them or push those changes.
> 
>     * As you described, I believe this would change, `_windows_exec = "PowerLog.exe"` to `_windows_exec = "PowerLog3.0.exe"`.

Yes we can do that for now, but not sure if the executable name will stay the same in a next release for Window. I think it's ok for now.
> 
>     * What is the default path it resides at? We can add it to the documentation, to set the PATH accordingly.

I will paste the default path and I think we can add it as a backup check (like what we do on Mac). Users can change where it's installed tough.
> 
>     * We can add to the documentation that a user specifies the `output_dir`  to the one they have write permissions to.

If we don't have write permissions to the `output_dir` we won't be able to write the emission.csv, right? I think we can check if we have write access (by creating an empty file `emission.csv` for example) and fallback on a temporary directory instead.
> 
> 
> Let me know if you have additional suggestions.

adijo (approved):

Lothiraldan:
I updated the PR but I sill have to test it on Windows. @kngoyal I think the backup fallback on OS X is flawed, could you check it? I don't have access to a Mac OS machine.

kngoyal:
> I updated the PR but I sill have to test it on Windows. @kngoyal I think the backup fallback on OS X is flawed, could you check it? I don't have access to a Mac OS machine.

Looks about right on my machine.

```
shutil.which("/Applications/Intel Power Gadget/PowerLog")
'/Applications/Intel Power Gadget/PowerLog'
```

vict0rsch (approved):

Lothiraldan:
> > I updated the PR but I sill have to test it on Windows. @kngoyal I think the backup fallback on OS X is flawed, could you check it? I don't have access to a Mac OS machine.
> 
> Looks about right on my machine.
> 
> ```
> shutil.which("/Applications/Intel Power Gadget/PowerLog")
> '/Applications/Intel Power Gadget/PowerLog'
> ```

OK I think I know why I was confused, is `/Applications/Intel Power Gadget/PowerLog` the full-path to the OSX executable or its folder? For Windows, I've added the folder as a backup: `_windows_exec_backup = "C:\\Program Files\\Intel\\Power Gadget 3.5\\"`

kngoyal:
> > > I updated the PR but I sill have to test it on Windows. @kngoyal I think the backup fallback on OS X is flawed, could you check it? I don't have access to a Mac OS machine.
> > 
> > 
> > Looks about right on my machine.
> > ```
> > shutil.which("/Applications/Intel Power Gadget/PowerLog")
> > '/Applications/Intel Power Gadget/PowerLog'
> > ```
> 
> OK I think I know why I was confused, is `/Applications/Intel Power Gadget/PowerLog` the full-path to the OSX executable or its folder? For Windows, I've added the folder as a backup: `_windows_exec_backup = "C:\\Program Files\\Intel\\Power Gadget 3.5\\"`

Yeah it's the executable path.

Lothiraldan (approved):
LGTM

Issue #90, opened 2020-12-03
Code fails silently when no hardware is detected


acatovic:
* CodeCarbon version: 1.0.0
* Python version: 3.7.5
* Operating System: WSL (used under MS Windows 10)

### Description

It seems when used under WSL, Intel Power Gadget and RAPL are not available. So you run the start()/stop() code and everything seems fine, but the emissions file is never written. There are probably lot of WSL users running jupyter notebooks and other DS/ML code under Windows in this way. It would be good to actually check that hardware is present.

It can easily be fixed in BaseEmissionsTracker class initialization code with the following:

if len(self._hardware) == 0:
    sys.exit("No hardware detected")

You need to import sys at the beginning, and you should place the above code after all your if/else checks on GPUs/CPUs.

### What I Did

from codecarbon import OfflineEmissionsTracker
tracker = OfflineEmissionsTracker(country_iso_code="CAN")
tracker.start()
# GPU intensive training code
tracker.stop()

vict0rsch:
Thank you for your feedback @acatovic, we can look into this. @Lothiraldan or @kngoyal I don't have a Windows machine at hand do you?

In  the mean time, are you able to do a PR with what you'd like to see @acatovic ? That'd be great :)

acatovic:
Absolutely @vict0rsch, will do, thanks!

acatovic:
@vict0rsch I think you need to grant me access (or push privileges), otherwise I can't push my own branch and do a PR. Thanks!

vict0rsch:
@acatovic Can you do that from a fork? If that is somehow a problem I will but I think it is better practice :)

Lothiraldan:
@acatovic I can understand that the behavior is not what you expected. We will likely have some environment where we don't have access to any hardware related metrics to compute the emission. In addition, in a team some users might run the same code (with code-carbon code) on both environments that are supported and environments that are not.

If we would expose an helper function indicating if we have any supported hardware metrics, would that works for you? Something like:

```
if not code_carbon.has_supported_sources():
    sys.exit(1)
```

acatovic:
@vict0rsch absolutely, i did a fork, code change and PR has now been submitted.

[...]

PR #91, opened 2020-12-07
Fix to raise exception when no CPU/GPU hardware is detected - issue #90

acatovic:
I added a check in BaseEmissionsTracker to raise ValueError if no CPU/GPU is detected, i.e. if hardware list is empty. I also had to remove exception suppression from OfflineEmissionsTracker otherwise it will not work in that case - not sure why the exception suppression was there in the first place (you don't have one for EmissionsTracker).

With this simple fix we don't force a program exit, i.e. sys.exit(), instead we gave users a graceful option of capturing the exception and proceeding with their code. Example:

```

def tracker_start(tracker):
    if tracker is not None:
        tracker.start()

def tracker_stop(tracker):
    if tracker is not None:
        tracker.stop()

try:
    tracker = EmissionsTracker()
except ValueError:
    tracker = None

def train():
    # some insane BERTesque model...

tracker_start(tracker)
train()
tracker_stop(tracker)
```

The above can be applied in those instances where CodeCarbon is liberally scattered throughput project code AND you are unsure whether you will run into a situation where you're running the code on Windows machine with WSL.

If using the decorator, then to gracefully handle the exception you need to:

```
from codecarbon import track_emissions

def train():
        # some insane BERTesque model...

try:
    @track_emissions()
    def run_train():
        train()
except ValueError:
    train()
```

Not very elegant, but again, this is extreme case where you want to handle all the scenarios. 99% people benchmarking energy usage when training are going to be doing it in "real" environment with some nice CPUs and GPUs, so this PR essentially just gives an exception to those users (like my naive self!) who get overly excited and try this lovely library on some whimsical Windows machine under WSL1 :-).

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -7,7 +7,6 @@ import logging
 import os
 import time
 import uuid
-import warnings
 from abc import ABC, abstractmethod
 from datetime import datetime
 from functools import wraps
@@ -79,7 +78,7 @@ class BaseEmissionsTracker(ABC):
         
         # Print warning if no supported hardware is found'
         if not self._hardware:
-            warnings.warn('CODECARBON : No CPU/GPU tracking mode found. This '\
+            logger.warning('CODECARBON : No CPU/GPU tracking mode found. This '\
                 'may be due to your code running on Windows WSL, or due to '\
                 'unsupported hardware (see '\
                 'https://github.com/mlco2/codecarbon#infrastructure-support)')
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -75,7 +75,7 @@ class BaseEmissionsTracker(ABC):
         elif cpu.is_rapl_available():
             logger.info("CODECARBON : Tracking Intel CPU via RAPL interface")
             self._hardware.append(CPU.from_utils(self._output_dir, "intel_rapl"))
-        
+
         # Print warning if no supported hardware is found'
         if not self._hardware:
             logger.warning(
@@ -217,7 +217,7 @@ class OfflineEmissionsTracker(BaseEmissionsTracker):
     Offline implementation of the `EmissionsTracker`
     In addition to the standard arguments, the following are required.
     """
-    
+
     @suppress(Exception)
     def __init__(
         self,
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -78,10 +78,12 @@ class BaseEmissionsTracker(ABC):
         
         # Print warning if no supported hardware is found'
         if not self._hardware:
-            logger.warning('CODECARBON : No CPU/GPU tracking mode found. This '\
-                'may be due to your code running on Windows WSL, or due to '\
-                'unsupported hardware (see '\
-                'https://github.com/mlco2/codecarbon#infrastructure-support)')
+            logger.warning(
+                "CODECARBON : No CPU/GPU tracking mode found. This "
+                "may be due to your code running on Windows WSL, or due to "
+                "unsupported hardware (see "
+                "https://github.com/mlco2/codecarbon#infrastructure-support)"
+            )
 
         # Run `self._measure_power` every `measure_power_secs` seconds in a background thread
         self._scheduler.add_job(
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -75,6 +75,9 @@ class BaseEmissionsTracker(ABC):
         elif cpu.is_rapl_available():
             logger.info("CODECARBON : Tracking Intel CPU via RAPL interface")
             self._hardware.append(CPU.from_utils(self._output_dir, "intel_rapl"))
+        
+        if not self._hardware:
+            raise ValueError('Hardware list empty - no CPU/GPU found.')
 
         # Run `self._measure_power` every `measure_power_secs` seconds in a background thread
         self._scheduler.add_job(
@@ -208,8 +211,7 @@ class OfflineEmissionsTracker(BaseEmissionsTracker):
     Offline implementation of the `EmissionsTracker`
     In addition to the standard arguments, the following are required.
     """
-
-    @suppress(Exception)
+    
     def __init__(
         self,
         country_iso_code: str,
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -7,6 +7,7 @@ import logging
 import os
 import time
 import uuid
+import warnings
 from abc import ABC, abstractmethod
 from datetime import datetime
 from functools import wraps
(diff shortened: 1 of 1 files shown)


codecarbon/emissions_tracker.py
```diff
@@ -75,6 +75,9 @@ def __init__(
         elif cpu.is_rapl_available():
             logger.info("CODECARBON : Tracking Intel CPU via RAPL interface")
             self._hardware.append(CPU.from_utils(self._output_dir, "intel_rapl"))
+        
+        if not self._hardware:
+            raise ValueError('Hardware list empty - no CPU/GPU found.')
```
kngoyal, line 80: @acatovic any error / log command messages should be written like `"CODECARBON : Hardware list empty - no CPU/GPU  tracking mode found."`

codecarbon/emissions_tracker.py
```diff
@@ -208,8 +211,7 @@ class OfflineEmissionsTracker(BaseEmissionsTracker):
     Offline implementation of the `EmissionsTracker`
     In addition to the standard arguments, the following are required.
     """
-
-    @suppress(Exception)
+
```
kngoyal, line 214: We want to suppress any exceptions or errors from Codecarbon so that it does not fail user code. @acatovic

kngoyal (changes requested):

codecarbon/emissions_tracker.py
acatovic: Understood, thank @kngoyal

codecarbon/emissions_tracker.py
acatovic: Thanks for the comments @kngoyal. In that case my proposal for raising an exception will not work. What do you think of these two alternatives:

- Print a Python warning "CODECARBON : Hardware list empty - no supported hardware found - emitting emissions.csv"

- sys.exit("CODECARBON : Hardware list empty - no supported hardware found - emitting emissions.csv")

The first one still lets users know that no supported hardware is found, but does not impact code flow. The 2nd one is a hard exit.

Also I suggest making the warning/error string more generic, since I'm assuming CodeCarbon will not work on ARM, Power, M1 (new Apple machines) etc, correct?

acatovic:
@kngoyal I pushed an update, please have a look if it's better now. I changed from exception to warning, i've put back exception suppression. To me this looks a lot nicer now - user will get a visible warning but code will continue to execute.
image

acatovic (commented):
I pushed an update to this PR - I changed from exception to warning and i've put back exception suppression. To me this looks a lot nicer now - user will get a visible warning but code will continue to execute (see Jupyter Notebook example below):
image

codecarbon/emissions_tracker.py
```diff
@@ -75,6 +76,13 @@ def __init__(
         elif cpu.is_rapl_available():
             logger.info("CODECARBON : Tracking Intel CPU via RAPL interface")
             self._hardware.append(CPU.from_utils(self._output_dir, "intel_rapl"))
+        
+        # Print warning if no supported hardware is found'
+        if not self._hardware:
+            warnings.warn('CODECARBON : No CPU/GPU tracking mode found. This '\
```
Lothiraldan, line 82: A warning sounds good, but let's user the logger instead of the warnings module:
```suggestion
            logger.warning('CODECARBON : No CPU/GPU tracking mode found. This '\
```

acatovic: Perfect @Lothiraldan I did the update on PR as you suggested and removed warnings import as it's unnecessary.

Side-comment, there is a nice explanation in Python docs (https://docs.python.org/3/howto/logging.html) when to use warnings vs logging module for a warning, and it's in line with your suggested change:

"warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning

logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted"

codecarbon/emissions_tracker.py
```diff
@@ -75,6 +75,13 @@ def __init__(
         elif cpu.is_rapl_available():
             logger.info("CODECARBON : Tracking Intel CPU via RAPL interface")
             self._hardware.append(CPU.from_utils(self._output_dir, "intel_rapl"))
+        
+        # Print warning if no supported hardware is found'
+        if not self._hardware:
+            logger.warning('CODECARBON : No CPU/GPU tracking mode found. This '\
+                'may be due to your code running on Windows WSL, or due to '\
+                'unsupported hardware (see '\
+                'https://github.com/mlco2/codecarbon#infrastructure-support)')
```
LiamConnell, line 84:
```suggestion
            logger.warning(
                "CODECARBON : No CPU/GPU tracking mode found. This "
                "may be due to your code running on Windows WSL, or due to "
                "unsupported hardware (see "
                "https://github.com/mlco2/codecarbon#infrastructure-support)"
            )
```

LiamConnell: Looks like the linter was tripping this up. Use this and we can merge!

LiamConnell (changes requested):

codecarbon/emissions_tracker.py
acatovic: @LiamConnell done as requested, see latest commit, thank you!

codecarbon/emissions_tracker.py
```diff
@@ -203,13 +212,13 @@ def _measure_power(self) -> None:
         self._last_measured_time = time.time()
 
 
+@suppress(Exception)
```
LiamConnell, line 215: This decorator doesn't work at the class level. It should be left in the constructor method
```suggestion
```

codecarbon/emissions_tracker.py
```diff
@@ -203,13 +212,13 @@ def _measure_power(self) -> None:
...
 class OfflineEmissionsTracker(BaseEmissionsTracker):
     """
     Offline implementation of the `EmissionsTracker`
     In addition to the standard arguments, the following are required.
     """
-
-    @suppress(Exception)
```
LiamConnell, line 212: This line should be left as it is

LiamConnell (changes requested):
The exception suppression should be left at the method level

codecarbon/emissions_tracker.py
acatovic: @LiamConnell that's very strange, I didn't add any suppression :-(, I did however sync my branch with the upstream before adding your previous suggestion....I will remove it and push again, cheers!

acatovic: @LiamConnell Ok now I understand what's happening - there is lot of back-and-forth happening here. In the original codecarbon codebase that I forked, the exception suppression was already there. My original solution was to actually handle unsupported/lack of hardware via ValueError exception, and I removed exception suppression. @kngoyal then requested I put the suppression back and instead handle this issue via a warning. So I put the suppression back and implemented the warning. Now I've been requested to remove the suppression :-o.

I think there needs to be some agreement regarding certain design patterns, such as exception handling. However I don't think this should be tied to this PR. I've implemented an explicit warning for the issue raised, general exception handling and suppression should be handled somewhere else, when a consensus/general agreement is reached in the project.

LiamConnell: I think I understand where the confusion comes from: It comes from _where_ you added the suppression back in. You added it as a class decorator rather than a method decorator. It only works as a method decorator. 

There is agreement on the design pattern: all public methods of the EmissionsTracker objects have exceptions suppressed. The reasoning is that this library should NEVER cause the user's code to fail, even if the library fails to track carbon. 

You still need to add this line back in, since the method must be decorated.

acatovic: I see @LiamConnell , thanks for the clarification and apologies for the confusion! I have put the suppression back on the constructor and pushed the changes. Phew!

LiamConnell:
One last thing will allow us to merge! The precommit hook is currently failing due to some annoying formatting issue.  I think I'm unable to solve this on my end since this is a fork not a branch. This will solve it @acatovic : 

```
pip install pre-commit
pre-commit install
pre-commit run --all-files
git commit -am "precommit formatting changes"
git push
```

Goyal-Kamal (approved):

LiamConnell (approved):

acatovic:
Thanks for the support @LiamConnell  and @Goyal-Kamal !

PR #97, opened 2021-01-20
Installing dataclasses conditionally on the python version.

mstechly:
There are two version of `dataclasses` package – one that's installable with `pip` and the other one which is internal for python from `3.7`. The first one is needed if you're using python `3.6`.

The problem arises when you're using python 3.7 and have one package that installs the "external dataclasses" (i.e. `codecarbon`) and another one which uses "internal dataclasses". 
Given that both packages have the same name, the code imports "external" when it should be using "internal", which leaves to an error because they actually differ.

This PR solves this by installing the "external" `dataclasses` only if python version is lower than 3.7.


mstechly:
@LiamConnell – could I get some comments on this?
Is there sth wrong with this approach, any insight why the builds might be failing / how to re-run them?

LiamConnell:
Hi @mstechly. 

The "greetings" GH action is confusing to me as well and I've asked @Lothiraldan to help look into it. 

The "build 3.8" test that is failing is due to a numerical precision error, which I've fixed recently in a different PR. Can you try merging the latest from master and trying again?

LiamConnell:
@mstechly The GH action has been removed. Merging the most recent changes from master should let this pass all the CI steps.

mstechly:
I've merged `master` in and all the checks are passing :) 
Thanks!

Lothiraldan:
The conda package might need to be fixed as well: https://github.com/mlco2/codecarbon/blob/master/.conda/meta.yaml

mstechly:
@Lothiraldan I guess I should also change `requirements-dev.txt` then?

Lothiraldan:
Yes and unfortunately I don't think conda support environment markers:
```
Received dictionary as spec.  Note that pip requirements are not supported in conda-build meta.yaml.  Error message: Invalid version ''3.7'': invalid character(s)
```

One possible solution would be to have two recipes, one for Python < 3.7 and one for Python >= 3.7 but it will add to the maintenance cost.

mstechly:
@Lothiraldan It seems that another solution would be to use some templating tool like `jinja`? Then you could be generating `meta.yaml` on the fly dependently on the system / have two versions of `meta.yaml` that are generated this way.

Either way – I don't use `conda` and don't understand how it works, so would appreciate your help here :)

Lothiraldan:
@mstechly Unfortunately the `meta.yaml` is static and not processed at installation time but rather at build time. So it isn't possible to use jinja at installation time.

mstechly:
@Lothiraldan Makes sense.

On the other hand, I stumbled on [this guide](https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html) and I think using [coupling keys](https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#coupling-keys) could be a solution to this problem.

Though I don't see `conda_build_config.yaml` anywhere and I don't know if that's a good way to proceed in this project, so before I jump on implementing that, I'd like to hear your opinion :)

mstechly:
@Lothiraldan any comments regarding my previous suggestions>?>

Lothiraldan:
Hi @mstechly I gave coupling keys a try in one of our internal conda package that have a similar challenge and unfortunately I couldn't get it to work and found it was quite difficult to debug the slow build process. If you can get it to work, I will happily review it.

Keep in mind that the results of any pre-processing would be multiple conda archives and manually maintain several conda packages is indeed more maintenance work but those files shouldn't change much.

mstechly:
@Lothiraldan Sounds like a lot of _fun_ and not something I think I can do a better job than you did.

It looks like there's no good way out :/ 
I think this suggestion is the best one so far but adds maintenance cost as you've mentioned:
https://github.com/mlco2/codecarbon/pull/97#issuecomment-771026877
?

Lothiraldan:
This is now ready for review

Lothiraldan (approved):

PR #98, opened 2021-01-20
Feature/add canada regions

LiamConnell:
This adds a new dataset: `canada_energy_mix.json`. 

It uses the same logic as the global energy mix conversion to emissions rate. 

When calculating the regional emissions, the code will look for emissions data (as is the case with USA data) and then fall back on energy mix if emissions data is not present.


LiamConnell:
This PR is ready for review

Lothiraldan:
What is the source of the canada data?

LiamConnell:
> What is the source of the canada data?

I found the data in a previous PR from months ago here: https://github.com/mlco2/codecarbon/pull/4/files. Since it was under the `2016` directory, I was assuming that it is from the same data source as the other files.

LiamConnell:
I've updated this PR to also add the Canadian regional map to CarbonBoard. This PR is ready for review

vict0rsch:
add link to https://github.com/mlco2/codecarbon/blob/master/codecarbon/data/cloud/impact.csv in the docs

codecarbon/core/emissions.py
```diff
@@ -78,78 +78,98 @@ def get_private_infra_emissions(self, energy: Energy, geo: GeoMetadata) -> float
...
+            )
 
-        country_emissions_data = self._data_source.get_country_emissions_data(
-            geo.country_iso_code.lower()
-        )
+            if geo.region not in country_emissions_data:
+                # TODO: Deal with missing data, default to something
+                raise Exception(
+                    f"Region: {geo.region} not found for Country with ISO CODE : {geo.country_iso_code}"
+                )
```
Lothiraldan, line 109: This exception is not catched by the following `except DataSourceException`, is that expected?

codecarbon/input.py
```diff
@@ -64,6 +70,23 @@ def get_country_emissions_data(self, country_iso_code: str) -> Dict:
...
         :return: emissions in lbs/MWh and region code
         """
-        with open(self.country_emissions_data_path(country_iso_code)) as f:
-            country_emissions_data: Dict = json.load(f)
-        return country_emissions_data
+        try:
+            with open(self.country_emissions_data_path(country_iso_code)) as f:
+                country_emissions_data: Dict = json.load(f)
+            return country_emissions_data
+        except KeyError:
```
Lothiraldan, line 77: In which case could we get a `KeyError`?

LiamConnell: I just added a comment specifying. The KeyError comes from line 39, when the filepath of the country emissions data file is looked up from the config object. If there is no file, it raises a KeyError

Lothiraldan (approved):
LGTM

Issue #101, opened 2021-01-21
Writing emissions.csv to non-default output_dir, emissions are always 0.0


enjeeneer:
* CodeCarbon version: 1.0.0
* Python version: 3.7.9
* Operating System: MacOS

### Description

I've been trying to write emissions.csv to a chosen `output_dir` (rather than the default working directory) to track emissions data across experiments in the same place. When I do so, I always get emissions of `0.0`, but when I write to the default directory this is not the case and I get a small non-zero result. I've found this to be happen across many different experiments.

In the code below, `e_path` is just the path to the directory I am trying to write to. This may be a mistake on my end, in which case any help is much appreciated, otherwise it looks like a bug.

### What I Did

```
from codecarbon import EmissionsTracker

mnist = tf.keras.datasets.mnist

(x_train, y_train), (x_test, y_test) = mnist.load_data()
x_train, x_test = x_train / 255.0, x_test / 255.0


model = tf.keras.models.Sequential(
    [
        tf.keras.layers.Flatten(input_shape=(28, 28)),
        tf.keras.layers.Dense(128, activation="relu"),
        tf.keras.layers.Dropout(0.2),
        tf.keras.layers.Dense(10),
    ]
)

loss_fn = tf.keras.losses.SparseCategoricalCrossentropy(from_logits=True)

model.compile(optimizer="adam", loss=loss_fn, metrics=["accuracy"])

tracker = EmissionsTracker(project_name='test', output_dir=e_path)
tracker.start()
model.fit(x_train, y_train, epochs=50)
tracker.stop()
```

LiamConnell:
Hi @scottjeen thanks for bringing this up. I just replicated the problem on my machine and I see the same thing.

LiamConnell:
@scottjeen I think the problem is actually due to a bug with the scheduler. I've submitted a PR that is now pending review. In the meantime, could you try adding `measure_power_secs=3` like so:

`tracker = EmissionsTracker(project_name='test', output_dir=e_path, measure_power_secs=3)`

I think you will get a non-zero power/emissions reading

scottjeen:
@LiamConnell Thanks for your reply, unfortunately this doesn't solve the problem on my machine, `0.0` is still written to emissions.csv. When I run again without passing an `output_dir` a non-zero reading is produced, so it seems that the issue is directly related to `output_dir`. Sorry to bring bad news!

LiamConnell:
[...]

Issue #92, opened 2020-12-09
emissions always equal to 0


QS5526:
* CodeCarbon version:1.0.0
* Python version:3.7.4
* Operating System:windows 10

### Description
I have run the following code on my laptop
###########################
import tensorflow as tf

from codecarbon import EmissionsTracker

mnist = tf.keras.datasets.mnist

(x_train, y_train), (x_test, y_test) = mnist.load_data()
x_train, x_test = x_train / 255.0, x_test / 255.0

model = tf.keras.models.Sequential(
    [
        tf.keras.layers.Flatten(input_shape=(28, 28)),
        tf.keras.layers.Dense(128, activation="relu"),
        tf.keras.layers.Dropout(0.2),
        tf.keras.layers.Dense(10),
    ]
)

loss_fn = tf.keras.losses.SparseCategoricalCrossentropy(from_logits=True)

model.compile(optimizer="adam", loss=loss_fn, metrics=["accuracy"])

tracker = EmissionsTracker()
tracker.start()
model.fit(x_train, y_train, epochs=1000)
emissions: float = tracker.stop()
print(emissions)
#################################

But the emissssion is always equal to 0.0

What does this result mean?

QS5526:
With  the attached code, why my emission result is equal to 0:
[codecarbon_test.zip](https://github.com/mlco2/codecarbon/files/5671229/codecarbon_test.zip)

Thank you for your reply.

DavidGRX:
Hi:
  I'm having a similar problem. Output is always 0.0.   
I've installed the Intel tool to have CPU support. 
Is there any way to increase the precision of the output? My process may consume grams of C02 (far from kg)

Thank you for the tool!

LiamConnell:
Hi @DavidGRX and @QS5526, This seems to be a bug that arose from the scheduler. We have a PR (#102) that is pending review. In the meantime, you can try setting the parameter `measure_power_secs=3` to your `EmissionsTracker` object. This should give a non-zero power/emissions reading

QS5526:
Hi @LiamConnell, thank you for your answer but unfortunately, setting the parameter measure_power_secs (=3or 1, or 0.01,...) , doesn't work in my case.
My power emission is still equal to "none" or "0".
Do you have an other solution to fix the problem?

LiamConnell:
Ok @QS5526  I've added a line to your script to print out all logs. Please send the output of this script so I can look at the logs:

```
import tensorflow as tf
import logging

[...]

PR #102, opened 2021-01-21
Add final measurement on process completion

LiamConnell:
When the process finishes running, the scheduler is shutdown. However, there might have been some computation (and therefor power) usage between the last scheduled measurement and shutdown. Running the method once more will account for this. 

Multiple issues have come in (#101 #92) regarding power measurements of 0. It's likely (although not confirmed) that this is happening when the process completes before the scheduler runs a single time.


LiamConnell:
This PR is ready for review

kngoyal (approved):

kngoyal:
Good catch, LGTM !

PR #103, opened 2021-01-22
set logging level for module logger not global

LiamConnell:
The way we set the logging level would currently override the user's global logging config. This fixes it so that we only modify CC's logger's level.

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -23,8 +23,8 @@ from codecarbon.external.hardware import CPU, GPU
 from codecarbon.input import DataSource
 from codecarbon.output import BaseOutput, EmissionsData, FileOutput
 
-logging.basicConfig(level=os.environ.get("CODECARBON_LOGLEVEL", "WARN"))
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


codecarbon/emissions_tracker.py
```diff
@@ -23,8 +23,8 @@
 from codecarbon.input import DataSource
 from codecarbon.output import BaseOutput, EmissionsData, FileOutput
 
-logging.basicConfig(level=os.environ.get("CODECARBON_LOGLEVEL", "WARN"))
 logger = logging.getLogger(__name__)
+logger.setLevel(level=os.environ.get("CODECARBON_LOGLEVEL", "WARN"))
```
Goyal-Kamal, line 27: can we update this in the documentation for a user to set Logging Level by setting `CODECARBON_LOGLEVEL` environment variable

Lothiraldan:
We need to set the level for the logger `codecarbon`, if we only set it for the logger `codecarbon.emission_tracker`, other loggers for others modules won't be configured correctly. It's also easier for user to set it up programmatically, for example:

```
logging.getLogger("codecarbon").setLevel(logging.ERROR)
```

LiamConnell:
Good catch @Lothiraldan . I've fixed this

Lothiraldan (approved):
LGTM

PR #105, opened 2021-01-22
add gpu_ids parameter to decorator

LiamConnell:
The decorator was missing the parameter `gpu_ids`

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -279,6 +279,7 @@ def track_emissions(
     offline: bool = False,
     country_iso_code: Optional[str] = None,
     region: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


Goyal-Kamal (approved):

kngoyal (approved):

codecarbon/emissions_tracker.py
```diff
@@ -279,6 +279,7 @@ def track_emissions(
     offline: bool = False,
     country_iso_code: Optional[str] = None,
     region: Optional[str] = None,
+    gpu_ids: Optional[List] = None,
```
kngoyal, line 282: can we also update the documentation here https://github.com/mlco2/codecarbon/blob/master/docs/edit/parameters.rst
for the new input param for both the decorator and class constructors

PR #107, opened 2021-01-22
Enable offline mode with cloud configuration

LiamConnell:
A likely scenario comes when code carbon runs on cloud infrastructure that does not have access to the internet. We enable the user to set `cloud_provider` and `cloud_region` instead of `country_iso_code` and `region`

If cloud parameters AND location parameters are set, the tracker will use the cloud parameters. If parameters are incorrectly configured, the tracker will fail silently in the background. 

The decorator will fail (not silently) if set to offline mode and neither cloud nor location details are provided.


Lothiraldan (approved):
We need to link to the list of valid regions/cloud providers as well

PR #115, opened 2021-02-16
[WIP] create HTTP Output

LiamConnell:


LiamConnell:
This PR is ready for review. There is one extra change to the visualization included here: instead of rounding values at the data layer, we round at the display layer so that precise values are displayed in choropleth maps

codecarbon/output.py
```diff
@@ -61,3 +68,21 @@ def out(self, data: EmissionsData):
...
+    Send emissions data to HTTP endpoint
+    """
+
+    def __init__(self, endpoint_url: str):
+        self.endpoint_url: str = endpoint_url
+
+    def out(self, data: EmissionsData):
+        payload = dataclasses.asdict(data)
+        payload["user"] = getpass.getuser()
+        resp = requests.post(self.endpoint_url, json=payload)
```
Lothiraldan, line 84: Need to add a timeout and catch any exception that could happens here

Lothiraldan (changes requested):

codecarbon/output.py
```diff
@@ -61,3 +68,25 @@ def out(self, data: EmissionsData):
...
+            payload = dataclasses.asdict(data)
+            payload["user"] = getpass.getuser()
+            resp = requests.post(self.endpoint_url, json=payload, timeout=10)
+            if resp.status_code != 201:
+                LOGGER.warning(
+                    "CODECARBON : HTTP Output returned an unexpected status code: ",
+                    resp,
+                )
+        except Exception as e:
+            LOGGER.error(e)
```
Lothiraldan, line 92: Let's log the traceback as well:

```suggestion
            LOGGER.error(e, exc_info=True)
```

Lothiraldan (approved):
LGTM

PR #118, opened 2021-03-01
Live API data - CO2Signal

LiamConnell:


vict0rsch (approved):

codecarbon/core/emissions.py
```diff
@@ -78,6 +79,11 @@ def get_private_infra_emissions(self, energy: Energy, geo: GeoMetadata) -> float
         :param geo: Country and region metadata
         :return: CO2 emissions in kg
         """
+        if co2_signal.is_available():
+            try:
+                return co2_signal.get_emissions(energy, geo)
+            except co2_signal.CO2SignalAPIError as e:
```
vict0rsch, line 85: catch any exception

codecarbon/core/co2_signal.py
```diff
@@ -0,0 +1,35 @@
+import requests
+
+from codecarbon.core.units import EmissionsPerKwh, Energy
+from codecarbon.external.geography import GeoMetadata
+
+URL = "https://api.co2signal.com/v1/latest"
+CO2_SIGNAL_API_TOKEN = None
```
Lothiraldan, line 7: It would be better to pass the token to the `is_available` and `get_emissions` functions or create an intermediary object to store it

codecarbon/core/co2_signal.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def get_emissions(energy: Energy, geo: GeoMetadata):
+    if geo.latitude:
+        params = {"lat": geo.latitude, "lon": geo.longitude}
+    else:
+        params = {"countryCode": geo.country_2letter_iso_code}
+    resp = requests.get(
+        URL, params=params, headers={"auth-token": CO2_SIGNAL_API_TOKEN}
+    )
```
Lothiraldan, line 24: Add a timeout and a test for it. You can test it with a very low timeout, like 0.00000001

codecarbon/core/emissions.py
Lothiraldan: Requests itself can raises `requests.exceptions.Exception` in various situations and the parsing of the response body can raises various exceptions, let's catch `except Exception` here

Lothiraldan (commented):
Apart from the error handling and missing timeout, LGTM

PR #135, opened 2021-04-01
feat/fallback

QinganZhao:
Fall back mode by calculating the CPU power consumptions using TDP constants


LiamConnell:
lgtm

LiamConnell (approved):

codecarbon/external/hardware.py
```diff
@@ -90,6 +101,28 @@ def _get_power_from_cpus(self) -> Power:
...
 
+    def _get_power_from_constant(self) -> Power:
+        """
+        Get CPU power from constant mode
+        :return: power in KW
+        """
+        cpu_info = cpuinfo.get_cpu_info()
+        if cpu_info:
+            model_raw = cpu_info["brand_raw"]
+            model = model_raw.split(" CPU")[0].replace("(R)", "").replace("(TM)", "")
```
Lothiraldan, line 112: Let's extract this line into a separate function so we can more easily test it. The reasoning behind that is that CPU-Info has the following warning:
```
These fields are pulled directly from the CPU and are unverified. They may contain expected results. Other times they may contain wildly unexpected results or garbage. So it would be a bad idea to rely on them.
```

I'm guessing that parsing this field is gonna fails in various condition, ARM, AMD, etc... It would be easier to update and test the parsing if the logic is isolated into a separate function

codecarbon/external/hardware.py
```diff
@@ -90,6 +101,28 @@ def _get_power_from_cpus(self) -> Power:
...
+            cpu_power_df = DataSource().get_cpu_power_data()
+            cpu_power_df_model = cpu_power_df[cpu_power_df["Name"] == model]
+            if len(cpu_power_df_model) > 0:
+                power = cpu_power_df_model["TDP"].tolist()[0] * 0.5
```
Lothiraldan, line 116: Let's extract the magic value `0.5` to a constant

Lothiraldan (changes requested):

codecarbon/external/hardware.py
```diff
@@ -82,6 +88,12 @@ def _get_power_from_cpus(self) -> Power:
         Get CPU power from Intel Power Gadget
         :return: power in kW
         """
+        if self._mode == "constant":
+            logger.warning(
+                "CODECARBON : No CPU/GPU tracking mode found. Falling back on CPU constant mode."
```
vict0rsch, line 93: Actually this could be triggered with GPU but no CPU

codecarbon/external/hardware.py
```diff
@@ -90,6 +102,43 @@ def _get_power_from_cpus(self) -> Power:
...
+    def _get_power_from_constant(self) -> Power:
+        """
+        Get CPU power from constant mode
+        :return: power in KW
+        """
+        cpu_info = cpuinfo.get_cpu_info()
+        if cpu_info:
+            model_raw = cpu_info["brand_raw"]
+            model = self._parse_cpu_model(model_raw)
+            cpu_power_df = DataSource().get_cpu_power_data()
```
vict0rsch, line 126: Does this mean that the 2000+line CSV will be read and parsed _every time_ the cpu power is required? Like every 15seconds? We should have that cached no?

codecarbon/external/hardware.py
```diff
@@ -90,6 +102,43 @@ def _get_power_from_cpus(self) -> Power:
...
+            cpu_power_df_model = cpu_power_df[cpu_power_df["Name"] == model]
+            if len(cpu_power_df_model) > 0:
+                power = (
+                    cpu_power_df_model["TDP"].tolist()[0]
+                    * CONSUMPTION_PERCENTAGE_CONSTANT
+                )
+            else:
+                logger.warning(
```
vict0rsch, line 134: Should this kind of warning only be sent the first time? I fear having this every 15 seconds will be annoying :(

vict0rsch (changes requested):

codecarbon/external/hardware.py
vict0rsch: Also, I'm not sure we should warn all the time. Thoughts @nikolaskaris @LiamConnell @Lothiraldan @vl4py @benoit-cty ?

QinganZhao: This is already able to be triggered with GPU but no CPU - I'll modify the warning message

vict0rsch (approved):

codecarbon/data/hardware/cpu_power.csv
```diff
@@ -0,0 +1,2144 @@
+Name,TDP
```
Lothiraldan, line 1: The file needs to be declared in `setup.py` `package_data` or it won't be packaged

codecarbon/core/cpu.py
```diff
@@ -209,3 +211,36 @@ def get_cpu_details(self) -> Dict:
...
+        """
+        Parse the model name from the raw name extracted from cpuinfo library
+        :return: parsed CPU name
+        """
+        return (
+            raw_name.split(" @")[0]
+            .replace("(R)", "")
+            .replace("(TM)", "")
+            .replace(" CPU", "")
+        )
```
Lothiraldan, line 230: I don't think this function needs to be a method, it's not using `self`, let's extract it and add at least one test to check that's it parsing correctly the CPU model.

codecarbon/core/cpu.py
```diff
@@ -209,3 +211,36 @@ def get_cpu_details(self) -> Dict:
...
+
+    def _get_power_from_constant(self) -> int:
+        """
+        Get CPU power from constant mode
+        :return: power in Watt
+        """
+        cpu_info = cpuinfo.get_cpu_info()
+        if cpu_info:
+            model_raw = cpu_info["brand_raw"]
+            model = self._parse_cpu_model(model_raw)
```
Lothiraldan, line 240: This can fails as we don't have any guarantee on the format of `model_raw`. We need to catch the exception to avoid crashing the user script and decide what to do in that case.

Lothiraldan (changes requested):

codecarbon/core/cpu.py
QinganZhao: I don't think the parsing function would fail even if the format is not guaranteed. The parsing is basically just split and replace - it would do nothing for that step if any of those patterns are not detected.

Lothiraldan: So far, assuming that `model_raw` is always a string, I don't see a way for it to fails. We can tackle it later if you prefer.

codecarbon/external/hardware.py
Lothiraldan: I think it's now only happening when creating the EmissionTracker object, could you confirm @QinganZhao?

Lothiraldan (approved):
LGTM, thank you for this very important feature!

codecarbon/core/cpu.py
QinganZhao: I'll add an exception in case in the future `cpuinfo` library will return something other than string (or even they changed the key name for "brand raw"). But right now it seems that it will always return a string.
https://github.com/workhorsy/py-cpuinfo

codecarbon/external/hardware.py
QinganZhao: Yes - I made the changes hours ago

vict0rsch (approved):

Issue #139, opened 2021-04-02
Contributing


SaboniAmine:
As Data for Good involves developers new to the project, a contributing update could be helpful.

It would be interesting to add to the existing one : 

- Code style conventions
- Testing routines
- Local deployment routines
- Review process

Would you see anything else ?

PR #140, opened 2021-04-05
Contributing update

SaboniAmine:
Fix: #139  

-> Coding style guidelines & review process inferred from the repo and might not be well interpreted, all the referenced practices are open to discussion!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,8 +9,8 @@ repos:
     args: [--filter-files"]
     hooks:
     -   id: isort
--   repo: https://github.com/python/black
-    rev: stable
+-   repo: https://github.com/psf/black
+    rev: 20.8b1
     hooks:
     - id: black
       args: [--safe]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,132 @@
-# Release process
+# Contributing to Code Carbon
+
+
+- [I have a question...](#questions)
+- [I found a bug...](#bugs)
+- [I have a feature request...](#features)
+- [I have a contribution to share...](#process)
+
+## <a name="questions"></a> Have a Question?
+
+Please see the [FAQ](https://mlco2.github.io/codecarbon/faq.html) for questions.
+
+
+## <a name="bugs"></a> Found a Bug?
+
+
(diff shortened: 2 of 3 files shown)


Lothiraldan (commented):
Looking good so far, we are also using https://pre-commit.com/ to enforce code style and linting, we should probably mention it in the contributing document, I will try to add it myself if no-one did it before me.

SaboniAmine:
Alright, thanks for the review.
As I am not very familiar with pre-commit (especially the isort task), help would indeed be needed to describe it.

Lothiraldan:
@SaboniAmine I've pushed a draft about code-carbon and a bit about testing and packaging, let me know what you think about it. It's a draft so we likely need to update the content form.

SaboniAmine:
Looks fine to me, I guess the expected logs will be helpful for troubleshooting.
Thanks for the packaging description item, which raises questions to me (which can be treated in a separate issue) :
- as code-carbon is tested against py38, should we reference a specific conda meta file ?
- should we freeze dependencies versions ? Is there a setup where versions could be incoherent with the current components of the package ?

Lothiraldan:
>     * as code-carbon is tested against py38, should we reference a specific conda meta file ?

We don't need a specific conda file for Python 3.8. We do have a specific Python 3.6 conda file because we need a backports of dataclasses for Python 3.6. The py37-plus conda file is working for any Python version greater or equals to Python 3.7 (See https://github.com/mlco2/codecarbon/blob/master/.conda/py37-plus/meta.yaml#L21 vs https://github.com/mlco2/codecarbon/blob/master/.conda/py36/meta.yaml#L21)
> 
>     * should we freeze dependencies versions ? Is there a setup where versions could be incoherent with the current components of the package ?

Codecarbon is a library, not a product so we should be as flexible as possible when it comes to dependencies. We don't want to have users raising issues like `codecarbon` is requiring `requests==2.25.0` but I have an environment where I'm stuck with version `requests==2.24.0` because Tensorflow requires it for example. This being said, we can probably put some lower bounds in our dependencies as we probably won't work with `requests==0.0.1` for example, but that requires checking the minimum viable version for each of our dependencies and manually test it.

Also when we extract the API / visualizer, we can probably be more precise in our requirements as the API / visualizer don't need to be in the same Python environment as other libraries so the risk of conflicts is much less likely to happen.

SaboniAmine:
Thanks for the explanations, it would indeed be counterproductive to be less flexible on the client package.

vict0rsch (approved):

Issue #14, opened 2020-09-01
Config file(s)


vict0rsch:
We could have config files to load default params.

Hierarchy could be: `args` > `./.codecarbon.conf` > `~/.codecarbon.conf` so that users can store recurring and shared arguments like `save_to_file`, `project_name`, `output_dir` etc. 

Thoughts?

PR #142, opened 2021-04-07
Using hierarchical configs

vict0rsch:
cf #14

Using `configparser` we can now configure all the `EmissionsTracker`'s arguments as follows:

* in `~/.codecarbon.config`
* in `./.codecarbon.config` = `$PWD/.codecarbon.config`
* as environment variables (e.g. `CODECARBON_MEASURE_POWER_SECS=10`)
* as previously, as arguments in the Python script

For instance:

* `~/.codecarbon.config`

	```ini
	[codecarbon]
	measure_power_secs=10
	save_to_file=false
	emissions_endpoint="localhost:7777"
	```
* `./.codecarbon.config`

	```ini
	[codecarbon]
	save_to_file=true
	output_dir=/Users/victor/emissions
	co2_signal_api_token=hahahehe
	```
* env. variables

	```bash
	export CODECARBON_GPU_IDS="0, 1"
	```

* code:
	```python
	tracker = EmissionsTracker(co2_signal_api_token="hohohuhu")
    ```

Yields attributes:
```python
{
	  "measure_power_secs": 10,
	  "save_to_file": True,
	  "emissions_endpoint": "localhost:7777",
	  "output_dir": "/Users/victor/emissions",
	  "co2_signal_api_token": "hohohuhu",
	  "gpu_ids": [0, 1],
}
```

**WIP**: tests
**TODO**: github actions


codecarbon/core/config.py
```diff
@@ -38,18 +39,25 @@ def parse_env_config() -> dict:
...
-global_config.read(Path("~/.codecarbon.config").expanduser().resolve())
-local_config.read(Path("./.codecarbon.config").expanduser().resolve())
-env_config.read_dict(parse_env_config())
-
-full_config.read_dict(global_config)
-full_config.read_dict(local_config)
-full_config.read_dict(env_config)
-
-config = full_config["codecarbon"]
+try:
```
vict0rsch, line 42: I added this in order for the user's process not to fail if they wrote faulty config files. But maybe we want it to fail? Thoughts?

vict0rsch: example faulty config:

```ini
measure_power_secs=10
```

^ is missing the section header `[codecarbon]` and `configparser` will fail

vict0rsch: caveat: the logger is defined in `emmissions_tracker.py` and so I only use prints here

codecarbon/core/config.py
```diff
@@ -0,0 +1,63 @@
...
+    local_config.read(Path("./.codecarbon.config").expanduser().resolve())
+    env_config.read_dict(parse_env_config())
+
+    full_config.read_dict(global_config)
+    full_config.read_dict(local_config)
+    full_config.read_dict(env_config)
+except Exception:
+    print("[CODECARBON] Config exception caught:")
+    print(traceback.format_exc(), end="")
+    print("Ignoring configuration, falling back to arguments [/CODECARBON]\n")
```
vl4py, line 59: `[/CODECARBON]` ?

codecarbon/emissions_tracker.py
```diff
@@ -293,9 +323,13 @@ def __init__(
                     f"Exception occured {e}"
                 )
 
-        self.country_2letter_iso_code: Optional[str] = (
-            country_2letter_iso_code.upper() if country_2letter_iso_code else None
-        )
+        self.country_2letter_iso_code: Optional[
+            str
+        ] = country_2letter_iso_code or config.get("country_2letter_iso_code")
```
vl4py, line 328:
```suggestion
        self.country_2letter_iso_code: Optional[str] = (
            country_2letter_iso_code or config.get("country_2letter_iso_code")
        )
```

codecarbon/core/config.py
vl4py: IMHO I'd rather let it crash.

vl4py (commented):
Nice to already support configuration at all levels. Kudos 💯

codecarbon/core/config.py
vict0rsch: haha no just a html-like of saying 'end of message'

codecarbon/emissions_tracker.py
vict0rsch: unfortunately this is not on me but on `black`

vl4py: They were not kidding when they said that
> Black is a PEP 8 compliant opinionated formatter. 

You must bit quite opinionated to split time hints this way.

tests/test_config.py
```diff
@@ -0,0 +1,77 @@
...
+            "CODECARBON_TEST_KEY": "this_other_value",
+        },
+    )
+    def test_parse_env_config(self):
+        self.assertDictEqual(
+            parse_env_config(),
+            {"codecarbon": {"test": "test-VALUE", "test_key": "this_other_value"}},
+        )
+
+    def custom_mock_open(self, *args, **kwargs):
```
vict0rsch, line 43: Custom mock to selectively return the right `read_data` as a function of the `path` so that the two calls to `ConfigParser().read(...)` are patched in `get_hierarchical_config`

codecarbon/core/config.py
vict0rsch: ok

tests/test_config.py
```diff
@@ -9,9 +10,28 @@
     get_hierarchical_config,
 )
 from pathlib import Path
+import builtins
+
+OPEN = builtins.open
 
 
 class TestConfig(unittest.TestCase):
+    def get_custom_mock_open(
```
vict0rsch, line 19: Custom mock to selectively return the right `read_data` as a function of the `path` so that the two calls to `ConfigParser().read(...)` are patched in `get_hierarchical_config`

codecarbon/emissions_tracker.py
```diff
@@ -28,6 +29,19 @@
...
+# Imagine:
+#   1/ emissions_endpoint=localhost:8000 in ~/.codecarbon.config
+#   2/ Inside the script, the user cannot disable emissions_endpoint with
+#   EmissionsTracker(emissions_endpoint=None) since the config logic will use the one in
+#   the config file.
+#
+# To fix this, a complex move would be to have default values set to the sentinel:
+# _sentinel = object()
+# see: https://stackoverflow.com/questions/67202314/
+#      python-distinguish-default-argument-and-argument-provided-with-default-value
```
Lothiraldan, line 43: That could be problematic in some cases

tests/test_emissions_tracker.py
```diff
@@ -38,11 +42,20 @@ class TestCarbonTracker(unittest.TestCase):
...
+        self.temp_path = Path(self.temp_dir.name)
+        self.emissions_file_path = self.temp_path / "emissions.csv"
+        # builtins.open is patched not to open ~/.codecarbon.config nor
+        # ./.codecarbon.config so that the user's local configuration does not
+        # alter tests
+        patcher = mock.patch(
+            "builtins.open", new_callable=get_custom_mock_open(empty_conf, empty_conf)
+        )
+        self.addCleanup(patcher.stop)
+        patcher.start()
```
Lothiraldan, line 55: You likely wants to clean the patcher in `tearDown` as well

Lothiraldan (commented):
LGTM overall, few comments

vict0rsch:
@Lothiraldan  cheers ; any idea why the tests fail?

codecarbon/emissions_tracker.py
vict0rsch: How is it handled in comet? Is there a sentinel, something else or nothing at all?

It looks like it's not handled but I could be wrong

tests/test_emissions_tracker.py
vict0rsch: I thought `addCleanup` would do that

benoit-cty:
Hi,
Why did you merge the 'api' branch with this branch ?
The config reader you code is for the package, isn't it ?

vict0rsch:
I was thinking we'd need those more for the api release sinceI expect the number of params to increase, like api keys etc. 

So I thought we can keep master as is with those few parameters script-based and release that feature with the api but we could change this

benoit-cty:
> I was thinking we'd need those more for the api release sinceI expect the number of params to increase, like api keys etc.
> 
> So I thought we can keep master as is with those few parameters script-based and release that feature with the api but we could change this

I see, it is related to the API in that sense. But I think you do not want to merge the api branch like you did to avoid conflict because they have been many change since then (see PR152).
So I think it is better do keep you PR open without the api merge and merge it to master after the api has been merge to master.

tests/test_emissions_tracker.py
```diff
@@ -103,6 +116,7 @@ def raise_timeout_exception(*args, **kwargs):
         heavy_computation(run_time_secs=2)
         emissions = tracker.stop()
         self.assertEqual(1, mocked_requests_get.call_count)
```
khillion, line 118: This assertion breaks in Github action as well as local tests with tox

tests/test_emissions_tracker.py
```diff
@@ -76,7 +89,7 @@ def test_carbon_tracker_TWO_GPU_PRIVATE_INFRA_CANADA(
         self.assertEqual(
```
khillion, line 89: The assertion just above breaks too (line 88, I cannot comment directly this line)

khillion (commented):
@vict0rsch I have highlighted the assertion in the tests that seem to break both locally (for me) and in github actions. It is odd you could have tests running on your local instance.

Hope this helps, I do not know the code yet to quickly fix this.

vict0rsch:
@khillion thanks for taking a look, I'm aware those are two problems from the tox logs but I just can't wrap my head around why this would fail on this branch but not on master...

vict0rsch:
TODO: update readme

benoit-cty (approved):

PR #148, opened 2021-04-16
Fix test flakyness in TestCO2Signal.test_get_emissions_TIMEOUT

Lothiraldan:
Sometimes (usually in the python 3.6 test environement) the raised exception
was not the expected one.

### DIFF
--- a/tests/test_co2_signal.py
+++ b/tests/test_co2_signal.py
@@ -39,5 +39,7 @@ class TestCO2Signal(unittest.TestCase):
         assert round(result, 5) == 0.58765
 
     def test_get_emissions_TIMEOUT(self):
-        with self.assertRaises(requests.exceptions.ConnectionError):
+        with self.assertRaises(
+            (requests.exceptions.ConnectionError, requests.exceptions.ReadTimeout)
+        ):
             co2_signal.get_emissions(self._energy, self._geo, timeout=0.000001)


SaboniAmine:
Could not reproduce on my setup, but the fix looks good to me.
I however got a question about co2 signal token : is it injected or the API really works without token ?

SaboniAmine (approved):

Lothiraldan:
> Could not reproduce on my setup, but the fix looks good to me.
> I however got a question about co2 signal token : is it injected or the API really works without token ?

It looks like it is injected https://github.com/mlco2/codecarbon/blob/master/codecarbon/core/co2_signal.py#L25 & https://github.com/mlco2/codecarbon/blob/master/codecarbon/emissions_tracker.py#L116

SaboniAmine:
As the usage seems to be limited ([rate limit](https://docs.co2signal.com/#introduction)), it might be interesting to let users inject their own token if they have one.
I do not know if this will blocking in the future (neither how the users might use this function), but the 30 API calls / hour limit rate seems to be something that we should care about.

Lothiraldan:
> As the usage seems to be limited ([rate limit](https://docs.co2signal.com/#introduction)), it might be interesting to let users inject their own token if they have one.
> I do not know if this will blocking in the future (neither how the users might use this function), but the 30 API calls / hour limit rate seems to be something that we should care about.

Users can already do it https://github.com/mlco2/codecarbon/blob/master/codecarbon/emissions_tracker.py#L48 I think

SaboniAmine:
Great ! It might be good to document somewhere in the contributing as my feeling is that one day it will involve some unexpected behavior. I will open a PR tomorrow.
Thanks for the fix.

PR #157, opened 2021-04-29
API: Carbon Database 

alencon:
Hi teams, 

I have updated a new database. What do you think about that model ? 

**issue:**  
The user entity  is not really adapted , I would like to discuss on that part . 
What is your thoughts ?  
From my  opinion  , a user could belong to different teams.

Here the model : 
CodecarbonDB


benoit-cty:
Thanks Marc,
I think Users could be connected to evry entities in a "n to n" way, with a status telling the authorizations they have.
You could also add an API Key entities, also connected to users, and projects.

But : why did you add 75 commits from master ? You have to remove them. Let's talk about that on phone ;-)

benoit-cty (changes requested):
You have to remove all master commit, sorry.

SaboniAmine:
Thanks for the proposal of update!
I see no change except the provided image.
Should this be transformed as an issue ? We could open a PR after we agree the new model, with a first implementation in mind.

vict0rsch:
I agree with @SabAmine issues would be better suited to discuss the model, I'll do it :)

benoit-cty:
Wrong merge direction.

PR #160, opened 2021-05-02
refacto: isolate infra implementation domain objects,

SaboniAmine:
experiments / org in memory repo tests


benoit-cty (approved):
OK,
You could avoid pre-commit failed by running:
`pre-commit run --all-files --show-diff-on-failure --color=always`
Before commit.

PR #166, opened 2021-05-06
Count RAM consumption

vict0rsch:
* using 3W/8GB -> https://www.crucial.com/support/articles-faq-memory/how-much-power-does-memory-use
* looking for children processes (**help needed: is this how we should do it?**)


vict0rsch:
@Lothiraldan any idea/opinion on the methodology and tests?

vict0rsch:
@Lothiraldan my main concern is actually about https://github.com/mlco2/codecarbon/pull/166/files#diff-cf42f273abce88a8c0b9a5f3133f00f5b53a8aa0f8542f0ab10830bf0b6fecc6R190 do you think it's a good enough way of getting the memory used by the main process (where the tracker was instantiated) and its children?

Lothiraldan:
> @Lothiraldan my main concern is actually about https://github.com/mlco2/codecarbon/pull/166/files#diff-cf42f273abce88a8c0b9a5f3133f00f5b53a8aa0f8542f0ab10830bf0b6fecc6R190 do you think it's a good enough way of getting the memory used by the main process (where the tracker was instantiated) and its children?

It looks ok but I would check two things manually:
* How fast it can retrieves the memory of lots processes
* How it behaves if a children dies between the time it was returned in the children list and the time we try to get its RSS

benoit-cty:
What do you think of having a global option to tell CodeCarbon if we want to monitor only the process or the whole machine ?
I think it could be useful to have the option to run CC as a side program to monitor the whole computer. For example if you do not use Python for the code to monitor or if you want to monitor the whole server.

I say it here because the RAM monitoring will be the first part to monitor the process and not the whole machine, right ?

vict0rsch:
Precision can't be too high https://pythonspeed.com/articles/measuring-memory-python

benoit-cty (approved):
I've remove an info log, because it is already logged in debug:
```
[codecarbon INFO @ 14:29:03] RAM Power Consumption : Power(kW=0.012594283007999998)
[codecarbon DEBUG @ 14:29:03] Energy consumed for all RAM : Energy(kwh=0.0007485257246086861)=Power(kW=0.012594283007999998) power x Time(seconds=7.649838447570801)
```

PR #167, opened 2021-05-06
Unique logger

vict0rsch:
* Use a single name for all loggers : `"codecarbon"`
* make `log_level` an init arg to the tracker to change it across the package
* `log_level` is *not* case sensitive and should be a string in `{"DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"}`
* corollary: `log_level` can be set in a config file

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -3,7 +3,6 @@ Contains implementations of the Public facing API: EmissionsTracker,
 OfflineEmissionsTracker and @track_emissions
 """
 
-import logging
 import os
 import time
 import uuid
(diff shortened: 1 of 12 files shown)


codecarbon/emissions_tracker.py
```diff
@@ -79,70 +75,88 @@ def __init__(
...
                 CPU.from_utils(self._output_dir, "intel_power_gadget")
             )
         elif cpu.is_rapl_available():
-            logger.info("CODECARBON : Tracking Intel CPU via RAPL interface")
+            logger.info("Tracking Intel CPU via RAPL interface")
             self._hardware.append(CPU.from_utils(self._output_dir, "intel_rapl"))
         else:
             logger.warning(
-                "CODECARBON : No CPU tracking mode found. Falling back on CPU constant mode."
+                "No CPU tracking mode found. " + "Falling back on CPU constant mode."
```
SaboniAmine, line 150: un seul string ne suffit pas ?

codecarbon/emissions_tracker.py
```diff
@@ -79,70 +75,88 @@ def __init__(
...
             )
-            logger.info("CODECARBON : Tracking using constant")
+            logger.info("Tracking using constant")
             tdp = cpu.TDP().tdp
             if tdp:
                 self._hardware.append(CPU.from_utils(self._output_dir, "constant", tdp))
             else:
                 logger.warning(
-                    "CODECARBON : Failed to match CPU TDP constant. Falling back on a global constant."
+                    "Failed to match CPU TDP constant. "
```
SaboniAmine, line 158: idem

SaboniAmine (approved):

codecarbon/emissions_tracker.py
vict0rsch: c'était pour black :)

PR #170, opened 2021-05-08
User entity

SaboniAmine:
User management router & refacto

### DIFF
--- a/carbonserver/carbonserver/api/infra/repositories/repository_users.py
+++ b/carbonserver/carbonserver/api/infra/repositories/repository_users.py
@@ -1,9 +1,6 @@
-import hashlib
-import os
 import secrets
 from typing import List
 
-from pydantic import SecretStr
 
 from carbonserver.api import schemas
 from carbonserver.api.domain.user import User
@@ -23,9 +20,9 @@ class InMemoryRepository(User):
                 user_id=self.id,
                 name=user.name,
                 email=user.email,
-                hashed_password=_hash_password(user.password),
+                password= user.password,
                 api_key=_api_key_generator(),
-                is_active=user.is_active,
+                is_active=True,
             )
         )
         return self.users[self.id - 1]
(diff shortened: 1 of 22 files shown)


carbonserver/tox.ini
```diff
@@ -8,7 +8,7 @@ deps =
 
 commands =
     pip install -e .
-    python -m pytest
+    python -m pytest -vv
```
benoit-cty, line 11: Are you sure you want to commit this ?

carbonserver/carbonserver/api/infra/repositories/repository_users.py
```diff
@@ -0,0 +1,56 @@
...
+        user = [user for user in self.users if user.user_id == user_id][0]
+        return user
+
+    def list_users(self):
+        user_ids = [user.user_id for user in self.users]
+        return user_ids
+
+
+def _hash_password(password: str) -> SecretStr:
+    salt = _random_generator()  # Extract random generation to make it easier to mock
```
benoit-cty, line 43: How can you check the password of the user with a random salt when he will log in ?
We have to store the salt in the database, with the user's information, isn't it ?

benoit-cty (commented):
Thanks Amine. I put my questions in code.

carbonserver/carbonserver/api/infra/repositories/repository_users.py
SaboniAmine: Indeed, the salt should be stored too. However this implementation would require a lot of User internal methods to check the password validity (request salt + hashed key from db, hash the user input and compare results) or to update it.
I found another implementation that seems interesting : https://variable-scope.com/posts/storing-and-verifying-passwords-with-sqlalchemy. We could first use secret passwords for the first version, and try to implement this authentication logic after. Wdyt ?

carbonserver/tox.ini
SaboniAmine: I guess it will be useful when we will add those tests to the CI.

PR #172, opened 2021-05-12
run carbonserver with docker-compose

khillion:
Hi all, 

This is to start using `docker-compose` to run the carbonserver app.

* Run the carbonserver app with `docker-compose`
* Switch between sqlite or postgres (comment/uncomment in `docker-compose.yml` the database)
```yaml
    environment:
      # DATABASE_URL: sqlite:///./code_carbon.db
      DATABASE_URL: postgresql://${DATABASE_USER:-codecarbon-user}:${DATABASE_PASS:-supersecret}@postgres/${DATABASE_NAME:-codecarbon_db}
```

> Access the app on port `8008` instead of default `8000`, see `ports` in `docker-compose.yml`.

Using postgresql instead of sqlite leads to some issues when testing some endpoints (database rules seem to be respected only with psql):
* For instance, if you try to `PUT` emission with the given examples it fails for several reasons:
    * `experiment_id` is not an `Integer`
    * Since `experiment_id` is a `ForeignKey`, corresponding experiment entry needs to be created too

### DIFF
--- a/carbonserver/Dockerfile
+++ b/carbonserver/Dockerfile
@@ -1,0 +1,19 @@
+# Dockerfile
+
+# pull the official docker image
+FROM python:3.9.4-slim
+
+# set work directory
+WORKDIR /carbonserver
+
+# set env variables
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
+# install dependencies
+RUN apt update && apt install -y gcc
+COPY requirements-dev.txt .
+RUN pip install -r requirements-dev.txt
+
+# copy project
+COPY . .
--- a/carbonserver/Dockerfile
+++ b/carbonserver/Dockerfile
@@ -1,7 +1,7 @@
 # Dockerfile
 
 # pull the official docker image
-FROM python:3.9.4-slim
+FROM python:3.8.10-slim
 
 # set work directory
 WORKDIR /carbonserver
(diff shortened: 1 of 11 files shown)


khillion:
To test,`docker-compose up` in the project root directory and visit `localhost:8008/docs`.

carbonserver/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Dockerfile
+
+# pull the official docker image
+FROM python:3.9.4-slim
```
benoit-cty, line 4: Or maybe 3.8 as it is the default for Ubuntu 20.04?

docker-compose.yml
```diff
@@ -28,6 +28,19 @@ services:
   #     - codecarbon_api
   #   restart: unless-stopped
 
+  carbonserver:
+    build: ./carbonserver
+    command: uvicorn main:app --host 0.0.0.0
```
benoit-cty, line 33: I suggest to add - - reload to allow dynamic reloading on file changes.

benoit-cty (commented):
Thanks you for your contribution.

docker-compose.yml
khillion: Indeed, that is more convenient for development environment, thanks!

benoit-cty (approved):
I'm OK to merge to api branch.

PR #177, opened 2021-05-21
Seeding API db template

SaboniAmine:
We could use the execute(sql_statement) operation to generate first sets of data to manipulate the API.
This does not support heavy load as all optimisation is made by the db engine, a better usage of Alembic API is possible


benoit-cty (approved):
Perfect: It works with an empty database, and with an existing.

PR #180, opened 2021-05-26
Update data gcp impact.csv

AlexisBogroff:
On the official Google GCP data there are 2 values to consider:
- CFE (Carbon Free Emission): the percentage of zero-emission source usage
- Intensity, the (negative) impact while not on a Carbon-Free-Emission source

I assumed the impact should be computed as follows:
impact = intensity * (1-CFE)

Please confirm this methodology! :D


vict0rsch:
Data from https://cloud.google.com/sustainability/region-carbon

also Region Picker: https://googlecloudplatform.github.io/region-picker/

vict0rsch (approved):

PR #186, opened 2021-06-11
Simplify attribute setting from arguments & conf

vict0rsch:
In order to allow people to overwrite a `config` value with None, I implement the logic of default arguments with a `_sentinel` dummy object whose role is simply to be unique and impossible to be mistaken for a user-provided argument.

Addresses the current `emissions_tracker.py` intro:

"""

/!\ Warning: current implementation prevents the user from setting any value to None
from the script call

Imagine:

  1. `emissions_endpoint=localhost:8000` in `~/.codecarbon.config`
  2. Inside the script, the user cannot disable `emissions_endpoint` with `EmissionsTracker(emissions_endpoint=None)` since the config logic will use the one in the config file if the default argument is `None`.

Alternative: `EmissionsTracker(emissions_endpoint=False)` would work

**To fix this**, a complex move would be to have default values set to the sentinel:
_sentinel = object()
see: https://stackoverflow.com/questions/67202314/python-distinguish-default-argument-and-argument-provided-with-default-value 

"""

^ This is implemented in the PR in `BaseEmissionsTracker._get_conf(var, default, return_type)`

so that 

```python
        self._emissions_endpoint = (
            emissions_endpoint
            if emissions_endpoint is not None
            else conf.get("emissions_endpoint", None)
        )
        self._save_to_file = (
            save_to_file
            if save_to_file is not None
            else conf.getboolean("save_to_file", True)
        )
        if self._save_to_file == "False":
            self._save_to_file = False
```

becomes 

```python
        self._emissions_endpoint = self._get_conf(emissions_endpoint)
        self._save_to_file = self._get_conf(save_to_file, True, bool)
```

at the cost of all args defaulting to `_sentinel` in the `__init__(self, ...)` and a **new requirement [varname](https://github.com/pwwang/python-varname)**


benoit-cty (approved):
Nice work, I was just thinking of something like this.

SaboniAmine (approved):
Looks cool! Could it be also used to inject secrets / api_keys once the user will be registered with the cli ?

vict0rsch:
Yeah anything that's part of the constructor can be read from home/cwd config files and env vars using _conf

benoit-cty:
I've fixed the missing packag, but now we have an error that I don't know how to fix it:
`conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform linux-64: {'varname'}{
`

vict0rsch:
@Lothiraldan any idea beyond extracting from the package what we need so that it's not a dependency anymore?

vict0rsch:
source -> https://github.com/pwwang/python-varname/blob/master/varname/core.py#L294

vict0rsch:
```python
In [4]: tracker._conf
Out[4]: 
{'log_level': 'info',
 'project_name': 'forbidden',
 'measure_power_secs': 15,
 'output_dir': '.',
 'emissions_endpoint': None,
 'api_endpoint': 'https://api.codecarbon.io',
 'co2_signal_api_token': None,
 'save_to_file': True,
 'save_to_api': False,
 'api_call_interval': 8,
 'hardware': ['CPU(Intel Power Gadget)'],
 'gpu_ids': None}
```

Lothiraldan:
> @Lothiraldan any idea beyond extracting from the package what we need so that it's not a dependency anymore?

I would avoid using it altogether. I'm not sure how it would support new Python versions, alternative Python implementations (like Pypy) and I have high doubt it would be fast and effective.

That would likely means that instead of calling: `_get_conf(log_level, "info")` we will do `_get_conf(log_level, "log_level", "info")` which is a bit repetitive but not that much.

vict0rsch:
makes sense ; I just did

benoit-cty:
Done in https://github.com/mlco2/codecarbon/pull/187

PR #189, opened 2021-06-17
Routers refacto

SaboniAmine:
Emissions / Experiments / Projects refacto : 

- db, repository and service are injected into a container
- Starting from a DataForGood Team, Projects -> Experiments -> Run creation (with manual id chaining) is now valid. Should the API create default values if the users does not references correct ids ? Should part of this logic be handled in the package ?


benoit-cty (approved):

SaboniAmine:
Default values would be handled in a following PR (out of this one's scope as discussed with Benoit)

PR #190, opened 2021-06-18
Refacto Run

benoit-cty:
Make use of dependency injection design pattern for managing runs.

### DIFF
--- a/carbonserver/carbonserver/api/infra/repositories/repository_runs.py
+++ b/carbonserver/carbonserver/api/infra/repositories/repository_runs.py
@@ -4,7 +4,6 @@ from typing import List
 
 from dependency_injector.providers import Callable
 
-from carbonserver.api import schemas
 from carbonserver.api.domain.runs import Runs
 from carbonserver.api.schemas import Run, RunCreate
 from carbonserver.database.sql_models import Run as SqlModelRun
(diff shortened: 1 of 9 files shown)


carbonserver/carbonserver/api/infra/repositories/repository_runs.py
```diff
@@ -1,59 +1,146 @@
...
+                timestamp=run.timestamp,
+                experiment_id=run.experiment_id,
             )
+            session.add(db_run)
+            session.commit()
+            session.refresh(db_run)
+            return self.map_sql_to_schema(db_run)
+
+        # db_run = sql_models.Run(
+        #     timestamp=run.timestamp,
```
SaboniAmine, line 39: All commented code can be removed. If needed in the future we will rewrite it easier (and it will always be available in the history).

carbonserver/carbonserver/api/infra/repositories/repository_runs.py
```diff
@@ -1,59 +1,146 @@
...
-        return schemas.Run(
+        return Run(
             id=run.id,
             timestamp=run.timestamp,
             emission_id=run.emission_id,
             experiment_id=run.experiment_id,
         )
+
+
+class InMemoryRepository(Runs):
```
SaboniAmine, line 99: The InMemory Repo can be removed, as it will be hard to maintain now that we are making the logic evolve.
The tests on the router / service are now enough to validate that the domain interface / logic is respected by the repository.

carbonserver/carbonserver/api/routers/runs.py
```diff
@@ -1,21 +1,65 @@
...
+@router.get(
+    "/runs/",
+    tags=RUNS_ROUTER_TAGS,
+    status_code=status.HTTP_200_OK,
+)
+@inject
+def list_runs(
+    run_service: RunService = Depends(Provide[ServerContainer.run_service]),
+) -> List[Run]:
+    return run_service.list_run()
```
SaboniAmine, line 55: service method could be rename list_runs to match router / repo ?

SaboniAmine (changes requested):
This looks fine! Could you add automated tests on run router / service ? 
As it is basic crud, you can copy / replace tests from the teams service / router

carbonserver/carbonserver/api/infra/repositories/repository_runs.py
benoit-cty: OK, thanks

SaboniAmine (approved):

PR #191, opened 2021-06-22
add unused var

vict0rsch:


benoit-cty:
It was just a test to check CI.

PR #193, opened 2021-06-22
add copied/pasted example from dash documentation

vict0rsch:


benoit-cty:
Do we still need this PR ? It seems like it his an old branch of the dashboard that is now in `master`.

PR #196, opened 2021-06-24
API Functionnal test script

benoit-cty:
I add a script that call the API to test all the routers in a functional way : the API is tested as a blackbox in an operational way.

To use it:
- `docker-compose up -d`
- `cd carbonserver && tox -e integration`
- `cd .. && docker-compose down`

### DIFF
--- a/carbonserver/carbonserver/api/routers/authenticate.py
+++ b/carbonserver/carbonserver/api/routers/authenticate.py
@@ -27,5 +27,6 @@ def auth_user(
         return Token(access_token="a", token_type="access")
     else:
         raise HTTPException(
-            status_code=status.HTTP_401_UNAUTHORIZED, detail="Incorrect password!"
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Incorrect password or email!",
         )
--- a/carbonserver/tests/api/test_api.py
+++ b/carbonserver/tests/api/test_api.py
@@ -1,8 +1,12 @@
+import os
 import random
 
 import requests
 
-URL = "http://localhost:8008"
+# Get the API utl to use from an env variable if exist
+url = os.getenv("CODECARBON_API_URL")
+URL = url if url else "http://localhost:8008"
+
 experiment_id = project_id = user_id = api_key = org_id = team_id = email = None
 org_name = org_description = org_new_id = None
 team_name = team_description = team_new_id = emission_id = None
@@ -10,13 +14,26 @@ password = "Secret1!îstring"
 
 
 def is_key_value_exist(list_of_dict, key, value):
+    """
+    Check if at least one value of a key is equal to the specified value.
+    """
     for d in list_of_dict:
         if d[key] == value:
             return True
     return False
 
 
-def test_user_create():
+def is_key_all_values_equal(list_of_dict, key, value):
+    """
+    Check if all values of a key are equal to the specified value.
+    """
+    for d in list_of_dict:
+        if d[key] != value:
+            return False
+    return True
+
+
+def test_api_user_create():
     email = f"test-{random.randint(1, 20_000_000)}@test.com"
     payload = {"email": email, "name": "toto", "password": password}
     r = requests.post(url=URL + "/user", json=payload, timeout=2)
@@ -25,7 +42,7 @@ def test_user_create():
     assert r.json()["is_active"] == True  # noqa
 
 
-def test_user_signup():
+def test_api_user_signup():
     global user_id, api_key, org_id, team_id, email
     email = f"test-{random.randint(1, 20_000_000)}@test.com"
     payload = {"email": email, "name": "toto", "password": password}
@@ -39,19 +56,19 @@ def test_user_signup():
     team_id = r.json()["teams"][0]
 
 
-def test_users_list():
+def test_api_users_list():
     r = requests.get(url=URL + "/users", timeout=2)
     assert r.status_code == 200
     assert is_key_value_exist(r.json(), "id", user_id)
 
 
-def test_get_user():
+def test_api_get_user():
     r = requests.get(url=URL + "/user/" + user_id, timeout=2)
     assert r.status_code == 200
     assert r.json()["email"] == email
 
 
-def test_auth_success():
+def test_api_auth_success():
     payload = {"email": email, "password": password}
     r = requests.post(url=URL + "/authenticate/", json=payload, timeout=2)
     assert r.status_code == 200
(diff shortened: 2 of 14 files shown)


carbonserver/tests/api/test_api.py
```diff
@@ -0,0 +1,109 @@
+import random
+
+import requests
+
+URL = "http://localhost:8008"
+experiment_id = project_id = user_id = api_key = org_id = team_id = email = None
```
SaboniAmine, line 6: Proposal : execute this script from tox.ini [integration] env, with the injection of following parameters : 
- DB URL, which might be a shared dev/QA db hosted on clever
- ids, to let us clean the db after the test execution to make it idempotent
- Authorisations conditions, which might be injected later in this script, or might not, as we could separate it in a second integration test.
- Complex test cases. Test data can be handled in a separate file (using pytest fixture import functions) if it becomes difficult to read in this script.

carbonserver/tests/api/test_api.py
```diff
@@ -0,0 +1,109 @@
...
+    user_id = r.json()["id"]
+    api_key = r.json()["api_key"]
+    org_id = r.json()["organizations"][0]
+    team_id = r.json()["teams"][0]
+
+
+def test_auth_success():
+    payload = {"email": email, "password": password}
+    r = requests.post(url=URL + "/authenticate/", json=payload, timeout=2)
+    print(r.json())
```
SaboniAmine, line 28: we could log payloads for important tests to see payloads even if the test is green, but prints should be removed

carbonserver/carbonserver/api/routers/authenticate.py
```diff
@@ -25,5 +25,6 @@ def auth_user(
         return Token(access_token="a", token_type="access")
     else:
         raise HTTPException(
-            status_code=status.HTTP_401_UNAUTHORIZED, detail="Incorrect password!"
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Incorrect password or email!",
```
SaboniAmine, line 29: In the following PR about errors management, should this be extracted in 2 different messages ?

carbonserver/tests/api/test_api.py
```diff
@@ -0,0 +1,109 @@
...
+    assert r.status_code == 201
+    run_id = r.json()["id"]
+
+
+def test_create_emission():
+    global run_id
+    payload = {
+        "timestamp": "2021-04-04T08:43:00+02:00",
+        "run_id": run_id,
+        "duration": random.randint(1, 100),
```
SaboniAmine, line 102: Random in test can lead to complex debugging and should be controlled IMO (this case is still trivial tho). 
What is the intention behind those values ? Can they be fixed ?

carbonserver/carbonserver/api/routers/authenticate.py
benoit-cty: No, it is a security breach to tell if it is the password or the email that is wrong.
And it could be use to know if a person/email use the service.

carbonserver/tests/api/test_api.py
benoit-cty: Ok, I think only the email address need to be random to avoid error.

Thinking about this : we need endpoint to delete each entity. So we could delete all entries at the end of test.

SaboniAmine: A postgres test DB can be generated for tests using this interface : https://github.com/testcontainers/testcontainers-python

carbonserver/tests/api/test_api.py
benoit-cty: The script is run in Github Action with command `tox -e with_api_server`. DATABASE_URL is set by Github Action and CODECARBON_API_URL is set by Tox.

For "ids" we can't force the ids through the API.

When we have a delete entry we could delete all entries we add.

carbonserver/tests/api/test_api.py
benoit-cty: Locally I use "docker-compose up" to setup a DB.
In Github Action, it is set up with "Service containers".

benoit-cty:
Rebase with master. Update to new data in emission.
Ready to merge.

carbonserver/tests/api/test_api_black_box.py
```diff
@@ -0,0 +1,246 @@
...
+    """
+    for d in list_of_dict:
+        if d[key] != value:
+            return False
+    return True
+
+
+def test_api_user_create():
+    assert URL is not None
+    email = f"test-{random.randint(1, 20_000_000)}@test.com"
```
SaboniAmine, line 40: The random values will flood the local databases (and one day lead to a very confusing debugging session).
If you set fixed emails, you can execute delete operations from a pytest fixture, following this syntax : 
```python
from sqlalchemy import delete

USER_EMAIL = 'user@integration.test'

@pytest.fixture()
def del_user_after_test():
    """Fixture to destroy user"""
    # Setup: orm configuration

    yield # Test execution

    # Clean up user before ending test execution by pytest
    (
        delete(user_table).
        where(user_table.c.email == USER_EMAIL)
    )

def test_api_user_createt(del_user_after_test):
    """Test that creates user"""
    email = USER_EMAIL

    ...
```

carbonserver/tests/api/test_api_black_box.py
```diff
@@ -0,0 +1,246 @@
...
+def test_api_auth_success():
+    payload = {"email": email, "password": password}
+    r = requests.post(url=URL + "/authenticate/", json=payload, timeout=2)
+    assert r.status_code == 200
+    assert r.json()["access_token"] == "a"
+    assert r.json()["token_type"] == "access"
+
+
+def test_api_auth_fail():
+    payload = {"email": "toto@free.fr", "password": "password"}
```
SaboniAmine, line 83: This test needs a specific db setup / clean fixture

SaboniAmine (changes requested):
This is great, thanks!
We should keep integration tests as fast, this is really cool.
I've proposed you something to remove the test email randomness.

benoit-cty:
Need to move test script to `tests/integration/`

carbonserver/tests/api/test_api_black_box.py
```diff
@@ -0,0 +1,246 @@
...
+        "description": "Test to delete",
+        "team_id": team_new_id,
+    }
+    r = requests.post(url=URL + "/project/", json=payload, timeout=2)
+    assert r.status_code == 201
+    assert r.json()["team_id"] == team_new_id
+    project_id = r.json()["id"]
+
+
+@pytest.mark.xfail(reason="Not implemented yet")
```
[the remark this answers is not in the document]
benoit-cty: Change to "/projects/team/" + team_new_id that will exist after PR 227

carbonserver/tests/api/test_api_black_box.py
benoit-cty: No more random and delete of created user.
I suggest to implement other delete when we have a delete route because deleting directly into database is not real blackbox test.

SaboniAmine (approved):

PR #197, opened 2021-06-25
Harmonization of API routes

benoit-cty:
Small changes to harmonize the API routes.

### DIFF
--- a/carbonserver/carbonserver/api/routers/projects.py
+++ b/carbonserver/carbonserver/api/routers/projects.py
@@ -16,7 +16,7 @@ router = APIRouter(
 projects_temp_db = []
 
 
-@router.put("/projects", tags=PROJECTS_ROUTER_TAGS, status_code=status.HTTP_201_CREATED)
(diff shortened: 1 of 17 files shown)


carbonserver/carbonserver/api/routers/projects.py
```diff
@@ -16,7 +16,7 @@
 projects_temp_db = []
 
 
-@router.put("/projects", tags=PROJECTS_ROUTER_TAGS, status_code=status.HTTP_201_CREATED)
+@router.put("/project", tags=PROJECTS_ROUTER_TAGS, status_code=status.HTTP_201_CREATED)
```
SaboniAmine, line 19: Could we suffix with / all routers by default ?

SaboniAmine (approved):

carbonserver/carbonserver/api/routers/projects.py
benoit-cty: Trailing / removed to conform with "Rule #1: A trailing forward slash (/) should not be included in URIs" https://blog.restcase.com/7-rules-for-rest-api-uri-design/

PR #202, opened 2021-06-29
CLI for Codecarbon

benoit-cty:
A quick API CLI to create an experiment.

Run it with:
`pip install --editable . --upgrade && python codecarbon/cli/init.py`
And it will show you your personal experiment ID:
`Welcome to CodeCarbone, here is your experiment id : 3a202149-8be2-408c-a3d8-baeae2de2987`


vict0rsch:
@benoit-cty now the cli is registered with pip so this is enough:

```
$ codecarbon init
Welcome to CodeCarbon, here is your experiment id : 664bc448-ae0d-46a2-bcc4-851d32246623
```

and this creates the appropriate config file so if you run it again you'll get the same experiment id registered locally

+ some colors

2021-06-30 at 11 13

benoit-cty:
I've fixed the co2_signal test by removing the timeout parameter, but we still have a problem : `codecarbon.core.co2_signal.CO2SignalAPIError: Invalid authentication credentials`
Maybe we have to disable the test in the CI ? (How to do it ?)

vict0rsch:
maybe add `@pytest.mark.integ_test` as in https://github.com/mlco2/codecarbon/blob/master/tests/test_cpu.py#L12 since `pytest` is run without the `integ_test` marker https://github.com/mlco2/codecarbon/blob/master/tox.ini#L13

SaboniAmine:
Integration tests can be segregated in tox backend configuration (`carbonserver/tox.ini`). You just have to change the pytest path arg.
This configuration can be launched with `tox -e integration` command. We could add a conditional task in Github Actions & play it (with a more complex setup) for appropriate events.

benoit-cty:
I've added the decorator to disable test_get_emissions_TIMEOUT in CI and added a `tox -e all` to run all test ignoring the decorator if needed.

codecarbon/core/co2_signal.py
```diff
@@ -4,11 +4,10 @@
 from codecarbon.external.geography import GeoMetadata
 
 URL = "https://api.co2signal.com/v1/latest"
+CO2_SIGNAL_API_TIMEOUT = 30
```
vict0rsch, line 7: Is this not too large?

examples/api_call_debug.py
```diff
@@ -6,11 +6,9 @@
 
 
 @track_emissions(
-    project_name="just_sleep",
-    api_endpoint="http://app-d6bc59c3-0f69-4d8b-a6a3-bc39a9ceb0c2.cleverapps.io",
-    experiment_id="82ba0923-0713-4da1-9e57-cea70b460ee9",
-    measure_power_secs=10,
-    api_call_interval=2,
+    # api_endpoint="http://app-d6bc59c3-0f69-4d8b-a6a3-bc39a9ceb0c2.cleverapps.io",
```
vict0rsch, line 9: is this a voluntary comment?

vict0rsch (approved):
LGTM

PR #210, opened 2021-07-08
Separate CPU and GPU consumption in emissions output

### LINKED ISSUE #195: Separate CPU and GPU consumption in emissions output
* CodeCarbon version: 1.2.0
* Python version: 3.6
* Operating System: MAC

### Description

At the moment total energy contains the sum of all energy consumed by the CPU and GPU and it's the only measure that exist in the output results. It would be best to store the CPU and GPU consumption separately as well.

Minervabooks:
This change adds two new fields in Emission data : cpu_power and gpu_power both in the package and in the API.

Closes #195


benoit-cty:
Thank's for your great contribution. I've added some logs to help understanding what happened. You could activate debug by adding `log_level = DEBUG` to `.codecarbon.config`.

Can you:
- Use `pre-commit  run --all-files` before committing. It could be done automatically with `pre-commit  install`. That will format the code and do some checks.
- Add RAM energy, like CPU and GPU energy ("_total_cpu_power" is in fact "_total_cpu_energy" as you compute `self._total_cpu_power += energy`)
- Add the power of RAM,CPU and GPU (using `hardware.total_power().W`) so we have a number that will be independent of the api call frequency and the user could compare it to the computer specs.
- Add the new data to ApiClient to send them to the API, because I think you don't send them.
- Add instantaneous emission by computing emissions.emissions/emissions.duration*1000 g.CO2eq/s so we have a number that will be independent of the api call frequency.

Well I ask many things, sorry, maybe we could do it in more than one PR ;)

Minervabooks:
Hello @benoit-cty ,

Can you please review my latest commit on this PR? Specially how the ram/cpu/gpu power are being calculated. Thanks

benoit-cty:
Thanks a lot.
It works from package to API.
I've just made a change from `self._cpu_power += hardware.total_power()` to `self._cpu_power = hardware.total_power()` to do not sum up the power. And take W instead of KW.

I've added the emissions_rate to send the g.CO2eq/s to the API in addition to the cumulative kg.CO2.

Minervabooks:
@benoit-cty The test failed because it was missing the new added emissions_rate in the emissions_valid_headers.csv file. The tests now are successful.

benoit-cty (approved):
Thanks for fixing the test.
Looks good to me :+1: 

@vict0rsch I've keep the name "emissions" in the package for historical reason but use emissions_sum in the API.
What do you think of this PR ?

PR #215, opened 2021-07-23
Feature: Make output CSV file name configurable

jaketae:
This PR allows users to configure the file name of the output CSV (previously set to `emissions.csv`). 

## Example

```python
from code carbon import EmissionsTracker
tracker = EmissionsTracker(output_file="custom_name.csv")
```

## Summary

* This change is backwards compatible since the output file name still defaults to `emissions.csv`. 
* This feature can be useful in large-scale contexts  in which multiple experiments have to be run (otherwise each experiment would override each other's `emissions.csv`). 
* This feature originated from discussions between engineering and carbon footprint working groups in Hugging Face's [BigScience project](https://github.com/bigscience-workshop/bigscience).

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -158,9 +158,9 @@ class BaseEmissionsTracker(ABC):
         :param api_endpoint: Optional URL of Code Carbon API endpoint for sending
                              emissions data
         :param api_key: API key for Code Carbon API, mandatory to use it !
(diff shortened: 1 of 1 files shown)


benoit-cty (approved):
Thanks for contributing !

benoit-cty:
Thanks for your contribution, we will be happy to discuss with you of your use case and how the new Code Carbon API could help you.

PR #216, opened 2021-07-26
Fuzzy cpu matching

AlexisBogroff:
Match CPU model name with the CPU list using fuzzy matching, rather than strict equality.

More flexible, enables to match the right model even with:

- addition: _AMD Ryzen Threadripper 1950 16-Core Processor_ matches _AMD Ryzen Threadripper 1950_
- deletion: _AMD Ryzen 1950_ also matches _AMD Ryzen Threadripper 1950_

Still prevents from matching with a wrong model:
- _AMD Ryzen 3_ returns no matches, since it actually matches multiple models
- Greedy mode can force the selection of any model matched
- A threshold can be lowered to consider matches with a similarity ratio lower than 100%*

For additional info on similarity ratio see documentation of [the fuzzywuzzy package](https://github.com/seatgeek/fuzzywuzzy)

Limits:
- Multiple matches (with similarity ratio 100%) when single match would be expected. E.g. _AMD Ryzen 3 1200 PRO_ matches both _AMD Ryzen 3 1200_ and _AMD Ryzen 3 PRO 1200_ although only _AMD Ryzen 3 PRO 1200_ should be matched.

It can be compensated in some cases by forcing the match of any of these two models using greedy mode.

### DIFF
--- a/codecarbon/core/cpu.py
+++ b/codecarbon/core/cpu.py
@@ -7,8 +7,8 @@ import shutil
 import subprocess
 import sys
 import time
-from typing import Dict
-from fuzzywuzzy import process
+from typing import Dict, Union
+from fuzzywuzzy import fuzz
 
 import cpuinfo
 import pandas as pd
@@ -40,22 +40,6 @@ def is_rapl_available():
             exc_info=True,
         )
         return False
-
-
-def parse_cpu_model(raw_name) -> str:
-    """
-    Parse the model name from the raw name extracted from cpuinfo library
-    :return: parsed CPU name
-    """
-    if type(raw_name) == str:
-        model = (
-            raw_name.split(" @")[0]
-            .replace("(R)", "")
-            .replace("(TM)", "")
-            .replace(" CPU", "")
-        )
-        return model
-    return ""
 
 
 class IntelPowerGadget:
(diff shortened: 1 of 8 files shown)


tests/test_cpu.py
```diff
@@ -80,9 +81,108 @@ def test_intel_rapl(self):
...
+        tdp = TDP()
+        model = "Intel Core i7-8850H"
+        self.assertEqual(tdp._get_cpu_power_from_registry(model), 45)
+        model = "AMD Ryzen Threadripper 1950X"
+        self.assertEqual(tdp._get_cpu_power_from_registry(model), 180)
+        model = "AMD Ryzen Threadripper 1950X 16-Core Processor"
+        self.assertEqual(tdp._get_cpu_power_from_registry(model), 180)
+
+
+    def test_get_matching_cpu(self):
```
SaboniAmine, line 96: Could we add test cases on obvious non-matches ? fuzzy non-matches ?
To see where is the limit of the CPU parsing.

codecarbon/core/cpu.py
```diff
@@ -231,31 +211,142 @@ def get_cpu_details(self) -> Dict:
...
+    @staticmethod
+    def _get_single_direct_match(ratios:list, max_ratio:int, cpu_df:pd.DataFrame) -> str:
+        idx = ratios.index(max_ratio)
+        cpu_matched = cpu_df['Name'].iloc[idx]
+        return cpu_matched
 
-    def _get_power_from_constant(self) -> int:
+
+    def _get_matching_cpu(self, model_raw:str, cpu_df:pd.DataFrame,
+                          threshold_direct=100, threshold_token_set=100,
```
SaboniAmine, line 270: i guess the thresholds could be private vars, there is no need for them to be set outside of this function (this could lead to inconsistencies if a user choses to update those values and the match differs from its usual CPU).
Wdyt ?

tests/test_cpu.py
```diff
@@ -80,9 +81,108 @@ def test_intel_rapl(self):
...
+            None,
+        )
+
+        # /! LIMIT !\ The following matches with both:
+        # "AMD Ryzen 3 1200", and "AMD Ryzen 3 PRO 1200"
+        # However, it should only match with AMD Ryzen 3 PRO 1200
+        # (the words are in reversed order)
+
+        # model = "AMD Ryzen 3 1200 PRO"
+        # self.assertEqual(
```
SaboniAmine, line 178: this test should be uncommented as it exposes the matcher limit

SaboniAmine (changes requested):
Looks good to me, thanks !
I've added some changes to simplify the future maintenance

codecarbon/core/cpu.py
AlexisBogroff: Yes I agree! Thanks

AlexisBogroff: I'm now thinking on the contrary that we could set these variables as arguments in the main function. This could help the user force a match when it has not matched automatically. For example, one could set _greedy_ to True, or lower the _thresholds_ in order to force the match of a cpu. The user could then check by itself whether the cpu has a close tdp to its actual cpu.

SaboniAmine: I don't think it should be the user's responsibility to manage the cpu match, but if you do then your description seems a good way to do it.

SaboniAmine (approved):

codecarbon/core/cpu.py
AlexisBogroff: Ok I've left it as private vars!

tests/test_cpu.py
AlexisBogroff: I've added some fuzzy non-matches. Do you think we should find more of them?

PR #218, opened 2021-07-27
feat: add response model to API output

### LINKED ISSUE #205: API : Add a response-model to routes
In FastAPI we could add a BaseModel to API response: https://fastapi.tiangolo.com/tutorial/response-model/

This will give us:
- An output model check 
- A better docs because FastAPI will display the output model in /docs

SaboniAmine:
Adding response model to describe API outputs.
The objects returned by the API are defined in the schemas.py file, we do not suffix them as we identify the input objects with the Create suffix when needed. Otherwise the types can be shared between the application and its HTTP users.
Fix #205


benoit-cty (approved):
:+1:

PR #226, opened 2021-08-03
Api errors management

SaboniAmine:
This PR proposes a error handling method : 
- Raise a custom exception when a potential error can occur (as in the database session management in this first handler)
- Log the technical cause to enable troubleshooting, and format a human readable error message in the CustomException.error.message field
- Declare a custom exception handler : 
```python
async def custom_exception_handler(request: Request, exc: CustomException):
    return JSONResponse({"detail": exc.error.message}, status_code=400) # HTTP status might be customised too

```
- Inject in the application before injecting routers:  
```python
server.add_exception_handler(CustomException, custom_exception_handler)
```

### DIFF
--- a/carbonserver/carbonserver/api/infra/database/database_manager.py
+++ b/carbonserver/carbonserver/api/infra/database/database_manager.py
@@ -27,8 +27,6 @@ class Database:
     def create_database(self) -> None:
         Base.metadata.create_all(self._engine)
 
-    # Les exceptions de la base de données sont remontées ici, un switch case peut être mis en place pour remonter
-    # les exceptions connues pour la base de données
     @contextmanager
     def session(self) -> Callable[..., AbstractContextManager]:
         session: Session = self._session_factory()
@@ -37,23 +35,26 @@ class Database:
 
         except exc.IntegrityError as e:
             session.rollback()
+            logger.error(e.orig.args[0], exc_info=True)
             raise DBException(
-                error=DBError(code=DBErrorEnum.INTEGRITY_ERROR, message=e.orig.args[0])
+                error=DBError(code=DBErrorEnum.INTEGRITY_ERROR, message="Relation not found")
             )
         except exc.DataError as e:
             session.rollback()
+            logger.error(e.orig.args[0], exc_info=True)
             raise DBException(
-                error=DBError(code=DBErrorEnum.DATA_ERROR, message=e.orig.args[0])
+                error=DBError(code=DBErrorEnum.DATA_ERROR, message="Invalid data")
             )
         except exc.ProgrammingError as e:
             session.rollback()
+            logger.error(e.orig.args[0], exc_info=True)
             raise DBException(
                 error=DBError(
-                    code=DBErrorEnum.PROGRAMMING_ERROR, message=e.orig.args[0]
+                    code=DBErrorEnum.PROGRAMMING_ERROR, message="Wrong schema"
                 )
             )
         except Exception:
-            logger.exception("Session rollback because of exception")
+            logger.error("Session rollback because of exception", exc_info=True)
             session.rollback()
             raise
         finally:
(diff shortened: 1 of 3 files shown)


carbonserver/main.py
```diff
@@ -69,4 +83,4 @@ def init_server(container):
 
 @app.get("/")
 def default():
-    return {"status": "OK"}
+    raise Exception
```
Minervabooks, line 86: This returns Generic Error while calling "localhost:8008/", shouldn't be changed to something like Not Found?

carbonserver/carbonserver/api/infra/database/database_manager.py
```diff
@@ -37,23 +35,28 @@ def session(self) -> Callable[..., AbstractContextManager]:
...
-                error=DBError(code=DBErrorEnum.DATA_ERROR, message=e.orig.args[0])
+                error=DBError(code=DBErrorEnum.DATA_ERROR, message="Invalid data")
             )
         except exc.ProgrammingError as e:
             session.rollback()
+            logger.error(e.orig.args[0], exc_info=True)
             raise DBException(
                 error=DBError(
-                    code=DBErrorEnum.PROGRAMMING_ERROR, message=e.orig.args[0]
+                    code=DBErrorEnum.PROGRAMMING_ERROR, message="Wrong schema"
```
Minervabooks, line 55: How can I test this? Is there a particular scenario where this exception is thrown?

carbonserver/carbonserver/api/infra/database/database_manager.py
```diff
@@ -37,23 +35,28 @@ def session(self) -> Callable[..., AbstractContextManager]:
 
         except exc.IntegrityError as e:
             session.rollback()
+            logger.error(e.orig.args[0], exc_info=True)
             raise DBException(
-                error=DBError(code=DBErrorEnum.INTEGRITY_ERROR, message=e.orig.args[0])
+                error=DBError(
+                    code=DBErrorEnum.INTEGRITY_ERROR, message="Relation not found"
```
Minervabooks, line 41: How can I test this? Is there a particular scenario where this exception is thrown?

Minervabooks (commented):
Is there a way to also catch errors when the POST request fail because of wrong body?

At the moment it only shows Unprocessable Entity log?

carbonserver/carbonserver/api/infra/database/database_manager.py
SaboniAmine: The http validation layer should avoid this to happen if we keep pydantic schemes up to date.
I guess you can try a custom query in a new repository method which tries to insert an incorrect object.

carbonserver/carbonserver/api/infra/database/database_manager.py
SaboniAmine: Trying to create a run with an inexistant project id should throw this

SaboniAmine:
For validation errors we could override the default validation handler and implement a custom response / header.
Currently the returned payload in case of incorrect POST is : 
```
{
  "detail": [
    {
      "loc": [
        "body",
        "organization_id"
      ],
      "msg": "value is not a valid uuid",
      "type": "type_error.uuid"
    }
  ]
}
```
Would you see something to add ?

carbonserver/main.py
SaboniAmine: This is a mistake I will come back to a default response. Would you see something better than {"status": "OK"} ?

carbonserver/carbonserver/api/infra/database/database_manager.py
benoit-cty: I've tested it by adding a not existing field in sql_model.

benoit-cty (approved):
Really nice to have comprehensive error message.

SaboniAmine:
The counterpart to human readable being that we now have to handle manually exceptions to transmit the correctly formatted message to the HTTP layer otherwise they will fall in the Generic error handling.
Thanks for the review!

PR #236, opened 2021-08-21
Method to flush emission to CSV

benoit-cty:
Addressing #217 after talking with @stas00 (see https://github.com/mlco2/codecarbon/pull/235)  I add a new method `flush()` that compute emissions and add them to the CSV, and/or call the API if you use it.
I use it in a Keras callback and it do the job of adding a line in the CSV after each epoch.
You can see it in `mnist_callback.py`

I've added documentation. **But no unit test.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,16 @@ with EmissionsTracker() as tracker:
     # GPU Intensive code goes here
 ```
 
+### Flush data when running
 
+By default, Code Carbon only write the CVS output file when it stop.
+But for long run you may want to get intermediate data.
+It is possible to call the flush() methode to do so.
+
+Have a look to [./examples/mnist_callback.py](./examples/mnist_callback.py) for an example.
+
+Note that if you use the API it will also call it when you call flush().
+You could set *api_call_interval* to -1, so that it will not be called automatically and then force a call at the end of each epoch to get the emission of each epoch.
 
 ### Using comet.ml
 
--- a/codecarbon/core/api_client.py
+++ b/codecarbon/core/api_client.py
@@ -52,14 +52,19 @@ class ApiClient:  # (AsyncClient)
         self._previous_call = time.time()
         if self.run_id is None:
             # TODO : raise an Exception ?
-            logger.error(
-                "add_emissionadd_emission need a run_id : the initial call may "
+            logger.debug(
+                "ApiClient.add_emission need a run_id : the initial call may "
                 + "have failed. Retrying..."
             )
             self._create_run(self.experiment_id)
+            if self.run_id is None:
+                logger.error(
+                    "ApiClient.add_emission still no run_id, arborting for this time !"
+                )
+            return False
         if carbon_emission["duration"] < 1:
             logger.warning(
-                "Warning : emission not send because of a duration smaller than 1."
+                "ApiClient : emissions not send because of a duration smaller than 1."
             )
             return False
         emission = EmissionCreate(
@@ -80,10 +85,10 @@ class ApiClient:  # (AsyncClient)
             payload = dataclasses.asdict(emission)
             url = self.url + "/emission"
             r = requests.post(url=url, json=payload, timeout=2)
-            logger.debug(f"Successful upload emission {payload} to {url}")
             if r.status_code != 201:
                 self._log_error(url, payload, r)
                 return False
+            logger.debug(f"ApiClient - Successful upload emission {payload} to {url}")
         except Exception as e:
             logger.error(e, exc_info=True)
             return False
@@ -96,7 +101,7 @@ class ApiClient:  # (AsyncClient)
         """
         if self.experiment_id is None:
             # TODO : raise an Exception ?
-            logger.error("FATAL The API _create_run need an experiment_id !")
+            logger.error("ApiClient FATAL The API _create_run need an experiment_id !")
             return None
         try:
             run = RunCreate(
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -149,7 +149,16 @@ with EmissionsTracker() as tracker:
     # GPU Intensive code goes here
 ```
 
+### Flush data when running
 
+By default, Code Carbon only write the CVS output file when it stop.
+But for long run you may want to get intermediate data.
+It is possible to call the flush() methode to do so.
```
stas00, line 156:
```suggestion
By default, Code Carbon only writes the CVS output file when it stops.
But for a long run you may want to save intermediate data.
It is possible to call the flush() method to do so.
```

examples/mnist_callback.py
```diff
@@ -0,0 +1,45 @@
+import tensorflow as tf
+from tensorflow.keras.callbacks import Callback
+
+from codecarbon import EmissionsTracker
+
+"""
+This sample code show how to use CodeCarbon as a Keras Callback
+to save emissions after each epoch.
```
stas00, line 8:
```suggestion
This sample code shows how to use CodeCarbon as a Keras Callback
to save emissions after each epoch.
```

examples/README.md
```diff
@@ -9,6 +9,7 @@ pip install -r requirements-examples.txt
 ## Examples
 * [mnist.py](mnist.py): Usage using explicit `CO2Tracker` objects.
 * [mnist_decorator.py](mnist_decorator.py): Using the `@track_co2` decorator.
+* [mnist_callback.py](mnist_callback.py): Using Keras callbacks to save emission after each epoch.
```
stas00, line 12:
```suggestion
* [mnist_callback.py](mnist_callback.py): Using Keras callbacks to save emissions after each epoch.
```

stas00 (commented):
Thank you for adding this new API, @benoit-cty 

I added a few language improvement suggestions.

benoit-cty:
Thanks for the language improvement !
I've added unit tests so it's ready for review.

SaboniAmine (approved):

codecarbon/core/api_client.py
```diff
@@ -52,14 +52,19 @@ def add_emission(self, carbon_emission: dict):
...
             self._create_run(self.experiment_id)
+            if self.run_id is None:
+                logger.error(
+                    "ApiClient.add_emission still no run_id, arborting for this time !"
+                )
+            return False
         if carbon_emission["duration"] < 1:
             logger.warning(
-                "Warning : emission not send because of a duration smaller than 1."
+                "ApiClient : emissions not send because of a duration smaller than 1."
```
stas00, line 67:
```suggestion
                "ApiClient : emissions not sent because of a duration smaller than 1."
```

codecarbon/core/api_client.py
```diff
@@ -52,14 +52,19 @@ def add_emission(self, carbon_emission: dict):
...
-            logger.error(
-                "add_emissionadd_emission need a run_id : the initial call may "
+            logger.debug(
+                "ApiClient.add_emission need a run_id : the initial call may "
                 + "have failed. Retrying..."
             )
             self._create_run(self.experiment_id)
+            if self.run_id is None:
+                logger.error(
+                    "ApiClient.add_emission still no run_id, arborting for this time !"
```
stas00, line 62:
```suggestion
                    "ApiClient.add_emission still no run_id, aborting for this time !"
```

codecarbon/core/api_client.py
```diff
@@ -96,7 +101,7 @@ def _create_run(self, experiment_id):
         """
         if self.experiment_id is None:
             # TODO : raise an Exception ?
-            logger.error("FATAL The API _create_run need an experiment_id !")
+            logger.error("ApiClient FATAL The API _create_run need an experiment_id !")
```
stas00, line 104:
```suggestion
            logger.error("ApiClient FATAL The API _create_run needs an experiment_id !")
```

codecarbon/emissions_tracker.py
```diff
@@ -301,6 +306,29 @@ def start(self) -> None:
         self._last_measured_time = self._start_time = time.time()
         self._scheduler.start()
 
+    @suppress(Exception)
+    def flush(self) -> Optional[float]:
+        """
+        Write emission to disk or call the API depending of the configuration
```
stas00, line 312:
```suggestion
        Write emission to disk or call the API depending on the configuration
```

stas00 (approved):
Tested this last incarnation to work correctly for our needs

Added a few tiny language corrections.

vict0rsch:
Added `on_csv_write: Optional[str]`:

```
:param on_csv_write: "append" or "update". Whether to always append a new line
                        to the csv when writing or to update the existing `run_id`
                        row (useful when calling`tracker.flush()` manually).
                        Accepts one of "append" or "update".
```

codecarbon/output.py
```diff
@@ -81,7 +81,13 @@ class FileOutput(BaseOutput):
...
     """
 
-    def __init__(self, save_file_path: str):
+    def __init__(self, save_file_path: str, on_flush: str = "append"):
+        if on_flush not in {"append", "update"}:
+            raise ValueError(
+                f"Unknown `on_flush` value: {on_flush}"
+                + " (should be one of 'append' or 'update'"
+            )
+        self.on_flush: str = on_flush
```
benoit-cty, line 90: Isn'it better to name the parameter "on_csv_write" also for harmonization ?
And someone could want to use _update_ with flush().

vict0rsch: agreed for `on_csv_write`

I don't understand what you mean for the second part however

SaboniAmine: I guess Benoît suggests that a flush action could lead to update values already stored in the csv file.
I however doubt that this would be better than append-only mode, as it could lead to desynchronization issues if the update fails.

benoit-cty (commented):
Nice additions Victor :+1:

Issue #232, opened 2021-08-13
root only perms file-related crash: /sys/class/powercap/intel-rapl/intel-rapl:0/energy_uj


stas00:
* CodeCarbon version: master
* Python version: 3.8
* Operating System: Ubuntu 20.04

### Description

```
[codecarbon INFO @ 15:19:53] Unable to read Intel RAPL files at [RAPLFile(name='Processor Power_0(Watt)', path='/sys/class/powercap/intel-rapl/intel-rapl:0/energy_uj', energy_reading=Energy(kwh=0), power_measurement=0)]
                 Exception occurred [Errno 13] Permission denied: '/sys/class/powercap/intel-rapl/intel-rapl:0/energy_uj'
Traceback (most recent call last):
  File "/mnt/nvme1/code/huggingface/codecarbon-bs/codecarbon/core/cpu.py", line 200, in get_cpu_details
    list(map(lambda rapl_file: rapl_file.start(), self._rapl_files))
  File "/mnt/nvme1/code/huggingface/codecarbon-bs/codecarbon/core/cpu.py", line 200, in <lambda>
    list(map(lambda rapl_file: rapl_file.start(), self._rapl_files))
  File "/mnt/nvme1/code/huggingface/codecarbon-bs/codecarbon/core/rapl.py", line 22, in start
    self.energy_reading = self._get_value()
  File "/mnt/nvme1/code/huggingface/codecarbon-bs/codecarbon/core/rapl.py", line 17, in _get_value
    with open(self.path, "r") as f:
PermissionError: [Errno 13] Permission denied: '/sys/class/powercap/intel-rapl/intel-rapl:0/energy_uj'
```

On another system this file is readable by users:
```
l /sys/class/powercap/intel-rapl/intel-rapl:0/energy_uj
-r--r--r-- 1 root root 4.0K Jul 20 10:06 /sys/class/powercap/intel-rapl/intel-rapl:0/energy_uj
```

On mine it's:
```
l /sys/class/powercap/intel-rapl/intel-rapl:0/energy_uj
-r-------- 1 root root 4.0K Jul 24 15:06 /sys/class/powercap/intel-rapl/intel-rapl:0/energy_uj
```

I can manually fix it, but on the next reboot it goes back to root-only `r` perms.

Here is a wider net:
```
l /sys/class/powercap/intel-rapl/intel-rapl:0/
total 0
-r-------- 1 root root 4.0K Jul 24 15:10 constraint_0_max_power_uw
[...]

PR #239, opened 2021-08-26
Check RAPL file permission at launch

benoit-cty:
Following https://github.com/mlco2/codecarbon/issues/232 I propose an improvement to check permission on RAPL file to avoid having an error on each measurement and using the fallback.

This will not fix the issue, but we are not responsible of Ubuntu decision of rapl file permission.

### DIFF
--- a/codecarbon/core/cpu.py
+++ b/codecarbon/core/cpu.py
@@ -192,9 +192,15 @@ class IntelRAPL:
                 if "package" in name:
                     name = f"Processor Power_{i}(Watt)"
                     i += 1
-                self._rapl_files.append(
-                    RAPLFile(name, os.path.join(self._lin_rapl_dir, file, "energy_uj"))
-                )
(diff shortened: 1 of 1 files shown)


stas00:
Why the obvious error is logged as a warning? That would lead to incorrect results and the user won't be the wiser. How can a user know that some file having the wrong perms warning is important at all?

Also while at it why not give a user a suggestion on how to fix it?

benoit-cty:
Thanks @stas00 for challenging us and keep using Code Carbon despite all difficulty.
What do you think of:
```
                    logger.error(
                        "Unable to read Intel RAPL files for CPU power, we will use a constant for your CPU power."
                        + " Please view https://github.com/mlco2/codecarbon/issues/232#issuecomment-910394964"
                        + f" for workaround : {e}"
                    )
```

stas00:
Looks perfect to me, @benoit-cty! Thank you!

Issue #238, opened 2021-08-25
WARNING:apscheduler.scheduler:Execution of job "BaseEmissionsTracker._measure_power 


stas00:
@benoit-cty - Can we hold off on 2.0 release?

I'm now testing on the live multi-node setup and I'm getting bombarded by:
```
WARNING:apscheduler.scheduler:Execution of job "BaseEmissionsTracker._measure_power (trigger: interval[0:00:15], next run at: 2021-08-25 23:09:23 CEST)" skipped: maximum number of running instances reached (1)
WARNING:apscheduler.scheduler:Execution of job "BaseEmissionsTracker._measure_power (trigger: interval[0:00:15], next run at: 2021-08-25 23:09:38 CEST)" skipped: maximum number of running instances reached (1)
WARNING:apscheduler.scheduler:Execution of job "BaseEmissionsTracker._measure_power (trigger: interval[0:00:15], next run at: 2021-08-25 23:09:53 CEST)" skipped: maximum number of running instances reached (1)
WARNING:apscheduler.scheduler:Execution of job "BaseEmissionsTracker._measure_power (trigger: interval[0:00:15], next run at: 2021-08-25 23:10:08 CEST)" skipped: maximum number of running instances reached (1)
WARNING:apscheduler.scheduler:Execution of job "BaseEmissionsTracker._measure_power (trigger: interval[0:00:15], next run at: 2021-08-25 23:10:23 CEST)" skipped: maximum number of running instances reached (1)
WARNING:apscheduler.scheduler:Execution of job "BaseEmissionsTracker._measure_power (trigger: interval[0:00:15], next run at: 2021-08-25 23:10:38 CEST)" skipped: maximum number of running instances reached (1)
WARNING:apscheduler.scheduler:Execution of job "BaseEmissionsTracker._measure_power (trigger: interval[0:00:15], next run at: 2021-08-25 23:10:53 CEST)" skipped: maximum number of running instances reached (1)
WARNING:apscheduler.scheduler:Execution of job "BaseEmissionsTracker._measure_power (trigger: interval[0:00:15], next run at: 2021-08-25 23:11:08 CEST)" skipped: maximum number of running instances reached (1)
```
every few secs.

What is wrong and how do I fix this?

this problem wasn't there when I was last testing your PR on my machine (a single node), so perhaps this is something specific to the target HPC.

You can see how the tracker has been initialized here:

[...]

PR #241, opened 2021-08-26
add logger preamble and misfire_grace_time

vict0rsch:
Addressing #238 

```python
In [5]: e = EmissionsTracker(logger_preamble=" [Node1]")
[codecarbon INFO @ 15:52:23] [Node1] Tracking Intel CPU via Power Gadget
```

and 

```
        self._scheduler.add_job(
            self._measure_power,
            "interval",
            seconds=self._measure_power_secs,
            max_instances=1,
            misfire_grace_time=self._misfire_grace_time,
        )
```

### DIFF
--- a/codecarbon/external/logger.py
+++ b/codecarbon/external/logger.py
@@ -1,18 +1,48 @@
 import logging
 import os
+from typing import Optional
+
+
+def set_logger_format(custom_preamble: Optional[str] = ""):
+    logger = logging.getLogger("codecarbon")
+    format = f"[%(name)s %(levelname)s @ %(asctime)s]{custom_preamble} %(message)s"
+    formatter = logging.Formatter(format, datefmt="%H:%M:%S")
+    handler = logging.StreamHandler()
+    handler.setFormatter(formatter)
+
+    if logger.hasHandlers():
+        logger.handlers.clear()
+
+    logger.addHandler(handler)
+
+
+def set_logger_level(level: Optional[str] = None):
+
+    if level is None:
+        lower_envs = {k.lower(): v for k, v in os.environ.items()}
+        level = lower_envs.get("codecarbon_log_level", "INFO")
+
+    level = level.upper()
+    known_levels = {
+        "CRITICAL",
+        "FATAL",
+        "ERROR",
+        "WARN",
+        "WARNING",
+        "INFO",
+        "DEBUG",
+        "NOTSET",
+    }
+
+    logger = logging.getLogger("codecarbon")
+
+    if level not in known_levels:
+        logger.error(f"Unknown log level: {level}. Doing nothing.")
+        return
+    logger.setLevel(level)
+
+
+set_logger_format()
+set_logger_level()
 
 logger = logging.getLogger("codecarbon")
-formatter = logging.Formatter(
-    "[%(name)s %(levelname)s @ %(asctime)s] %(message)s",
-    datefmt="%H:%M:%S",
-)
-handler = logging.StreamHandler()
-handler.setFormatter(formatter)
-if not logger.handlers:
-    logger.addHandler(handler)
-env_level = os.environ.get("CODECARBON_LOG_LEVEL")
-if env_level is None:
-    env_level = os.environ.get("codecarbon_log_level")
-if env_level is None:
-    env_level = "INFO"
-logger.setLevel(level=env_level)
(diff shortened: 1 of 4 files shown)


stas00:
So I did several experiments:

- measure_power_secs=30 still warns, but much less often:
```
Run time of job "BaseEmissionsTracker._measure_power (trigger: interval[0:00:30], next run at: 2021-08-28 02:00:13 CEST)" was missed by 0:00:03.089666
```

- `max_instances=3` seems to not generate warnings, at least in the few experiments I have run - but this one requires hacking the source code, so can we make it an argument to the base Tracker class?

stas00:
oh, just noticed another one:
```
[codecarbon DEBUG @ 02:22:45] r007 Exception occurred while instantiating IntelPowerGadget : Platform not supported by Intel Power Gadget
  File "/gpfsssd/worksf/projects/rech/six/ura81os/code/codecarbon/codecarbon/core/cpu.py", line 27, in is_powergadget_available
    IntelPowerGadget()
  File "/gpfsssd/worksf/projects/rech/six/ura81os/code/codecarbon/codecarbon/core/cpu.py", line 66, in __init__
    self._setup_cli()
  File "/gpfsssd/worksf/projects/rech/six/ura81os/code/codecarbon/codecarbon/core/cpu.py", line 93, in _setup_cli
    raise SystemError("Platform not supported by Intel Power Gadget")
SystemError: Platform not supported by Intel Power Gadget
```
early on. Not sure if it was always there. I think it only emerged when debug was turned on, so probably it's ok.

stas00:
Not sure if relativism kicks in when one puts the magnifying glass on, but I can't get any runtime warnings about misfire if I enable debug - perhaps it slows everything down long enough for everything to succeed.

I've attached the log file (just the carbon logs) with the debug dump if you will find any good use for it.

[carbon.log.txt](https://github.com/mlco2/codecarbon/files/7069882/carbon.log.txt)

benoit-cty:
For IntelPowerGadget, yes it's due to debug. It just mean you are not in Windows, we have to made this look less scary.

Your log looks good. If your GPU usage is quite constant you may increase measure_power_secs to took measures every 300 s instead of 15s.

But all your machines measures are in a window of 2 seconds so it won't improve anything, just reduce log size.

stas00:
Great, will use a much longer `measure_power_secs` - I suppose on `stop` it will still take a full measure even if it's before  `measure_power_secs`, correct?

Also, a slightly unrelated question - do we need to do any exception handling on behalf of CC? It's an optional feature and must absolutely not crash the main training should something go wrong.

Thank you!

codecarbon/external/logger.py
```diff
@@ -1,18 +1,61 @@
 import logging
 import os
+from typing import Optional
+
+
+def set_logger_format(custom_preamble: Optional[str] = ""):
+    logger = logging.getLogger("codecarbon")
+    format = f"[%(name)s %(levelname)s @ %(asctime)s]{custom_preamble} %(message)s"
```
benoit-cty, line 8: Space or brackets could be added here for custom_preamble..

benoit-cty (approved):

benoit-cty:
Yes, on stop or flush we compute power based on the time between last measure.
The only drawback is that you probably stop your training before stopping CC, so CC will not see a realistic power consumption on stop : for CPU we have a real Joules consumption but for GPU we only have the instant power consumption at the exact time of measurement. So the measure_power_secs must not be too big as our measurement is not rocket science...

For crash we have `@suppress(Exception)` to avoid that on start / stop / flush, but @vict0rsch maybe we have to add it on `_measure_power` or it's already handled by the scheduler ? Because I saw we could `raise ValueError` in `_measure_power`

stas00:
We call `cc.stop` at the very end of the program, so after training has completed. But there are some cases where it `sys.exit()` on a different path. can probably try and catch those cases too. But since it saves the checkpoint after finishing the training, this is probably just as good.

codecarbon/external/logger.py
vict0rsch: sold, adding brackets

vict0rsch:
Summary of latests commits:

* `max_instances` argument for the scheduler's jobs
* preamble inside brackets (`[codecarbon INFO @ 02:22:45][preamble] ...`)
* no more `ValueError` in `_measure_power()`, rather a `logger.error`
* clean up the PowerGadget & RAPL `debug` logs so they look more normal

benoit-cty (approved):

stas00:
could this be merged, please?

stas00:
Thank you, @vict0rsch!

PR #243, opened 2021-09-01
Doc on deploy

benoit-cty:
This small PR describe how to deploy on Clever Cloud and add a missing dependency in the requirements.txt used by Clever-Cloud. requirements.txt is mandatory for Clever Cloud.

What do you think of removing requirements-dev.txt and requirements-test.txt as they are the same for now.
We could thing of a separation later.
Or we keep requirements-test.txt with only tox in it ?


SaboniAmine (approved):
We could remove the requirements-dev list I guess, as test / dev envs are the same for the moment.
We can recreate it if needed later.
Thanks for the documentation ! Which other elements would you like to see in the contributing ?

Issue #117, opened 2021-02-22
Issue with rapl energy on Linux - value seems wrong by a 100x factor


PierreRust:
* CodeCarbon version: codecarbon==1.0.0
* Python version: 3.8.3
* Operating System: Linux ubuntu 20.04 LTS

### Description

I've been measuring the energy consumption of a CPU-only inference (7 min long) using several tools : 
* powerAPI / smartwatt : powerapi.org/
* pyJoule : https://pyjoules.readthedocs.io/en/latest/index.html
* codecarbon

Codecarbon gives wildly different results : 2.15 Wh for powerAPI, 2.21 Wh for pyJoules and 0.035 Wh for codecarbon. Given the duration of the test and the processor in my test computer, I believe powerAPI and pyJoule to be right and codecarbon to have an issue.
I had a quick look on how codecarbon measure energy on linux, in `IntelRAPL.get_cpu_details` ; if I understand it correctly, you calculate the power of the cpu by reading the energy consumed during 10 milliseconds (and of course dividing it then by 10ms, to get an average power during these 10ms)  and then use that value to compute the energy for the tracker interval (15s by default). This seems to lead to under-estimating the global energy usage, there is no reason to be sure that the average power these 10ms is the same than the average power during the 10s interval. 
Besides, as the rapl counter simply counts joules, it would be easier and more accurate to look at this counter simply at the beginning and and of the 10s interval to get a measure of the energy used by the cpu during that interval.

Is there a reason for this process or have I misunderstood the implementation ? 

I have not tried codecarbon on Windows so I can not say if the reported value has the same issue.

vict0rsch:
Thanks for the feedback and sorry for the time it took for us to get back to you. It's a good suggestion, we'll improve.

benoit-cty:
Improvement has been made, do we still have the problem ?

PR #253, opened 2021-09-30
[RAPL] Measure energy directly

mathilde-leval:
Related issue: https://github.com/mlco2/codecarbon/issues/117

For CPUs with RAPL files, the energy is directly available. This is a proposal to use these energy values instead of estimating the average power and converting it back to energy.


hb9eik:
Is seems that the power is reported 1000x lesser than real, Please see issue #286

PR #265, opened 2021-11-01
Test RAPL energy measure

mathilde-leval:
Adds a test for the method `measure_power_and_energy` for RAPL CPUs


benoit-cty (approved):
Nice addition, thank you.

PR #269, opened 2021-11-04
New energy mix

benoit-cty:
- [x] Get electricity energy mix data data from https://www.globalpetrolprices.com/energy_mix.php
- [x] Add country ISO code Alpha 3 using https://raw.githubusercontent.com/datasets/country-codes/master/data/country-codes.csv (and fix not matching names)
 - [x] Export the last complete year in JSON
- [x] Change package code to use it

Drawbacks:
- We loose USA and Canada region specific mix.
- We have "fossil", not "coal/fuel/gaz", according to https://www.eia.gov/tools/faqs/faq.php?id=73&t=11, here is the 
Pounds of CO2 emitted per million British thermal units (Btu) of energy for various fuels:

Coal (anthracite) | 228.60
-- | --
Coal (bituminous) | 205.40
Coal (lignite) | 216.24
Coal (subbituminous) | 214.13
Diesel fuel and heating oil | 163.45
Gasoline (without ethanol) | 155.77
Propane | 138.63
Natural gas | 116.65


benoit-cty:
The value from https://www.epa.gov/egrid/data-explorer of 1 401 lb/kWh equal to 635 g eq CO2 / kWh that seems low for fossil.

I've added the carbon intensity for UE countries from https://ourworldindata.org/grapher/carbon-intensity-electricity
So we do not have to compute it from energy mix for them. See https://github.com/mlco2/codecarbon/blob/new_energy_mix/codecarbon/data/private_infra/03_add_eu_data.ipynb for more info.

benoit-cty:
I've added:
- Fallback if we did not know the country
- Put the carbon intensity of the various energies in a JSON
- More generic code to compute carbon intensity
- Unit test

codecarbon/core/emissions.py
```diff
@@ -157,58 +157,72 @@ def get_country_emissions(self, energy: Energy, geo: GeoMetadata) -> float:
...
         )
+        carbon_intensity = 0
+        energy_sum = energy_mix["total_TWh"]
+        energy_sum_computed = 0
+        # Iterate through each source of energy in the country
+        for energy_type, energy_per_year in energy_mix.items():
+            if "_TWh" in energy_type:
+                # Compute the carbon intensity ratio of this source for this country
+                carbon_intensity_for_type = carbon_intensity_per_source.get(
+                    energy_type[: -len("_Twh")]
```
mathilde-leval, line 211:
```suggestion
                    energy_type[: -len("_TWh")]
```
Just nitpicking

mathilde-leval (approved):

vict0rsch (approved):

PR #279, opened 2022-02-02
Add server-monitor

### LINKED ISSUE #278: Add an easy way to monitor a server
Code Carbon could monitor a whole server, without consideration about the type of processes running on it.

So I think it could be useful to have an easy way to do it, for example it must be as easy as:
```
>pip install codecarbon
>codecarbon init
>codecarbon monitor
```
We could use the example `api_call_debug.py` and package it in a command line.

benoit-cty:
Add an option to the cli to monitor a machine it makes it as easy as:
```
pip install codecarbon
codecarbon init
codecarbon monitor
```
Don't need any single line of code !

Will close #278

### DIFF
--- a/codecarbon/cli/main.py
+++ b/codecarbon/cli/main.py
@@ -1,4 +1,5 @@
 import click
+import time
 
 from codecarbon.cli.cli_utils import (
     get_api_endpoint,
@@ -7,6 +8,7 @@ from codecarbon.cli.cli_utils import (
 )
 from codecarbon.core.api_client import ApiClient, get_datetime_with_timezone
 from codecarbon.core.schemas import ExperimentCreate
+from codecarbon import EmissionsTracker
 
 DEFAULT_PROJECT_ID = "e60afa92-17b7-4720-91a0-1ae91e409ba1"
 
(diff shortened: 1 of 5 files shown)


vict0rsch:
Good idea. What about `codecarbon monitor N` and use `N` in `sleep(N)`?

benoit-cty:
> Good idea. What about `codecarbon monitor N` and use `N` in `sleep(N)`?

I don't see the point for sleep but I've added options:

```
~/codecarbon$ codecarbon monitor --help
Usage: codecarbon monitor [OPTIONS]

Options:
  --measure_power_secs INTEGER
  --api_call_interval INTEGER
  --api / --no-api
  --help                        Show this message and exit.
```

codecarbon/cli/main.py
```diff
@@ -53,3 +56,25 @@ def init():
...
+@codecarbon.command(
+    "monitor", short_help="Run an infinite loop to monitor this machine."
+)
+@click.option("--measure_power_secs", default=10)
+@click.option("--api_call_interval", default=30)
+@click.option("--api/--no-api", default=True)
+def monitor(measure_power_secs, api_call_interval, api):
+    experiment_id = get_existing_local_exp_id()
+    if api and experiment_id is None:
+        click.echo("ERROR: No experiment id, call with init option first.")
```
vict0rsch, line 70: Maybe be more explicit here: "call `codecarbon init` first"

vict0rsch (approved):

PR #280, opened 2022-02-08
Fix emissions rate logging

vict0rsch:


codecarbon/output.py
```diff
@@ -59,16 +59,10 @@ class EmissionsData:
     def values(self) -> OrderedDict:
         return OrderedDict(self.__dict__.items())
 
-    def substract_in_place(self, previous_emission):
```
benoit-cty, line 62: Delete unused code.

benoit-cty (approved):
Thanks !

PR #281, opened 2022-02-08
Package only for master

benoit-cty:
I suggest that we don't package at every push to limit the carbon emission of Code Carbon :wink: 

Only package when we merge to master.

Bonus : I've also removed duplicate run, to save more carbon and time !

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -3,9 +3,9 @@ name: package
 on: [push, pull_request]
 
 jobs:
-    build:
+    build-package:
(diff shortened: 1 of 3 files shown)


.github/workflows/package.yml
```diff
@@ -1,38 +1,39 @@
 name: package
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [master]
 
 jobs:
-    build:
```
benoit-cty, line 6: I just removed too much spaces.

vict0rsch (approved):

Issue #220, opened 2021-07-29
Release 1.3


SaboniAmine:
After the recent work performed both on the package & the API, it may be the time to release a new version of the package ?

Main features addition : 
- RAM consumption handling 
- CPU / GPU / RAM segregation in the output
- Documentation enhancement
- Data persistence API alpha version (which may need a propre documentation on usage / contribution)

Fixes : 
- RAPL units conversion 
- Package configuration management 

Would you see anything to add to the next release ?

benoit-cty:
Do we have to wait to have a website to create an API account ?
Or do we made a first release to collect info from users on what to put in this website ?
Or do we made a simple new version of carbonboard ?

vict0rsch:
+ Docs and tutorials for the API

benoit-cty:
Doc on how to name pre-realease : https://www.python.org/dev/peps/pep-0440/#pre-releases

PR #285, opened 2022-02-24
Release version 2.0.0a1

benoit-cty:
Linked to #220  but will not close it because this release is [an alpha](https://www.python.org/dev/peps/pep-0440/#pre-releases)


benoit-cty:
Hello @vict0rsch and @Lothiraldan I propose to add the suffix "a1" to version number so that users knows that it is an alpha release.

And PIP will not install it by default, users would have to force install it.

`--pre     Include pre-release and development versions. By default, pip only finds stable versions.`

vict0rsch (approved):

PR #293, opened 2022-03-29
Fix duration

benoit-cty:
- Remove support for Python 3.6 as it reached its end of life in December 2021.
- Fix duration measurement : The last_duration was computed before making measurement, and the _last_measured_time was set after, so we "lost" some time. Measurement on Windows Intel Power Gadget could take up to 1.6 seconds. For GPU on Linux up to 0.07 second.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.7", "3.8", "3.9"]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


hb9eik:
Which version can I install to test the solution ?
I'm really willing to test this solution.
Thanks again for all your work

codecarbon/core/rapl.py
```diff
@@ -30,10 +30,6 @@ def start(self) -> None:
         self.last_energy = self._get_value()
         return
 
-    def end(self) -> None:
```
benoit-cty, line 33: Dead code

.pre-commit-config.yaml
```diff
@@ -1,16 +1,16 @@
 repos:
     - repo: https://github.com/pycqa/isort
-      rev: 5.8.0
+      rev: 5.10.1
       hooks:
           - id: isort
             args: ["--filter-files"]
     - repo: https://github.com/psf/black
-      rev: 21.4b1
+      rev: 22.3.0
```
benoit-cty, line 8: Needed for https://github.com/psf/black/issues/2969

PR #297, opened 2022-04-11
Add output to loggers option

HerveMignot:
# Logger Output

## New feature

Add two classes to output emissions data to loggers:
* `LoggerOutput()`: output to Python logging logger.
* `GoogleCloudLoggerOutput()`: output to Google Cloud Logging. This sends emissions data of tracked scripts/workloads to some project Cloud Logging

Emissions data are logged as a JSON record, on `tracker.stop()`

See `Collecting emissions to a logger.md` for additional explanations & examples.

## Files modified

* `codecarbon/output.py`: add new classes
* `codecarbon/emissions_tracker.py`: add arguments for saving to loggers

Note: `.pre-commit-config.yaml` has been updated to black 22.3.0 since version 21 is broken by click 8.1.

 ## Files added

* `tests/test_logging_output.py`: `LoggerOutput` test case
* `examples/logging_demo.py`: example to demonstrate usage from the command line
* `Collecting emissions to a logger.md`: explanations & examples

## Misc

pre-commit has been applied.

## Roadmap

* Add option to format / structure output to loggers
* Add support for AWS & Azure logging facilities.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
           - id: isort
             args: ["--filter-files"]
     - repo: https://github.com/psf/black
-      rev: 21.4b1
+      rev: 22.3.0
       hooks:
           - id: black
             args: [--safe]
--- a/Collecting emissions to a logger.md
+++ b/Collecting emissions to a logger.md
@@ -1,0 +1,69 @@
+# Collecting emissions to a logger
+
+The `LoggerOutput` class (and `GoogleCloudLoggerOutput` subclass) allows to send emissions tracking to a logger.
+This is a specific, distinct logger than the one used by the Code Carbon package for its 'private' logs.
+It allows to leverage powerful logging systems, to centralize emissions to some central or cloud-based system, and build reports, triggers, etc. based on these data.
+
+This logging output can be used in parallel of other output options provided by Code Carbon.
+
+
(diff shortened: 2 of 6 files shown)


SaboniAmine:
Hello, thank your for your contribution!
Sending logs directly to a managed logging service is something we hadn't thought of. Could you please give us details about your usage of those logs ?

HerveMignot:
We use these logs (and especially the capabilities to collect them centrally through syslog based systems) to collect emissions from several data pipelines running on data platforms we develop. In the case of Google Cloud, Cloud Logging service allows to easily collect these logs and use Cloud Monitoring for dashboarding emissions by process.
Hope this helps.

SaboniAmine:
It really does, thank you!
We propose a hosted API, on which you could also persist data from your different pipelines, with a dashboard which will let users within an organisation centralize & share their data (this is still in building phase).
Would it be something useful in your case ?

HerveMignot:
This could be an option in some cases.

The advantage with the logging facility is that codecarbon messages end up with other logging message. Supervisors have a central source for logs to use.

benoit-cty:
Thanks for your contribution, it is now available for all in the last release https://pypi.org/project/codecarbon/2.1.0/

The dashboard Amine was talking about is live at http://dashboard.codecarbon.io/

PR #311, opened 2022-05-19
Fix Emission tracker decorator exception handling for offline tracker #310

benoit-cty:
@qmarcou add an exception handler to prevent the tracker to keep running after exception in code.

TODO:

- [x] Add tests
- [x] ~Add docs~ : This will be done in https://github.com/mlco2/codecarbon/pull/313
- [x] Bump version


benoit-cty:
@SaboniAmine do you have an idea why the test fail ?

SaboniAmine:
@benoit-cty a test data file is periodically not created, both on 3.8 & 3.9 python versions, but that doesn't happen locally. I'll have a look

SaboniAmine (approved):

Issue #314, opened 2022-05-31
Reduce precision of geolocalisation


benoit-cty:
Code Carbon collect geolocalization of IP, we need it to find carbon intensity, but we don't need high precision.

So we could reduce precision before sending geolocalization to the API.

benoit-cty:
Hello,
Thanks, I think the only thing to do is to reduce the number of decimals in the two lines:
https://github.com/mlco2/codecarbon/blob/master/codecarbon/core/api_client.py#L121

How many decimal? I don't know, having a 100km precision is enough.

I hope I've answered your question.

benoit-cty:
Thanks for this research.
> The first decimal place is worth up to 11.1 km: it can distinguish the position of one large city from a neighboring large city.
OK, 11.1km, give an area of 123 km² so it seams enough for privacy. Where 111 km could miss country like Luxembourg : only 57 km wide !

PR #317, opened 2022-06-30
Reduce precision of geolocalization

SangamSwadiK:
**Reference Issues**
Issue: #314 

**What does this implement/fix?**
Reduces precision of geolocalization to 1 decimal.


benoit-cty (approved):
Tested locally and the precision is reduced, but still accurate enough.
Thanks for you contribution !

Could you do the version bump ?
Or I could merge your contribution in  a new branch and do the bump from there.

codecarbon/core/api_client.py
```diff
@@ -118,8 +118,8 @@ def _create_run(self, experiment_id):
                 cpu_model=self.conf.get("cpu_model"),
                 gpu_count=self.conf.get("gpu_count"),
                 gpu_model=self.conf.get("gpu_model"),
-                longitude=self.conf.get("longitude"),
-                latitude=self.conf.get("latitude"),
+                longitude=round(self.conf.get("longitude"), 1),
```
benoit-cty, line 121:
```suggestion
                # Reduce precision for Privacy
                longitude=round(self.conf.get("longitude"), 1),
```

benoit-cty (changes requested):
Just a comment proposition

codecarbon/core/api_client.py
benoit-cty: Thanks

benoit-cty (approved):

PR #318, opened 2022-07-04
Adding [AMD EPYC 7662] to cpu_power.csv

Mr-Vicente:
Default TDP is 225W, which can be found here: https://www.amd.com/en/products/cpu/amd-epyc-7662


benoit-cty (approved):
Thanks for your contribution !

PR #323, opened 2022-07-19
Add max_energy_range_uj to deal with RAPL wraparound

### LINKED ISSUE #322: Intel RAPL wraparound
* CodeCarbon version: 2.1.3
* Python version:  3.10.4
* Operating System: Ubuntu 22.04 LTS

### Description

Intel RAPL appears to wrap around at a certain point, CodeCarbon does not appear to handle this gracefully.

### What I Did

Using the ` tracker = OfflineEmissionsTracker(country_iso_code = "GBR", save_to_file=False, log_level = "warning")` then `tracker.start()` to track power usage for a benchmark I get the following errors:
```
[codecarbon ERROR @ 15:25:40] In RAPLFile : Current energy value (Energy(kWh=0.00019679474826900597)) is lower than previous value (Energy(kWh=0.07262206146482492)) ! Source
 file : /sys/class/powercap/intel-rapl/intel-rapl:0/energy_uj
[codecarbon ERROR @ 15:25:55] In RAPLFile : Current energy value (Energy(kWh=0.000589476671025404)) is lower than previous value (Energy(kWh=0.07262206146482492)) ! Source f
ile : /sys/class/powercap/intel-rapl/intel-rapl:0/energy_uj
[codecarbon ERROR @ 15:26:10] In RAPLFile : Current energy value (Energy(kWh=0.00032675557584869596)) is lower than previous value (Energy(kWh=0.07274997709271289)) ! Source
 file : /sys/class/powercap/intel-rapl/intel-rapl:1/energy_uj
```

mp15:
When we detect a negative delta from RAPL energy usage then assume we have wrapped around and add max_energy_range_uj to it. According to Intel docs this should happen about every 60 secs under a heavy load. Fixes #322

### DIFF
--- a/codecarbon/core/cpu.py
+++ b/codecarbon/core/cpu.py
@@ -200,12 +200,19 @@ class IntelRAPL:
                 if "package" in name:
                     name = f"Processor Energy Delta_{i}(kWh)"
                     i += 1
+                # RAPL file to take measurement from
                 rapl_file = os.path.join(self._lin_rapl_dir, file, "energy_uj")
+                # RAPL file containing maximum possible value of energy_uj above which it wraps
+                rapl_file_max = os.path.join(
+                    self._lin_rapl_dir, file, "max_energy_range_uj"
+                )
                 try:
                     # Try to read the file to be sure we can
                     with open(rapl_file, "r") as f:
                         _ = float(f.read())
-                    self._rapl_files.append(RAPLFile(name=name, path=rapl_file))
+                    self._rapl_files.append(
+                        RAPLFile(name=name, path=rapl_file, max_path=rapl_file_max)
+                    )
                     logger.debug(f"We will read Intel RAPL files at {rapl_file}")
                 except PermissionError as e:
                     logger.error(
--- a/tests/test_energy.py
+++ b/tests/test_energy.py
@@ -12,20 +12,31 @@ class TestEnergy(unittest.TestCase):
         self.path = path.join(
             path.dirname(os.path.abspath(__file__)), "test_data/mock_rapl_data.txt"
         )
+        self.max_path = path.join(
+            path.dirname(os.path.abspath(__file__)), "test_data/mock_rapl_data_max.txt"
+        )
 
         # Setup the RAPL File with a high energy level
         with open(self.path, "w") as f:
             f.write("100")
+        # Setup the RAPL maximum value to be the energy level where we wrap
+        with open(self.max_path, "w") as f:
+            f.write("105")
 
-    def test_negative_delta_writes_no_delta(self):
+    def test_wraparound_delta_correct_value(self):
         first_rapl_measure = RAPLFile(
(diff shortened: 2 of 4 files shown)


mp15:
We probably should write a test to verify the delta functionality with mock rollover readings but this should be better than returning 0.

benoit-cty:
Thank you very much for this investigation !

If we understand well:
- RAPL write energy to a counter `energy_uj`.
- At some point the counter `energy_uj` goes above the constant `max_energy_range_uj` and fall back to zero.
- RAPL continue to write energy to `energy_uj`.

So if it goes to zero while codecarbon is running, we have the negative value problem.

With your patch codecarbon will add `max_energy_range_uj` to `last_energy` in case of negative energy to avoid the problem.

May I suggest to refactor your code like this ?
```py
        if self.last_energy > energy:
            logger.debug(
                f"In RAPLFile : Current energy value ({energy}) is lower than previous value ({self.last_energy}). Assuming wrap-around! Source file : {self.path}"
            )
            energy = energy + self.max_energy_reading
          self.power = self.power.from_energies_and_delay(
              energy, self.last_energy, duration
          )
          self.energy_delta = energy - self.last_energy
          self.last_energy = energy
          return
```

mp15:
I've followed your refactor but I have added a new_last_energy variable to it so that it works correctly the next time we call delta(duration). I assume we are accumulating the delta values across calls?
I've also modified the test to compare the results and use almostEqual (as Energy is held as a float internally and non-zero floats are fun to compare).

codecarbon/core/cpu.py
```diff
@@ -201,11 +201,16 @@ def _fetch_rapl_files(self):
                     name = f"Processor Energy Delta_{i}(kWh)"
                     i += 1
                 rapl_file = os.path.join(self._lin_rapl_dir, file, "energy_uj")
+                rapl_file_max = os.path.join(
```
SaboniAmine, line 204: Would you please add a brief comment to describe why this additional rapl file is needed ?

tests/test_energy.py
```diff
@@ -12,20 +12,28 @@ def setUp(self):
...
+            path.dirname(os.path.abspath(__file__)), "test_data/mock_rapl_data_max.txt"
+        )
 
         # Setup the RAPL File with a high energy level
         with open(self.path, "w") as f:
             f.write("100")
+        with open(self.max_path, "w") as f:
+            f.write("105")
 
     def test_negative_delta_writes_no_delta(self):
```
SaboniAmine, line 25: Would you please update this test's name (and the comment below, line 36) ?

SaboniAmine (approved):
Great work, thanks!
This corrects the previous patch we made on this and we have learnt something. 
Thanks for your contribution!

SaboniAmine (approved):

benoit-cty:
Thanks, you're right for `new_last_energy` :+1:

benoit-cty (approved):

Issue #302, opened 2022-04-22
Print version of package in log and API metadata


benoit-cty:
* CodeCarbon version:2.1.0

### Description
* We need to add version in starting log to let user know which version they are using.
* We need the package version in run metadata, so if a package as a problem we may could fix the database.

### How to do it

Since 2.1.0 we have version available through `codecarbon.__version__` so it is easy to log it and send it to the API.

pokacyb:
Hi @benoit-cty,

I'm trying this right now but running into a bit of issue.
Basically whatever I change in the code is not reflecting on output.

![Screenshot from 2022-06-30 21-43-37](https://user-images.githubusercontent.com/77896376/176763853-61df4f7a-eca4-4c9f-85f7-4b85fa8eb08b.png)

For example even after deleting files I still get the same terminal output.
I tried creating new codecarbon init files but same result.
Is there something I'm doing wrong?

benoit-cty:
Hi,
Thanks for your help 👍 
I think you are running the code from Pypi package, so you don't see local modification.
Did you install the code in "editable" mode with `pip install -e .` ?
(see https://github.com/mlco2/codecarbon/blob/master/CONTRIBUTING.md _Install from sources in development mode_ )

PR #325, opened 2022-07-28
302 print version of package in log and api metadata

Acruve15:
Solving [issue 302](https://github.com/mlco2/codecarbon/issues/302)


codecarbon/__init__.py
```diff
@@ -10,4 +11,4 @@
 
 __all__ = ["EmissionsTracker", "OfflineEmissionsTracker", "track_emissions"]
 
-__version__ = "2.1.3"
+__version__ = __version__
```
Fenkiou, line 14: You can just remove that line no?

examples/requirements-examples.txt
```diff
@@ -2,3 +2,4 @@ torch==1.8.1
 tensorflow
 keras-tuner
 torchvision
+sklearn
```
Fenkiou, line 5: How is this related to the issue you’re solving?

carbonserver/carbonserver/api/infra/database/sql_models.py
```diff
@@ -41,6 +41,7 @@ class Run(Base):
     experiment_id = Column(UUID(as_uuid=True), ForeignKey("experiments.id"))
     os = Column(String, nullable=True)
     python_version = Column(String, nullable=True)
+    codecarbon_version = Column(String, nullable=True)
```
benoit-cty, line 44: I'm wondering if we need a more generic metadata, like:
- `client_type` to store `codecarbon package`
- `client_version` to store `2.1.3`
But maybe nobody else will use CodeCarbon API to store data.
@SaboniAmine Any idea ?

SaboniAmine: As we only have data retrieved from the package client at the moment, I guess a single column is enough.
If needed we could change this in the future.

Acruve15: Or we could have a column `metadata` and store the metadata as a dictionary?

benoit-cty: It will be difficult to query it, no ? I prefer to keep specific field, but it's a good idea if we want to be open to custom field.

examples/requirements-examples.txt
benoit-cty: @SaboniAmine Do you think it's a problem to add an Sklearn test ?

Sklearn is smaller that Pytorch so it may be useful.

benoit-cty (approved):
It's OK for me.

carbonserver/carbonserver/database/alembic/versions/298059b19bde_add_codecarbon_version.py
```diff
@@ -0,0 +1,23 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "298059b19bde"
+down_revision = "edcd10edf11d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column("runs", sa.Column("codecarbon_version", sa.String))
```
SaboniAmine, line 19: We may add a default value here, such as "-1".

SaboniAmine (approved):
LGTM! Default value as commented is welcome but not mandatory.
Thanks @Acruve15 & Benoît for this addition!

carbonserver/carbonserver/database/alembic/versions/298059b19bde_add_codecarbon_version.py
Acruve15: @SaboniAmine I will add the default value in the evening :)

PR #331, opened 2022-08-19
setup: add missing dependent package dataclasses

kenplusplus:
Signed-off-by: Lu Ken <ken.lu@intel.com>


benoit-cty:
Can you explain why you need this ? `dataclasses` are included since Python 3.7 and we drop support for Python below 3.7.

benoit-cty:
Hello @kenplusplus Can you explain why you need it ?

benoit-cty:
We drop support for Python below 3.7.

Issue #346, opened 2022-11-14
Inconsistent units in EmissionsData object between emissions and emissions_rate


brother-wolf:
* CodeCarbon version: 2.1.4 (but applicable to any release version since 2021-07-21,  v2.0.0.a1+)
* Python version: 3.10.8 (but applicable to all Python versions)
* Operating System: Mac OS X (but applicable to all OSs) 

### Description

I was experimenting with CodeCarbon looking at the EmissionsData output from the EmissionsTracker with a view to extracting values form model training to store in MLFlow along with other values from the model training. I was interested in the emissions estimates via the fields `duration`, [emissions](https://github.com/mlco2/codecarbon/blob/abb2f8f113e5294494ad20def9c36a8f76d92d0f/codecarbon/emissions_tracker.py#L454), and `emissions_rate`. The documentation [states](https://github.com/mlco2/codecarbon/blob/ee81eab237ca3cc579028009717a609e5db8c9df/docs/output.html#L119) that the emissions_rate is emissions divided per duration.

What I noticed was that this is not the case and the emissions_rate is three orders of magnitude larger than it should be.
Looking at the code in emissions_tracker.py the emissions_rate is calculated as `emissions * 1000 / duration.seconds`. This was introduced in commit `32d4fcb92e0d56aba20d85d8d8ed7fa7f03a545b` on `2021-07-21 12:23` and has been identified as incorrect (but not actually fixed) in commit e73010d86cd6c3a2fced4c5bcd30f0fcee8c25be on `2022-02-08 09:35`. The fix puts a comment in code to show that the actual units of emissions_rate is `g/s`.

I would propose removing the `* 1000` multiplier to make the units in the `EmissionsData` object consistent but as I'm new to the codebase I don't know of any consequences of this. 

Alternatives would be to change:
* the units of emissions to be `g` which may also have knock-on effects
* the documentation to reflect the units inconsistency but this doesn't actually fix the problem
* drop the emissions_rate field as being unnecessary given the presence of both `emissions` and `duration` in the data structure

benoit-cty:
Hello,
Thanks for reporting this. We choose `g/s` as it gave a visually better number than `kg/s`. For total emissions we keep kg for the same reason.
Emission rate and total emissions are not comparable, so they are expressed in different units.
We propose to keep `g/s` and to update the documentation to make it more clear.

Here is a log sample of a run on one GPU:
[...]

PR #347, opened 2022-11-15
Fix unit

benoit-cty:
To close https://github.com/mlco2/codecarbon/issues/346

- [x] Generate doc
- [x] Check dashboard display
- [x] Migrate database by dividing emissions_sum by duration
```
UPDATE public.emissions
SET emissions_rate = emissions_sum / duration;
```


benoit-cty:
@SaboniAmine what do you think of this ? Do we have to release it as major version as the unit change may break someone workflow ?

SaboniAmine:
Lets discuss it this evening. We may propose a rc to be tested by extensive users before merging it to master.

PR #348, opened 2022-11-18
codecarbon-346|BW: making the units of the field 'emissions_rate' in …

brother-wolf:
…EmissionsData() consistent with fields 'emissions' and 'duration' in EmissionsData(); updating the log message to account for the change of units within EmissionsData() so the log messages remain consistent


benoit-cty (approved):
Thanks, I will merge it with my other PR on the subject to update the documentation accordingly to your modification.

PR #355, opened 2022-12-10
Add task-based tracking to enable inference emissions estimation

SaboniAmine:
First step to enable inference estimation estimation : 
- Adding 2 methods to actual tracker : `start_task` & `stop_task`

Next steps may be to 

- [x] Abstract the calls of those methods to a new decorator.
- [x] Add persistence methods to the task data type, and run the persistence at the `tracker.stop()` execution
- [ ] Define & handle production setups, for example when multiple workers run inferences on the same server instance. Would the PID be segregated ? If not, CPU, GPU & RAM measures may be duplicated between tasks.


sashavor:
Some thoughts after testing the code:
- Can we report grams and not kilograms? Most inferences will only be a few fractions of a gram of CO2 in any case..
- It's a bit hard to debug code because of the logging the tracker is doing (took me awhile to find the error because it kept printing out the emissions every second). Can we make it crash if there is an error in the code itself, for example?

benoit-cty:
> Some thoughts after testing the code:
> 
>     * Can we report grams and not kilograms? Most inferences will only be a few fractions of a gram of CO2 in any case..
> 
>     * It's a bit hard to debug code because of the logging the tracker is doing (took me awhile to find the error because it kept printing out the emissions every second). Can we make it crash if there is an error in the code itself, for example?

Hello, you could change the log level in `.codecarbon.config`:
```
[codecarbon]
log_level = ERROR
```

codecarbon/external/scheduler.py
```diff
@@ -33,6 +34,7 @@ def start(self, from_run=False):
         if from_run or self._stopped:
             self._stopped = False
             self._timer = Timer(self.interval, self._run)
+            self._timer.daemon = True
```
SaboniAmine, line 37: @benoit-cty this parameter makes the scheduler stop when an exception occurs in the main thread. 
The tracker now is able to shutdown gracefully when used as decorator or context manager. I think we should discourage the direct usage of start & stop functions, in the main or task-based tracker.

benoit-cty (approved):

PR #356, opened 2022-12-12
Chore/update pre commit

SaboniAmine:
Switch flake8 url to make pre-commit run again.
Updated pre-commit action version


benoit-cty (approved):
Thanks for the upgrade.

PR #366, opened 2023-01-18
Debug api

benoit-cty:
Fix bug with API model. Probably due to a new version of Pydantic that better controls what is send to model.


SaboniAmine (approved):
Merci!

PR #367, opened 2023-01-24
[API] Fix last_run empty error

benoit-cty:
When a project has no run, the API returns a "400:Bad Request" error.
Now it returns 200:OK with an empty json.


carbonserver/carbonserver/api/infra/repositories/repository_runs.py
```diff
@@ -196,5 +200,8 @@ def get_project_last_run(self, project_id, start_date, end_date) -> Run:
             )
 
             if res is None:
-                return None
+                logger.warning(
+                    f"get_project_last_run : No runs for project {project_id}"
+                )
+                raise EmptyResultException(f"No runs for project {project_id}")
```
benoit-cty, line 206: The other option is to `return Empty()` instead of raising exception.

SaboniAmine (approved):

PR #377, opened 2023-02-21
Adding TDP for Intel Xeon Platinum 8171M

najeteb:
Updating cpu_power.csv


benoit-cty:
Thanks for contributing.

PR #379, opened 2023-02-27
Dashboard return api

jl-datascientist:
Error solved in the Dashboard when moving to a project without data (for example, project "Federated Learning").


benoit-cty (approved):

jl-datascientist:
Ok

PR #385, opened 2023-04-02
simplify get_matching_cpu implementation

maxbachmann:
This PR simplifies the implementation of `get_matching_cpu`. In addition it replaces `fuzzywuzzy` with `rapidfuzz`, since `fuzzywuzzy` is not maintained anymore and is GPL licensed.

In theory this new implementation is faster, but since CPU lists should be pretty short this does not really matter here.

### DIFF
--- a/.conda/meta.yaml
+++ b/.conda/meta.yaml
@@ -32,8 +32,8 @@ requirements:
     - psutil
     - pynvml
     - py-cpuinfo
-    - fuzzywuzzy
     - prometheus_client
+    - rapidfuzz
 
 test:
   imports:
--- a/codecarbon/core/cpu.py
+++ b/codecarbon/core/cpu.py
@@ -7,13 +7,10 @@ import os
 import shutil
 import subprocess
 import sys
-import warnings
 from typing import Dict, Tuple
 
 import pandas as pd
-
-with warnings.catch_warnings(record=True) as w:
-    from fuzzywuzzy import fuzz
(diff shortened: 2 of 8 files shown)


benoit-cty (approved):
Thanks for looking at this !

benoit-cty:
There is an error in CI:
```
  ./src/rapidfuzz/cpp_common.pxd:396: TypeError : ('Got unexpected keyword arguments: ', 'process')
```
Do you have an idea ?

maxbachmann:
Ah yes it should have used `processor` instead of `process`.

benoit-cty:
Hello,
Sorry for the delay but some tests fail with rapidfuzzy : `pytest -v tests/test_cpu.py::TestTDP::test_get_matching_cpu`
And if I lower the threshold it make another test to fail.

maxbachmann:
With rapidfuzz `v3.0.0` strings are no longer preprocessed by default. Enabling it again should fix this.

maxbachmann:
@benoit-cty are you still interested in this? I could rebase it in this case. Otherwise I will close this PR.

benoit-cty:
Hello, yes we are still interested. If all test pass I'm ok to merge it.
Thanks for the reminder :+1:

maxbachmann:
@benoit-cty I rebased the change + fixed the new linter rules

benoit-cty:
Thanks a lot !

Issue #390, opened 2023-04-12
Add Python 3.11 Support - `dataclass` complains about not using `default_factory` for field


arrrrrmin:
* CodeCarbon version: `2.1.4`
* Python version: `Python 3.11.3`
* Operating System: `OSX 13.2.1`

### Description

Hi 👋 first of all thanks for publishing the package.
Upon importing `codecarbon` the `dataclass` wrapper complains about `Energy`https://github.com/mlco2/codecarbon/blob/2108c72ebfa81587d69656464173aecb8989cb0a/codecarbon/core/rapl.py#L12
> mutable default <class 'codecarbon.core.units.Energy'> for field energy_delta is not allowed: use default_factory

Here is the complete Exception:
```
tests/test_energy.py:2: in <module>
    from codecarbon import OfflineEmissionsTracker
../../Library/Caches/pypoetry/virtualenvs/neuromorphic-P4cmU1Sb-py3.11/lib/python3.11/site-packages/codecarbon/__init__.py:5: in <module>
    from .emissions_tracker import (
../../Library/Caches/pypoetry/virtualenvs/neuromorphic-P4cmU1Sb-py3.11/lib/python3.11/site-packages/codecarbon/emissions_tracker.py:16: in <module>
    from codecarbon.core import cpu, gpu
../../Library/Caches/pypoetry/virtualenvs/neuromorphic-P4cmU1Sb-py3.11/lib/python3.11/site-packages/codecarbon/core/cpu.py:18: in <module>
    from codecarbon.core.rapl import RAPLFile
../../Library/Caches/pypoetry/virtualenvs/neuromorphic-P4cmU1Sb-py3.11/lib/python3.11/site-packages/codecarbon/core/rapl.py:7: in <module>
    @dataclass
/opt/homebrew/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/dataclasses.py:1223: in dataclass
    return wrap(cls)
/opt/homebrew/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/dataclasses.py:1213: in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash,
/opt/homebrew/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/dataclasses.py:958: in _process_class
    cls_fields.append(_get_field(cls, name, type, kw_only))
[...]

PR #391, opened 2023-04-16
#390 Add Python 3.11 support

arrrrrmin:
`RAPLFile` dataclass is required to use fields `default_factory` in Python version >= 3.11. Believe the corresponding change, is this one: [whatsnew/3.11#dataclasses](https://docs.python.org/3/whatsnew/3.11.html#dataclasses).

All test but 2 pass for both 3.10 and 3.11
* `test_carbon_tracker_offline_constant_default_cpu_power`
* `test_ram_diff`
while `test/test_energy.py` passes for all version.


benoit-cty (approved):
Thanks for your contribution. I will integrate it in the next release.

PR #398, opened 2023-04-29
PUE support

benoit-cty:
I add an option to provide a  PUE to CodeCarbon.

I multiply the energy consumed by the PUE factor provided by the user.

The PUE is a user choice. It could be the PUE of the datacenter, or a bigger factor to take all the energy loss between extraction to consumption. For example [Energy loss is single biggest component of today’s electricity system ](https://yaleclimateconnections.org/2022/10/energy-loss-is-single-biggest-component-of-todays-electricity-system/).


inimaz (approved):
LGTM!

PR #401, opened 2023-04-30
feat: get gpu energy consumption directly from pynvml

inimaz:
Until now, GPU energy consumption has been calculated as `gpu_power * time`. And this gpu_power is measured every ``measure_power_secs``.
 The problem is that this way, if there are spikes of consumption, they will go unnoticed. 

# Goal of this PR
It is preferable to get the energy consumption instead, directly from pynvml.
This PR tries to accomplish that.


codecarbon/external/hardware.py
```diff
@@ -52,46 +52,64 @@ def start(self) -> None:
...
                 set(range(self.num_gpus))
             ), f"Unknown GPU ids {gpu_ids}"
         else:
             gpu_ids = set(range(self.num_gpus))
+        return gpu_ids
 
-        gpu_power = self._get_power_for_gpus(gpu_ids=gpu_ids)
-        return gpu_power
+    def total_power(self) -> Power:
+        # FIXME: do we still need this total_power?
```
inimaz, line 107: @benoit-cty maybe we do not need the total power of the gpu anymore? Now the power is taken from converting the energy by calculating the delta. (see `measure_power_and_energy` function)

benoit-cty:
I tested it with and without [gpu_burn](https://github.com/wilicc/gpu-burn) and it works well.

We have a problem for the first measure as the estimated power is above the expected.
With my GTX 1080 Ti we estimate a power of 323 W for the first measure, then 245 W. The max power given for my card is 250 W so I suspect we initialize the energy counter some time after we start `_last_measured_time`.
I tried to move it at the end of https://github.com/mlco2/codecarbon/blob/6e5df712d512966f6177a5672d3e60e1548439d9/codecarbon/emissions_tracker.py#L409 but it doesn't change anything.

By the way, it's probably not relative to this PR and we may have the same problem with CPU.

inimaz:
> it's probably not relative to this PR and we may have the same problem with CPU
@benoit-cty  Do you think it is blocking the merge? Or should we merge it and create an issue for this?

PS: I start the last energy as 0 here https://github.com/mlco2/codecarbon/blob/6b151244887e29acf8f3cec48c5d8e6e32ba2959/codecarbon/core/gpu.py#L40
maybe this is not true, maybe there is an initial boost just to start it up or sth like that. I could do a measurement there to set the initial `last_energy`

benoit-cty:
Maybe we could implement a `start` for GPU like we do for CPU : https://github.com/mlco2/codecarbon/blob/6b151244887e29acf8f3cec48c5d8e6e32ba2959/codecarbon/external/hardware.py#L204C26-L204C26

benoit-cty (approved):

Issue #382, opened 2023-03-16
how use from CLI to measure energy consumed during run of a given command


mkbane:
* CodeCarbon version: ALL
* Python version: N/A
* Operating System: Linux

### Description
I wish to use codecarbon to give me the energy consumed by my CPU for running a command eg "a.out", so wanted to do something like

codecarbon ./a.out

and get a figure e.g. of Joules of kWh. 

How do I do this? I cannot see it within the documentation

jennydaman:
@benoit-cty would you accept a PR for this? I think we should implement a command `codecarbon monitor [COMMAND...]` where:

- if `codecarbon monitor` is ran without positional arguments, monitor machine until CTRL-C (current behavior)
- if `codecarbon monitor` has positional arguments, call `subprocess.run(...)` and measure that

PR #431, opened 2023-07-23
Monitor subprocess

jennydaman:
Resolves https://github.com/mlco2/codecarbon/issues/382

This PR adds functionality to the `codecarbon monitor` command.

When you run `codecarbon monitor`, the previous behavior is kept the same.

When you run `codecarbon monitor -- some positional --arguments -lah`, the positional arguments are ran as a subprocess and monitored with the option `tracking_mode="process"`.

### DIFF
--- a/codecarbon/cli/monitor.py
+++ b/codecarbon/cli/monitor.py
@@ -1,7 +1,10 @@
 """
 Implementations of the ``codecarbon monitor`` subcommand.
 """
+import os
+import subprocess as sp
 import time
+from typing import Sequence
 
 from codecarbon import EmissionsTracker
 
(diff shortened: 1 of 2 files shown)


codecarbon/cli/monitor.py
```diff
@@ -0,0 +1,42 @@
...
+    Run and monitor a subprocess.
+
+    :return: return code of the subprocess.
+    """
+
+    with EmissionsTracker(
+        measure_power_secs=measure_power_secs,
+        api_call_interval=api_call_interval,
+        save_to_api=api,
+        tracking_mode="process",
```
benoit-cty, line 39: Not sure if it works with process. It need a test with a process that consume RAM to see the difference with __machine__ because __process__ does not work for GPU and CPU yet.

benoit-cty (changes requested):
Hello,
Thanks for the contribution. Can you add a documentation in https://github.com/mlco2/codecarbon/blob/master/docs/edit/usage.rst ?

SaboniAmine:
@benoit-cty besides documentation, is there any other blocker on this PR ?

benoit-cty:
> @benoit-cty besides documentation, is there any other blocker on this PR ?

It has to be tested to ensure it works as expected.

codecarbon/cli/monitor.py
inimaz: I confirm launching it like this does not work sadly. For instrumenting a process, it needs to be in the same process it is instrumenting. It will be cool in the future to do something like "instrument process 1234" passing the id.
For now I suggest to have `codecarbon monitor` to measure whole machine only but start/stop when the process starts/stops. See #1004 .

What do you think @jennydaman ?

inimaz:
Thanks for this PR @jennydaman, it is a good idea! interested in applying those changes? Sorry I started the change in #1004 without knowing of the existence of this branch. My bad :(

We have moved the cli from click to typer, this is why you see this many conflicts in this PR.
I think if you resolve the conflicts and remove the `tracking_mode=process` this PR is good to merge. What do you think?

jennydaman:
Hi @inimaz worry not about "stealing" work! This PR is a bit stale, I'll keep it in mind but it might take me some weekends to get to.

benoit-cty:
I close this PR as the sub-process monitoring has been done in #1004

PR #424, opened 2023-07-04
Fix dashboard data processing

EdAbati:
Fixes #423 

Solutions:
- Fixed `pandas.DataFrame.to_dict("records")`
- Updated column names to reflect data in https://github.com/mlco2/codecarbon/blob/master/codecarbon/data/private_infra/global_energy_mix.json
- Added unit tests to catch these tyupe of issues in the future
- Extracted some manipulation in `get_country_choropleth_data` to make it easier to test

Please let me know if this is something you are interested in, and if I can improve this PR in any way :)

PR #432, opened 2023-07-23
Fix to_dict("records")

jennydaman:
`to_dict("rows")` is not a valid value for parameter orient.

https://github.com/pandas-dev/pandas/issues/32515


jennydaman:
Oops, sorry duplicate of https://github.com/mlco2/codecarbon/pull/424

PR #433, opened 2023-07-23
Fix *_Twh key names

jennydaman:
The schema of the data added in https://github.com/mlco2/codecarbon/commit/4949003f1599c2b7a5b29c5571ffcc9adcdde7dc is different from before, breaking untyped code.

- `_TWh` suffixes was changed to `_Twh`
- `geothermal_TWh` data is no longer provided

While I fixed what was necessary to get things working, I did not bother fixing https://github.com/mlco2/codecarbon/blob/master/codecarbon/data/private_infra/our_world_in_data-2021_data.ipynb which is now out-of-date and misleading.


jennydaman:
Oops, sorry duplicate of https://github.com/mlco2/codecarbon/pull/424

PR #441, opened 2023-08-14
Load the organization list

benoit-cty:
- Load the organization list each time the page is loaded and not only when the server start.
- Fix a small bug in task monitoring
- Fix a typo in energy unit


inimaz (approved):
Looks good to me! kW to kWh + the dashboard refresh of organizations + check if the scheduler is already stopped before trying to stop it

benoit-cty:
Thanks for the review.

Issue #446, opened 2023-09-05
stop_task PeP8 inside emission_tracker is probably incorrect 


VictorCrd:
* CodeCarbon version: 2.1.4
* Python version: 3.11.1
* Operating System: Ubuntu 22.04

### Description

PeP8 definition of stop_task might need to be changed.

Function is currently returning task_emission_data (EmissionsData).

But PeP8 info is "the function is returning None / a float"

Correct PeP8 might be this:

    def stop_task(self, task_name: str = None) ->  EmissionsData:
        """
        Stop tracking a dedicated execution task. Delta energy is computed by task, to isolate its contribution to total
        emissions.
        :return: EmissionsData
        """
        task_name = task_name if task_name else self._active_task
        self._measure_power_and_energy()

        emissions_data = self._prepare_emissions_data(delta=True)

        task_duration = Time.from_seconds(
            time.time() - self._tasks[task_name].start_time
        )

        task_emission_data = emissions_data
        task_emission_data.duration = task_duration.seconds
        self._tasks[task_name].emissions_data = task_emission_data
        self._tasks[task_name].is_active = False
        self._active_task = None
        return task_emission_data

PR #452, opened 2023-09-15
Changed the output type of stop_task to EmissionsData

lucasvandijck:
As mentioned in Issue #446, the output type of the stop_task function should be EmissionsData.


inimaz:
Thanks a lot! Merging it

Issue #427, opened 2023-07-15
Add a geoip backup


benoit-cty:
CodeCarbon use [http://ip-api.com/json/](http://ip-api.com/json/) for IP geolocalization but sometime it fail so we could add [http://ip-api.com/json/](http://ip-api.com/json/) as a backup.

benoit-cty:
Done in https://github.com/mlco2/codecarbon/pull/453

PR #453, opened 2023-09-15
Added a backup for the geo API

lucasvandijck:
According to issue #427, I added IP-API as a backup for geo localisation.

Since IP-API does not offer three-letter country codes, I also added a call to api.first.org which looks up the code using the country name. I find this method not super clean, since there is no guarantee to get the right code, but I could not think of another way off the bat.

I also added a test to check this new backup strategy.

Let me know if you have any remarks!


benoit-cty:
Thank you, this is really nice ! Unfortunately the test is failing : https://github.com/mlco2/codecarbon/actions/runs/6197183554/job/16939148311?pr=453#step:5:820

lucasvandijck:
> Thank you, this is really nice ! Unfortunately the test is failing : https://github.com/mlco2/codecarbon/actions/runs/6197183554/job/16939148311?pr=453#step:5:820

I'm not quite sure, but it seems the tests aren't failing due to my code? When I run them locally I get all kinds of issues since I'm running on Mac without Intel Power Gadget, but when looking at the GitHub logs, most errors also seem to give similar outputs:

`No such file or directory: './intel_power_gadget_log.csv'`

inimaz:
Hi @lucasvandijck !
Thanks for adding this.
The error you are seeing in the tests is in this line https://github.com/mlco2/codecarbon/blob/master/tests/test_emissions_tracker.py#L129. Which is normal because we mock the `requests.get` and we check how many times it is called. Your change forces that test to call `get` twice instead of 1.
Do ```self.assertEqual(2, mocked_requests_get.call_count)``` instead of ``` self.assertEqual(1, mocked_requests_get.call_count)``` and you'll be good.

inimaz (approved):
Looks good! Thanks @lucasvandijck

PR #454, opened 2023-09-15
Removed depricated Jumbotron element

lucasvandijck:
I tried starting up the Dash dashboard, but ran into an issues regarding the [depricated Jumbotron element](https://dash-bootstrap-components.opensource.faculty.ai/docs/components/jumbotron/).

I switched it out for a separate Container and Div with appropriate padding so the code now runs again.


benoit-cty (approved):

benoit-cty:
Great, thank you :+1:

PR #456, opened 2023-09-24
Fix Power in methodology

### LINKED ISSUE #455: Incorrect use of the term 'Power'  in the Methodology webpage
### Description

In the first code block in the Methodology page you state:
```
P = Power Consumed by the computational infrastructure: quantified as kilowatt-hours.
```

This is incorrect because kilowatt-hours are a unit of energy (1kWh = 3.6e+6J)

You use the term 'power' however that is the energy over time which is measured in watts.

watts and watt-hours are related as follows:

take a 60 watt (0.06 kilowatt) laptop charger. To get the total energy consumed by that charger you first get the duration that it is on charge, say 1 hour.

Then multiply together 0.06 kilowatts * 1 hour = 0.06 kilowatt-hours.

For your scenario you want to change your use of the word 'Power' with 'Energy' because you multiply the total energy (in kWh) by the carbon intensity of the electricty consumed (in gCO₂eq/kWh) to output the total carbon emissions (in gCO₂eq).

benoit-cty:
Will close #455


inimaz (approved):
Nice catch! 
I was even checking if there was anything wrong in the code (https://github.com/mlco2/codecarbon/blob/master/codecarbon/core/units.py). But it looks like  we are safe :D, Power and Energy are well measured in their correct units (kW and kWh respectively).

PR #459, opened 2023-10-08
feat(output): unifying execution of 'live' output handlers

eledhwen:
Following-up on discussion in https://github.com/mlco2/codecarbon/pull/458


codecarbon/emissions_tracker.py
```diff
@@ -698,19 +702,18 @@ def _measure_power_and_energy(self) -> None:
...
+            and self._api_call_interval != -1
+            and self._measure_occurrence >= self._api_call_interval
+        ):
+            emissions = self._prepare_emissions_data(delta=True)
+            logger.info(
+                f"{emissions.emissions_rate * 1000:.6f} g.CO2eq/s mean an estimation of "
+                + f"{emissions.emissions_rate*3600*24*365:,} kg.CO2eq/year"
+            )
+            for handler in self._output_handlers:
+                handler.out(emissions)
```
inimaz, line 715: Hey @eledhwen  thanks! Sorry for my late reply on this. I completely missed this PR.
I like this idea. The problem I see is if here you do `handler.out(emissions)` you will output the emissions in delta mode in all output_handlers.  So as you suggested in [here]( https://github.com/mlco2/codecarbon/pull/458#issuecomment-1752002391), we need to flag the output_methods as `live` and `flush`/`exit`. Or some other kind of flagging.

In  this [PR](https://github.com/mlco2/codecarbon/pull/542/files#) I did something similar by adding `use_emissions_delta`  true/false to each output method (By default as `false` in the BaseOutput`). 

On the other hand `_output_handlers` only contains for now prometheus and the api, so we could merge this and do the flag in a separate PR. What do you think?

inimaz: Indeed, I like the idea of the methods better. And if not implemented, the output method will do nothing.
> In any case, it's a breaking change on the part of the BaseOutput API ... 

Not necessarily! If we keep having and `out` method that points by default to the `exit` method for instance (as a first version). Then we could deprecate this `out` method in favor of the others on the long run. 

In any case, I think this PR can be merged

eledhwen: I'll try to rebase and have a shot at it asap

eledhwen: I just pushed a new draft, I didn't look into tests nor documentation, let me know what you think.

Some notes:
* `BaseOutput` has 3 separate methods, `out`, `live_out` and `task_out`
* removal of `persistence_objs` which is now the same set of objects as `_output_handlers`
* I took some liberties with the `_prepare_emissions_data` and separated the aggregation from the delta computation, steering them toward a more SRP-compatible design
* I took some liberties with the time at which some emission data are computed, I preferred more straightforward code to slight optimizations, but this can be challenged

codecarbon/emissions_tracker.py
```diff
@@ -370,38 +372,35 @@ def _init_output_methods(self, api_key):
         """
         Prepare the different output methods
         """
-        self.persistence_objs: List[BaseOutput] = list()
-
         if self._save_to_file:
-            self.persistence_objs.append(
+            self._output_handlers.append(
                 FileOutput(
                     os.path.join(self._output_dir, self._output_file),
```
inimaz, line 378: Twice the `output_dir`. Maybe we can give as input of FileOutput:

- `output_dir`
-  `output_file_name`
- `on_csv_write`

WDYT?

codecarbon/output.py
```diff
@@ -131,7 +131,17 @@ class BaseOutput(ABC):
     """
 
     @abstractmethod
-    def out(self, data: EmissionsData):
+    def out(self, total: EmissionsData, delta: EmissionsData):
+        pass
+
+    #TODO: document lifecycle scope
+    @abstractmethod
+    def live_out(self, total: EmissionsData, delta: EmissionsData):
```
inimaz, line 139: I like that you give `total` AND `delta` to the `live_out`. We could use the total for metrics that are cumulative for instance.

inimaz (changes requested):
Good! I like it thanks!
You might have some conflicts when merging with `master`, since we have moved all the `output_methods` in `output.py` to a its own directory.

eledhwen:
> Good! I like it thanks! You might have some conflicts when merging with `master`, since we have moved all the `output_methods` in `output.py` to a its own directory.

Indeed, I just rebased from master and re-dispatched the changes to the various output definitions.

codecarbon/emissions_tracker.py
eledhwen: Done

inimaz:
Seeing the output of the CI ==> There are 2 things @eledhwen : 
1. Run `hatch run dev:precommit-install` and commit again, this will apply the correct formatting 
2. In file `codecarbon/output_methods/base_output.py` Change `from codecarbon.output import TaskEmissionsData` to `from codecarbon.output_methods.emissions_data import TaskEmissionsData`==> to avoid circular dependencies.

eledhwen:
> Seeing the output of the CI ==> There are 2 things @eledhwen :
> 
> 1. Run `hatch run dev:precommit-install` and commit again, this will apply the correct formatting
> 2. In file `codecarbon/output_methods/base_output.py` Change `from codecarbon.output import TaskEmissionsData` to `from codecarbon.output_methods.emissions_data import TaskEmissionsData`==> to avoid circular dependencies.

Done ✅

eledhwen:
@inimaz I also fixed the impacted tests, it should be good to go now.

inimaz (approved):

inimaz:
@eledhwen Thanks for the feature! Looks good.

PR #461, opened 2023-10-18
LuisBlanche/issue362

### LINKED ISSUE #362: Suggestion re log levels
* CodeCarbon version:  https://files.pythonhosted.org/packages/e3/62/b9bff8dfc6d32502de644285b6226d95b6830092f6f7af991a94ab059f38/codecarbon-2.1.4-py3-none-any.whl

* Python version: 3.6.8
* Operating System: Linux-3.10.0-1160.36.2.el7.x86_64-x86_64-with-centos-7.9.2009-Core

### Description

On page
https://mlco2.github.io/codecarbon/parameters.html
under "Input Parameters" it may be better to list the log levels in order of decreasing verbosity (and to note this), i.e. debug before (not after) info.

Under "env vars" (page https://mlco2.github.io/codecarbon/usage.html#configuration) would be useful to list further examples (full list). I particularly find CODECARBON_LOG_LEVEL="WARNING" saves so much o/p to terminal

LuisBlanche:
- fixes #362
- Update pre-commit versions


benoit-cty:
Can you add the example config given in the issue to inform users that they can reduce verbosity ?

LuisBlanche:
I think there is a problem with the Flake8 pre-commit hook. It raises a lot of :
`codecarbon/core/emissions.py:202:76: E231 missing whitespace after ':'`. errors on files that were not changed. 

And you can see from the following example that we cannot add a space after the ':' in that case. 

```python
logger.debug(
            f"We apply an energy mix of {emissions_per_kWh.kgs_per_kWh*1000:.0f}"
            + f" g.CO2eq/kWh for {geo.country_name}"
        )
 ```
 
Maybe something has changed with the Flake8 hook configuration ?

LuisBlanche:
I had to ignore the following Flake8 error codes in `.flake8` :
[E201](https://www.flake8rules.com/rules/E201.html) : Whitespace after "("
[E231](https://www.flake8rules.com/rules/E231.html) : Missing whitespace after ',', ';', or ':' 
[E226](https://www.flake8rules.com/rules/E226.html) : Missing whitespace around arithmetic operator 

NB: these were only raising in the CI , not in local when running `pre-commit run` . E231 is raised at times it shouldn't be. The other ones are mostly raised because of line breaks introduced by black (which is acceptable imho)

LuisBlanche:
@benoit-cty  what do you think about the error codes i had to ignore ? seems ok to merge ?

benoit-cty:
The problem was `actions/setup-python@v3`  that switch to Python 3.12, so I force 3.11.

I open a new issue for that : https://github.com/mlco2/codecarbon/issues/465

You could merge this PR :+1:

PR #470, opened 2023-11-04
fix: LICENSE year and full name

inimaz:


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) [year] [fullname]
+Copyright (c) [2019] [Code Carbon]
```
benoit-cty, line 3:
```suggestion
Copyright (c) 2019 Mila, BCG Gamma, CometML, Haverford College
Copyright (c) 2023 Code Carbon NGO
```

benoit-cty (changes requested):
@SaboniAmine : What do you think ?

LICENSE
SaboniAmine: LGTM! We need explicit consent from Victor, and maybe precise that the previous copyright was 2019-2023.

SaboniAmine (approved):
Thanks!

benoit-cty (approved):
Thanks !

PR #472, opened 2023-11-09
Fix SLURM RAM monitoring

### LINKED ISSUE #447: CodeCarbon spams Slurm controler with `scontrol`
Hello,

We are facing some issues with CodeCarbon in our facility because it spams our Slurm controler with `scontrol`.

If rate-limiting is enabled this also means that the jobs that use CodeCarbon are slowed down because they spend a lot of time waiting for `scontrol` to return.

As far as I could tell, the values that are read from `scontrol` won't change during the run so they could be cached by CodeCarbon instead of being fetched everytime.

We are willing to provide a PR if needed. If caching is used somewhere else, feel free to point us to the relevant code so that we use a similar implementation.

Rémi

benoit-cty:
Fix SLURM RAM monitoring :
- Fix error when calling scontrol
- Do only one call instead of one at each measure

Will close #447


benoit-cty:
Wrong branch, sorry.

PR #473, opened 2023-11-09
Fix SLURM RAM monitoring

### LINKED ISSUE #447: CodeCarbon spams Slurm controler with `scontrol`

benoit-cty:
Fix SLURM RAM monitoring :
- Fix error when calling scontrol
- Do only one call instead of one at each measure

Will close #447


benoit-cty:
Tested on a SLURM environment : there is only too call per sessions and no more error message.

codecarbon/external/test.ipynb
SaboniAmine, line 1: Can you clean this file ? Or move it to test folder ?

SaboniAmine (commented):
Thanks Benoît! Small comment about the notebook, not sure we want to merge it to master

codecarbon/external/test.ipynb
benoit-cty: Thanks, I will delete it.

vict0rsch:
I'll have a look next week 😊

vict0rsch:
There was an issue with `$SLURM_JOBID` which was changed to `$SLURM_JOB_ID`, making some queries like `scontrol show job $SLURM_JOBID"` fail.

vict0rsch:
This works as expected

```
from codecarbon.external.hardware import RAM

ram = RAM(tracking_mode="process")
ram.slurm_memory_GB
```

benoit-cty:
> There was an issue with `$SLURM_JOBID` which was changed to `$SLURM_JOB_ID`, making some queries like `scontrol show job $SLURM_JOBID"` fail.

Thanks for your review, it seems to be deprecated since 2017 so it is safe to remove the support of the old name !

codecarbon/external/test.ipynb
benoit-cty: It's deleted. Can you approve the PR if everything is OK ?
It has been rebased.

vict0rsch (approved):

PR #480, opened 2023-11-24
Better error message on invalid inputs

Erotemic:
I was getting an error when running this code:

```python
    emissions_tracker = OfflineEmissionsTracker(
        country_iso_code='USA',
        region='Virginia',
        cloud_provider='aws',
        cloud_region='us-east-1',
        country_2letter_iso_code='us'
    )

    emissions_tracker.start()
    emissions_tracker.flush()
```

I think it is because of a misconfiguration, removing everything but the country_iso_code seems to work, but I thought the error I ran into (about a length-0 data frame) was cryptic, so I checked for a failure and raised a better message.


inimaz (approved):
Thanks for checking this! The error is more clearly explained indeed.

PR #481, opened 2023-11-25
docs: typo in power usage RAM section

sumn2u:
I think here we should have 8 GB instead of 8 BG.


benoit-cty:
Yes, thanks for your contribution :+1:

inimaz (approved):
Thanks! Merging it

PR #485, opened 2023-12-20
Do not count *.ipynb example files in the git stats

inimaz:
# The issue
This is a **purely cosmetic** change.

Jupyter notebook appears as the primary language of CodeCarbon
image

This is misleading, since someone new arriving to the project, might think that this is mainly jupyter scripts.

# The solution
I added this `.gitattributes` file to not count the *.ipynb files in the git stats.
> Source: https://proandroiddev.com/removing-noise-from-your-github-language-stats-e96113f8183d

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.ipynb linguist-vendored
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 *.ipynb linguist-vendored
+*.json linguist-vendored
+*.csv linguist-vendored


.gitattributes
```diff
@@ -0,0 +1 @@
+*.ipynb linguist-vendored
```
benoit-cty, line 1: Nice, I don't know this feature ! May I suggest to add CSV and JSON as well ?

```suggestion
*.ipynb linguist-vendored
*.json linguist-vendored
*.csv linguist-vendored
```

inimaz: Good one! Adding it

benoit-cty (approved):
Thanks a lot !

PR #487, opened 2024-01-03
feat: guess IntelPowerGadget path

inimaz:


benoit-cty (approved):
Thanks for our Windows users !

Issue #445, opened 2023-09-05
RoCm support


fxmarty:
* CodeCarbon version: main
* Python version: all
* Operating System: Linux

### Description

It would be great to add RoCm support to codecarbon. For now, codecarbon requires and depends on `pynvml`, that is an interface only for Nvidia GPUs. However, using AMD GPUs is becoming easier and easier, and we can expect some researchers to make the switch from Nvidia.

A WIP alternative for RoCm seem to be https://github.com/RadeonOpenCompute/pyrsmi

I could possibly submit a PR if there is interest.

benoit-cty:
Hello,
Thank you, I will be happy to help you made a PR for this.
You have to modify https://github.com/mlco2/codecarbon/blob/master/codecarbon/core/gpu.py.

IlyasMoutawwakil:
@fxmarty @benoit-cty I was able to perform all `pynvml` operations in `gpu.py` using `amd-smi`, `pyrsmi` lacks a lot of the metrics, I will submit a PR.

PR #490, opened 2024-01-16
AMD GPUs support

IlyasMoutawwakil:
This PR adds support for amd gpus through [`amdsmi`](https://github.com/ROCm/amdsmi).
#445

### DIFF
--- a/codecarbon/core/gpu.py
+++ b/codecarbon/core/gpu.py
@@ -1,9 +1,22 @@
+from collections import namedtuple
 from dataclasses import dataclass, field
 
-import pynvml
-
 from codecarbon.core.units import Energy, Power, Time
+from codecarbon.core.util import is_amd_system, is_nvidia_system
 from codecarbon.external.logger import logger
+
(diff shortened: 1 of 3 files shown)


benoit-cty:
Thanks, it's really great!

Do you think it's possible to have a machine with both AMD and Nvidia GPU ?

Before merging, this PR needs unit tests and documentation.

IlyasMoutawwakil:
Apparently it's possible to have both (though very rare)
Not sure about unit tests, do you run NVIDIA GPU tests in a workflow ?

benoit-cty:
For the test, we do 'mock' to check the function call.
This way you can test if the code works with both nvidia and amd GPU, even if you don't have them.

fxmarty:
@IlyasMoutawwakil any update?

benoit-cty:
ROCm is not packaged for Python, but I think it is not blocking for this PR, we already have the issue for MacOSX and Windows.

To merge this PRR, we need tests and documentation.

Issue #493, opened 2024-01-23
Unable to access the Intel RAPL interface with an Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz processor.


AntoinePlanche:
* CodeCarbon version: 2.3.3
* Python version: Python 3.8.10
* Operating System: Linux eng-srv1.black.box 5.4.0-146-generic #163-Ubuntu SMP Fri Mar 17 18:26:02 U             TC 2023 x86_64 x86_64 x86_64 GNU/Linux

### Description

I'm trying to monitor the energy consumption of a python code that I'm running on my company's bare metal servers. The server's Intel RAPL interface exists, codecarbon detects it but returns to constant mode. The warning tells me to contact you.

### What I Did
input : python3 train_CDR_CNN.py
trace : 
[codecarbon INFO @ 17:14:47] [setup] CPU Tracking...
[codecarbon INFO @ 17:14:47] Tracking Intel CPU via RAPL interface
[codecarbon WARNING @ 17:14:48] No CPU tracking mode found. Falling back on CPU constant mode.
[codecarbon WARNING @ 17:14:49] We saw that you have a Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz but we don't know it. Please contact us.
[codecarbon INFO @ 17:14:49] CPU Model on constant consumption mode: Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz
[codecarbon INFO @ 17:14:49] >>> Tracker's metadata:
[codecarbon INFO @ 17:14:49]   Platform system: Linux-5.4.0-146-generic-x86_64-with-glibc2.29
[codecarbon INFO @ 17:14:49]   Python version: 3.8.10
[codecarbon INFO @ 17:14:49]   CodeCarbon version: 2.3.3
[codecarbon INFO @ 17:14:49]   Available RAM : 125.806 GB
[codecarbon INFO @ 17:14:49]   CPU count: 48
[codecarbon INFO @ 17:14:49]   CPU model: Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz
[codecarbon INFO @ 17:14:49]   GPU count: None
[codecarbon INFO @ 17:14:49]   GPU model: None

LuisBlanche:
[...]

PR #497, opened 2024-01-30
fix: emissions tracker #493

inimaz:
There is an if clause that is not taken into account. This might be the root cause of #493


benoit-cty (approved):

benoit-cty:
Nice catch ! I've made the bump to publish a new package this evening.

PR #498, opened 2024-01-30
feat: TDP of Intel Xeon E5-2680 v3

inimaz:
According to https://ark.intel.com/content/www/us/en/ark/products/81908/intel-xeon-processor-e5-2680-v3-30m-cache-2-50-ghz.html, the TDP of this is 120 W
image

PS: related to #493


benoit-cty (approved):
:+1:

PR #511, opened 2024-03-06
Draft: Fixed wrong return type hint and updated doc string

MyGodItsFull0fStars:
Hi, I made some changes to the `stop_task(...)` function of the `BaseEmissionTracker` class.

I changed the return type from `float` to the actual return type `EmissionsData`. 
This was probably an oversight, as in almost every other function the return type is a `float`.

I also changed the return type in the doc-string from `None` to the actual one.

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -471,11 +471,11 @@ class BaseEmissionsTracker(ABC):
         )
         self._active_task = task_name
 
-    def stop_task(self, task_name: str = None) -> float:
+    def stop_task(self, task_name: str = None) -> EmissionsData:
         """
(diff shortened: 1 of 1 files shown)


benoit-cty:
Hello,
Thanks, I think it's a good idea. It's a breaking change so it needs to be put in a major release, and do the same things on all methods.

MyGodItsFull0fStars:
Hello, 
okay, when I have time I can add more of these changes to this branch, which can later be merged for a major release.

Thanks for letting me know :)

benoit-cty (approved):

benoit-cty:
Hello, sorry for the late merge : it was not a breaking change as it was the doc and type hint that was wrong. My bad.

PR #512, opened 2024-03-06
Added return type hints and removed redunant return statements in cpu.py

MyGodItsFull0fStars:
Hi!

I modified a few things in the `cpu.py` file.
First, I either added or changed the return type hints of function signatures depending on the actual return type possible, `Optional[some type]`, if the return type might be `None` in some cases.

Next, a few dangling `return` keywords were at the end of a function definition. As they do not serve any purpose, I removed them.

### DIFF
--- a/codecarbon/core/cpu.py
+++ b/codecarbon/core/cpu.py
@@ -7,7 +7,7 @@ import os
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


benoit-cty (approved):
Thanks !

Issue #222, opened 2021-07-29
Use Electricity Map API in CodeCarbon API to compute emissions


benoit-cty:
Here is the documentation : http://static.electricitymap.org/api/docs/index.html

- Use a Redis cache to call the Electricity Map API.
- Find a way to tell the CC API if it have to compute emission or let the package do it.

PR #517, opened 2024-03-20
Update electricity mix to 2022

benoit-cty:
I ran the notebook to update the data from 2021 to 2022.

Please note that they are less carbon intensity data for 2022 than for 2021 : only 81 countries instead of 205 in 2021.

And we are in 2024 but only have data from 2022...


inimaz (approved):
Do we have a date somewhere in the docs saying when this electricity mix was last updated?
I think it would be helpful. Other than that, looks good. As you said it is a shame that we only have the data from 2022 :(

LuisBlanche (approved):
I think maybe we could try to use the Electricity map free tier to get the latest carbon intensity for all the countries up to date. (we can probably use github actions to automate refreshing the data)

inimaz:
> I think maybe we could try to use the Electricity map free tier to get the latest carbon intensity for all the countries up to date. (we can probably use github actions to automate refreshing the data)

@LuisBlanche indeed! Sounds good! Let's continue that discussion in this issue https://github.com/mlco2/codecarbon/issues/222.

For now, merging this branch since it has already updated the data.

PR #520, opened 2024-03-23
Switching to Hatch for package management

benoit-cty:
Duplicate, will be closed.


SaboniAmine (commented):
Dummy review to test if it is activating checks/

PR #526, opened 2024-04-05
Add i7-11370H TDP

benoit-cty:
Add the TDP of i7-11370H


inimaz (approved):
Looks good to me! 
Source: [Source]( https://www.intel.com/content/www/us/en/products/sku/196655/intel-core-i711370h-processor-12m-cache-up-to-4-80-ghz-with-ipu/specifications.html)

PR #528, opened 2024-04-11
docs: remove personal conversation, fix small typos

amv213:
Hello! This is my first contribution here so do not hesitate to let me know if I am doing anything wrong 🙌🏽 

I have just stumbled on this project, and going through the docs I noticed that there is a section at the end of https://mlco2.github.io/codecarbon/api.html that seems to be a leftover from a personal conversation. 

This is just a very small PR to clean that up. I have also fixed a couple of small typos here and there while I was at it.


amv213:
Sure, happy to remove just that last bit - will do! I had removed the whole block just as a wild guess of where the leftover conversation could have more logically started. It's true that anyway it doesn't really hurt to have that extra information somewhere, so might as well leave it in.

benoit-cty (approved):
Thanks !

Issue #532, opened 2024-04-17
Adding and Choosing TDP values


Lucas-Otavio:
* CodeCarbon version: 2.2.2
* Python version: 3.8.19
* Operating System: Linux

### Description

I was trying to track the emissions of a code and received this message:
```
[codecarbon WARNING @ 14:50:23] No CPU tracking mode found. Falling back on CPU constant mode.
[codecarbon WARNING @ 14:50:24] We saw that you have a Intel(R) Core(TM) i5-2400 CPU @ 3.10GHz but we don't know it. Please contact us.
```

### What I Did and wish was done

I searched for the proper TDP and did a PR to add it to the CSV file, but I wanted to be able to track it with the right reference value without waiting for a new release. It would be great if you could add a parameter to the function, allowing us to choose the TDP.

Moreover, there is a 3 times larger [file](https://github.com/sb-ai-lab/Eco2AI/blob/main/eco2ai/data/cpu_names.csv) on the eco2ai project, which could be very useful to avoid repetitive PRs to add one single TDP. I think I could contribute by writing some code to compare the two, add the missing ones and find inconsistent values (if any).

benoit-cty:
Hello,
It would be great to automate this part, yes !

Another way is to look at this code : https://github.com/Breakend/experiment-impact-tracker/blob/master/experiment_impact_tracker/cpu/data/scrape_tdp_values.py
See https://github.com/mlco2/codecarbon/issues/400

You could already specify your TDP using the parameter `default_cpu_power`.

But I realize that it has not been documented in https://mlco2.github.io/codecarbon/parameters.html , sorry.

Lucas-Otavio:
Thanks for the clarifications.

I checked that code and it seems they used a Wikipedia page that doesn't exist anymore. Maybe it could be achievable using the [Intel one](https://en.wikipedia.org/wiki/List_of_Intel_processors) and the pages on lists of [AMD processors Series](https://en.wikipedia.org/wiki/List_of_AMD_processors#See_also) (such as [this](https://en.wikipedia.org/wiki/List_of_AMD_Duron_processors)).

Lucas-Otavio:
Hello, 

I have begun doing the comparison I mentioned. Merging the 2 files, it reaches 3749 processors. 
Moreover, there are 174 with contrasting values. In 39 of them, the difference seems to be because eco2ai's value is in float, while codecarbon's is in int.

[...]

PR #534, opened 2024-04-23
Update cpu_power.csv

Lucas-Otavio:
As described in issue https://github.com/mlco2/codecarbon/issues/532, I merged the TDP values with the data in the eco2ai project. For contrasting values, a search was made to find the right one or the value from code carbon was kept; except when the difference was inferior to 1 (due to int-float difference), motivating the adoption of eco2ai's float value. The list of contrasting values can be found in the issue.

Reading the diff file, it also seems my code deleted repeated and contrasting lines within the original CSV file. Let me know if there is a reason why they should be maintained or if a given criteria should have been used to select among them.


benoit-cty:
Thank you !

It seems that the value is now in string instead of integer : https://github.com/mlco2/codecarbon/actions/runs/8800751288/job/24152593103?pr=534#step:5:819

Lucas-Otavio:
Hey, sorry for taking some time to come back.
I thought that maybe it happened because of the inclusion of float values, so I accounted for that when reading values from the data frame on core/cpu.py

benoit-cty (approved):
Thanks a lot !

PR #535, opened 2024-04-28
Switching to Hatch for package management

benoit-cty:
I suggest to use Hatch because it will provide a better dependencies management for CodeCarbon.

It also a way to move from setup.py to more modern pyproject.toml

It will also replace `tox`, `venv`, and `make`.

It will simplify the management of the project. For example to run the dashboard locally you just have to do :

```sh
hatch run dashboard:run
```

And Hatch will handle the installation of the needed Python version, the creation of the environment and the package dependencies.

All the requirements are defined id `pyproject.toml` and a lock is generated for different part of the project in a `requirements` folder.

TODO:
- [x] Migrate from setup.py
- [x] Migrate from Tox
- [x] Update contributing
- [x] Update readme
- [x] Update documentation
- [X] Migrate CI to use Hatch instead of Tox
- [x] Remove tox
- [x] Remove setup.py


.github/workflows/build.yml
```diff
@@ -10,20 +10,18 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.11"]
+        python-version: ["3.11"]
```
SaboniAmine, line 13: the python version that hatch uses to package impacts pip distribution & pypi/conda repositories. I don't think it's an issue to remove 3.8 builder however, 3.11 should be fine

.github/workflows/build.yml
```diff
@@ -39,8 +37,10 @@ jobs:
         pip install tox tox-gh-actions
     - name: Unit tests on api with tox
       run: |
-        cd carbonserver/
-        tox -e unit -vv
+        pipx install hatch
+        hatch run api:test-unit
+        # cd carbonserver/
```
SaboniAmine, line 42: remove commented code

docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
       POSTGRES_DB: ${DATABASE_NAME:-codecarbon_db}
       POSTGRES_PASSWORD: ${DATABASE_PASS:-supersecret}
       POSTGRES_USER: ${DATABASE_USER:-codecarbon-user}
-    image: postgres:12
+    image: postgres:13
```
SaboniAmine, line 40: we can go to latest (no major feature upgrade impacts our schema), maybe in a separate PR

pyproject.toml
```diff
@@ -0,0 +1,163 @@
...
+lint = ["black --check --diff {args:.}", "ruff check {args:.}", "mypy {args:.}"]
+format = ["black {args:.}", "ruff --fix --exit-non-zero-on-fix {args:.}"]
+
+[tool.hatch.envs.test]
+features = ["viz"]
+dependencies = [
+   "mock",
+   "pytest",
+   "responses",
+   # "tox",
```
SaboniAmine, line 97: to be removed

SaboniAmine (approved):
Small comments

.github/workflows/build.yml
benoit-cty: I've removed the matrix from Github Actions because, Hatch will take care of it:
```toml
[[tool.hatch.envs.test.matrix]]
# Python 3.7 does not works with pip-tools
python = ["3.8", "3.9", "3.10", "3.11", "3.12"]
```

.github/workflows/build.yml
benoit-cty: Done

docker-compose.yml
benoit-cty: Thanks, I've set it to 16

pyproject.toml
benoit-cty: Done

benoit-cty:
As I have to modify the config of [_checks_](https://github.com/mlco2/codecarbon/settings/branch_protection_rules/17671200), no other PR could be merge before rebasing on this one !

PR #539, opened 2024-05-01
Fix problems with Hatch

benoit-cty:
Following #535 here is some patch to better manage Hatch


pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 name = "codecarbon"
 dynamic = ["version"]
 readme = "README.md"
-license = ""
+license = {file = "LICENSE"}
```
benoit-cty, line 9: It is optional because already in `classifiers`, maybe it's better to remove this line ?

SaboniAmine (approved):
LGTM, thanks!

PR #542, opened 2024-05-04
feat: add logfire as output method

inimaz:
# Goal
Open the discussion on how codecarbon could send the output to logfire, a new monitoring framework https://docs.pydantic.dev/logfire/ 

# How to use this branch

```
pip install logfire
logfire auth
```
This redirects to a page where you can log in with your github user.

Once your logfire account is created, run the script:

```
hatch run python examples/logfire_metrics.py'
```

# Output
Example of the output that I obtain
image


inimaz:
As discussed @SabAmine, here it is a rough example using the log output. What do you think?

inimaz:
Example of a chart on `logfire` using the `codecarbon` data
image

inimaz:
Same example with emissions in g
image

codecarbon/emissions_tracker.py
```diff
@@ -416,21 +419,23 @@ def _init_output_methods(self, api_key):
...
         else:
             self.run_id = uuid.uuid4()
 
         if self._save_to_prometheus:
-            self._cc_prometheus_out = PrometheusOutput(self._prometheus_url)
-            self.persistence_objs.append(self._cc_prometheus_out)
+            self.persistence_objs.append(PrometheusOutput(self._prometheus_url))
+
+        if self._save_to_logfire:
+            self.persistence_objs.append(LogfireOutput())
```
SaboniAmine, line 438: `persistence_objs` seems highly related to the #458 composite output handling. Can we merge concepts in an elegant way ? Maybe @eledhwen would have an idea ?

inimaz:
@SaboniAmine  It should be ready to merge. Now it uses the composite output handling coming from master. And it calls `logfire.configure()` to setup the main instance of logfire (see last commit).

codecarbon/output_methods/http.py
```diff
@@ -41,6 +41,7 @@ class CodeCarbonAPIOutput(BaseOutput):
     run_id = None
 
     def __init__(self, endpoint_url: str, experiment_id: str, api_key: str, conf):
+        self.use_emissions_delta = True
```
SaboniAmine, line 44: On est sûrs de vouloir ça par défaut ? ça va pas changer le comportement chez nos users ?

SaboniAmine (commented):
Just a small but maybe impactful question. Maybe @benoit-cty has an answer ?

codecarbon/output_methods/http.py
inimaz: You are right, sorry this was coming from an implementation before the composite output handling, and I forgot to remove it. Now it is done ==> [063b7f8](https://github.com/mlco2/codecarbon/pull/542/commits/063b7f86af919aa39e91c40267e052f552c800d5)

SaboniAmine (approved):
Thanks for the work!

PR #561, opened 2024-05-28
logging: log the path of the file where emissions are saved

inimaz:
This is useful to know where the `codecarbon` instance is storing the emissions data.

Example of log:
```
[codecarbon INFO @ 17:13:34] Saving emissions data to file /home/inigo/codecarbon/emissions.csv
```


benoit-cty (approved):
Good idea

PR #562, opened 2024-05-28
feat: ensure only one instance of codecarbon is run per machine

inimaz:
Imagine I open 2 clis and run `codecarbon monitor` in each of them. This will result on both of them outputing very similar data to the same output. Now imagine I have multiple instances of codecarbon (e.g. a python script, some cli, some script somewhere that I forgot it is running...). This could become complex rapidly.

# Goal of this PR
By default it will allow only one instance per machine to measure at the same time. 
This is regulated via the parameter `allow_multiple_runs`.

-  If true --> multiple runs of codecarbon can be done in the same machine.
- If false --> only one at the same time.

Default is `false`.

# How to test this

## Option 1:  using 2 consoles
Open 2 consoles and run `hatch run python examples/logging_to_file.py` on both of them.

- Output of Console 1: Normal behavior
- Output of Console 2: 
```
[codecarbon ERROR @ 18:17:11] Error: Another instance of codecarbon is already running. Turn off the other instance to be able to run this one. Exiting.
```

## Option 2:  multiprocess
I have added an example of exclusive run using multiprocess.
Run it via `hatch run python examples/logging_to_file_exclusive_run.py`
Example of output:
```bash
~/codecarbon$ hatch run python examples/logging_to_file_exclusive_run.py 
[codecarbon INFO @ 09:43:43] offline tracker init
[codecarbon INFO @ 09:43:43] offline tracker init
[codecarbon INFO @ 09:43:43] offline tracker init
[codecarbon INFO @ 09:43:43] offline tracker init
[codecarbon ERROR @ 09:43:43] Error: Another instance of codecarbon is already running. Turn off the other instance to be able to run this one. Exiting.
[codecarbon ERROR @ 09:43:43] Error: Another instance of codecarbon is already running. Turn off the other instance to be able to run this one. Exiting.
[codecarbon ERROR @ 09:43:43] Error: Another instance of codecarbon is already running. Turn off the other instance to be able to run this one. Exiting.
[codecarbon INFO @ 09:43:43] offline tracker init
[codecarbon ERROR @ 09:43:43] Error: Another instance of codecarbon is already running. Turn off the other instance to be able to run this one. Exiting.
[codecarbon INFO @ 09:43:43] [setup] RAM Tracking...
[codecarbon INFO @ 09:43:43] [setup] GPU Tracking...
...
```
We see that 5 instances of `codecarbon` are started and 4 end up in error and exit.

### DIFF
--- a/codecarbon/lock.py
+++ b/codecarbon/lock.py
@@ -21,7 +21,7 @@ def acquire_lock():
     try:
         _create_lock_file()
     except FileExistsError as e:
-        logger.error(e)
+        logger.debug("Error:", e)
         logger.error(
             "Error: Another instance of codecarbon is already running. Turn off the other instance to be able to run this one. Exiting."
         )
--- a/codecarbon/lock.py
+++ b/codecarbon/lock.py
@@ -4,10 +4,8 @@ It creates a lock file in /tmp/.codecarbon.lock and removes it on exit.
 If the lock file already exists, it exits the program.
 """
 
-import atexit
 import errno
 import os
-import sys
 import tempfile
 
 from codecarbon.external.logger import logger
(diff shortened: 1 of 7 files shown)


codecarbon/lock.py
```diff
@@ -0,0 +1,60 @@
...
+def acquire_lock():
+    """Acquires a lock to ensure only one instance of codecarbon is running."""
+    try:
+        _create_lock_file()
+    except FileExistsError as e:
+        logger.debug("Error:", e)
+        logger.error(
+            "Error: Another instance of codecarbon is already running. Turn off the other instance to be able to run this one. Exiting."
+        )
+        sys.exit(1)
```
benoit-cty, line 30: I think it will kill the process that use CodeCarbon : It may be better to just stop the CodeCarbon scheduler.

benoit-cty (changes requested):

codecarbon/lock.py
inimaz: @benoit-cty  You are right, thanks! Checkout the new commit in #[1d2329d](https://github.com/mlco2/codecarbon/pull/562/commits/1d2329d6886d314741e57ce333fb2e25eabf9507)
I created a class to store the lock and have the 2 methods `lock.aquire` and `lock.release` handled by EmissionsTracker. Tell me what you think

benoit-cty (approved):

PR #568, opened 2024-06-06
Update impact.csv

### LINKED ISSUE #483: Are GCP emissions overstated b/c of missing CFE in calculation?
Looking at:

- https://github.com/mlco2/codecarbon/blob/master/codecarbon/core/emissions.py#L43C21-L43C21
- https://github.com/mlco2/codecarbon/blob/master/codecarbon/data/cloud/impact.csv
- https://cloud.google.com/sustainability/region-carbon#understanding

I'm curious about the factors used in the calculation. It seems like my entire usage is calculated as if it was using only non-renewable energy. Based on the explanation of the data from the GCP page, it seems like the calculation should scale based on the CFE%

In particular, for the intensity value used in the calculation

> For regions that are similar in CFE%, this will indicate the relative emissions for when your workload is not running on carbon free energy.

Can you please clarify?

I think the following two data centers explained would help me to understand how things are calcluated

| Region | Location | CFE% | intensity |
| -- | -- | -- | -- |
| asia-southeast1 | Singapore | 4% | 372 |
|  us-central1 | Iowa | 92% | 445 |

From the code link & calculation, would Singapore produce a lower carbon value for the same workload, despite the quite opposite CFE%?

eherra:
Resolves #483

Updated the most recent GCP Grid carbon intensity (impact) values to `impact.csv`.

Values copied from:
https://cloud.google.com/sustainability/region-carbon#data

Also found from:
[GoogleCloudPlatform/region-carbon-info, year2022](https://github.com/GoogleCloudPlatform/region-carbon-info/blob/main/data/yearly/2022.csv)


SaboniAmine:
Thanks for your contribution! Globally, impacts are getting lower on GCP. That's good news!

SaboniAmine (approved):

PR #584, opened 2024-06-19
Handle TRES in Slurm

### LINKED ISSUE #569: Could not find mem= after running `scontrol show job $SLURM_JOB_ID` to count SLURM-available RAM. Using the machine's total RAM.
* CodeCarbon version: 2.4.2
* Python version: 3.11.6
* Operating System: Red Hat Enterprise Linux 8.7 (Ootpa)

### Description

I'm trying to evaluate how consumption varies during the inference of an LLM model by changing some generation hyperparameters (temperature, top_k, etc.). For each different configuration of hyperparameters, I run a cycle of 10 generations with 10 different prompts saved in the 'instructions' list, and for each inference, I create a new instance of OfflineEmissionsTracker and start it. I'm running the script on an HPC cluster using SLURM as the workload scheduler. The problem I'm having, as mentioned in the title, is that codecarbon cannot obtain the memory requested in the .sh file and uses the total memory of the node as a reference. How can I solve this?
""
Below is the script for requesting resources and the part of the code where the tracker is instantiated.
""

photo_5989812149381350953_x
photo_5989812149381350952_y

benoit-cty:
Some Slurm environment seems to do not publish `AllocTRES` but only `TRES`.

Will close #569

### DIFF
--- a/codecarbon/external/hardware.py
+++ b/codecarbon/external/hardware.py
@@ -309,6 +309,9 @@ class RAM(BaseHardware):
     def _parse_scontrol(self, scontrol_str):
         mem_matches = re.findall(r"AllocTRES=.*?,mem=(\d+[A-Z])", scontrol_str)
         if len(mem_matches) == 0:
+            # Try with TRES, see https://github.com/mlco2/codecarbon/issues/569#issuecomment-2167706145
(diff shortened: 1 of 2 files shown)


inimaz (approved):
Looks good to me! This will help when there is not `AllocTRES` in the users' SLURM ram scontrol output.

PR #590, opened 2024-06-22
chore(carbonserver) rework api endpoints

prmths128:
Rework the remaining api endpoints to better represent the concepts

=> https://github.com/orgs/mlco2/projects/4/?pane=issue&itemId=67359449

### DIFF
--- a/carbonserver/carbonserver/api/routers/emissions.py
+++ b/carbonserver/carbonserver/api/routers/emissions.py
@@ -66,10 +66,24 @@ def read_emission(
 
 
 @router.get(
-    "/emissions/run/{run_id}", tags=EMISSIONS_ROUTER_TAGS, response_model=Page[Emission]
+    "/runs/{run_id}/emissions",
+    tags=EMISSIONS_ROUTER_TAGS,
+    response_model=Page[Emission],
 )
 @inject
 def get_emissions_from_run(
+    run_id: str,
+    emission_service: EmissionService = Depends(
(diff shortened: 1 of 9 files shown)


carbonserver/carbonserver/api/routers/emissions.py
```diff
@@ -77,3 +79,15 @@ def get_emissions_from_run(
...
+@router.get("/emissions", tags=EMISSIONS_ROUTER_TAGS, response_model=Page[Emission])
+@inject
+def get_emissions(
+    run_id: str,
+    emission_service: EmissionService = Depends(
+        Provide[ServerContainer.emission_service]
+    ),
+    params: Params = Depends(),
+) -> Page[Emission]:
+    return paginate(emission_service.get_emissions_from_run(run_id), params)
```
inimaz, line 93: what if no run_id is provided in the query? How to handle that?

inimaz (commented):
Minor comment, but other than that looks good, thanks!

carbonserver/carbonserver/api/routers/emissions.py
prmths128: run_id is mandatory so you can access emissions by either:
- /runs/{runs[0]['id']}/emissions
- /emissions?run_id=...

If you don't specify run_id in the second variant, you'll get a fastapi validation error

inimaz (approved):
Looks good, thanks!

PR #592, opened 2024-06-22
feat(api): add projects patch and delete methods

inimaz:


prmths128 (approved):

SaboniAmine:
Thanks Inigo! I'm a bit afraid that we ship delete routes without authentication, can we make sure it doesn't happen ?

SaboniAmine (approved):

inimaz:
>Thanks Inigo! I'm a bit afraid that we ship delete routes without authentication, can we make sure it doesn't happen ?
Sure! We should not deploy unless the authentication is in place

PR #595, opened 2024-06-26
feat(api): patch organizations

inimaz:
Continuing on the objectives of the API. This will allow the users to patch their organizations.

### DIFF
--- a/carbonserver/carbonserver/api/infra/repositories/repository_organizations.py
+++ b/carbonserver/carbonserver/api/infra/repositories/repository_organizations.py
@@ -148,6 +148,25 @@ class SqlAlchemyRepository(Organizations):
             )
             return res
 
+    def patch_organization(self, organization_id, organization) -> Organization:
+        with self.session_factory() as session:
+            db_organization = (
+                session.query(SqlModelOrganization)
+                .filter(SqlModelOrganization.id == organization_id)
+                .first()
+            )
+            if db_organization is None:
+                raise HTTPException(
+                    status_code=404, detail=f"Organization {organization_id} not found"
+                )
+
+            for attr, value in organization.dict().items():
+                if value is not None:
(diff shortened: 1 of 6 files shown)


carbonserver/carbonserver/api/infra/repositories/repository_organizations.py
```diff
@@ -148,6 +148,25 @@ def get_organization_detailed_sums(
...
+            )
+            if db_organization is None:
+                raise HTTPException(
+                    status_code=404, detail=f"Organization {organization_id} not found"
+                )
+
+            for attr, value in organization.dict().items():
+                if value is not None:
+                    setattr(db_organization, attr, value)
+            session.commit()
```
prmths128, line 166: don't you need 
 `session.add(db_organization)`
  before ?

inimaz: Good question! I went to check further. Apparently there is no need, when we do `session.query(...)` it adds the object to the persistent state and keeps track of any changes done to it, so if we do `session.add(object)` it will do nothing because it is already in the session. 
From the [sqlalchemy docs](https://docs.sqlalchemy.org/en/20/orm/session_state_management.html)
> Persistent - An instance which is present in the session and has a record in the database. You get persistent instances by either flushing so that the pending instances become persistent, or by **querying the database for existing instances** (or moving persistent instances from other sessions into your local session).

An article that helped: https://albertbyrone.medium.com/sqlalchemy-orm-state-management-827ac4c097dd.
Btw I did the same with `projects` and it works when I test a `patch` in local.

prmths128: alright !

prmths128 (approved):

PR #599, opened 2024-06-30
fix: :hammer: update Dockerfile and Docker -compose for local dev

LuisBlanche:
Some requirements were missing from the `Dockerfile` after migrating to hatch 

- Also added more logs on the entrypoint for debugging 
- Also added depends on pstgres for carbonserver to wait before starting

### DIFF
--- a/carbonserver/docker/Dockerfile
+++ b/carbonserver/docker/Dockerfile
@@ -1,31 +1,29 @@
-# Dockerfile
-
-# Use Ubuntu to install 3 versions of Python for testing
-# For production, you could use python:3.8-slim
 FROM ubuntu:22.04
 
-# set work directory
 WORKDIR /carbonserver
 
-# set env variables
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONUNBUFFERED 1
-# Prevent apt to ask for region
 ENV DEBIAN_FRONTEND noninteractive
 
-# install dependencies
 RUN apt-get update && apt-get upgrade -y && \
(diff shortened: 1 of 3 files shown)


inimaz:
Looks good to me! To test it:
```docker compose up--build -d```

Go to `localhost:8008/docs` to see the FastApi page

File: docker-compose.yml [hunk 1]
```diff
@@ -28,7 +28,7 @@ services:
     networks:
       - codecarbon_net
 
-  postgres:
+  postgres:
```
benoit-cty, line 31:
```suggestion
  postgres:
```

carbonserver/docker/Dockerfile
```diff
@@ -1,31 +1,29 @@
...
-# For production, you could use python:3.8-slim
 FROM ubuntu:22.04
 
-# set work directory
 WORKDIR /carbonserver
 
-# set env variables
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONUNBUFFERED 1
-# Prevent apt to ask for region
```
benoit-cty, line 13: I would have kept the comment.

carbonserver/docker/Dockerfile
```diff
@@ -1,31 +1,29 @@
-# Dockerfile
-
-# Use Ubuntu to install 3 versions of Python for testing
```
benoit-cty, line 3: I would have kept the comment.

benoit-cty (approved):
Thanks for the fix.

I just have some no-blocking comments.

PR #601, opened 2024-07-05
fix(api): python version to 3.11 + fix last alembic upgrade

inimaz:
Fix the alembic downgrades/upgrades to be able to work with the new schema.

# How to test this
1. Add the db url to your env variables
`export DATABASE_URL=YOUR_DB_URL`

1. Downgrade the db to start from scratch
`hatch run api:downgrade-db`
1. Upgrade to the latest status
`hatch run api:setup-db`

(This has been used to deploy to `dev`)


prmths128 (approved):
lgtm

PR #604, opened 2024-07-06
Feat/organization access control

prmths128:
This will add access control to the organizations list
The backward compatibility is maintained for now, in order to keep the PR small and avoid a big overhaul.
With this:
If the user is authenticated, only the users organizations will be listed.

Next steps:
- reject unauthenticated users
- extend the access control to projects, ...
- add tests to the e2e local server (since we can't mock the user there we'll need a API Token auth strategy first)

### DIFF
--- a/carbonserver/carbonserver/api/routers/organizations.py
+++ b/carbonserver/carbonserver/api/routers/organizations.py
@@ -8,6 +8,7 @@ from fastapi import APIRouter, Depends
 from starlette import status
 
 from carbonserver.api.dependencies import get_token_header
+from carbonserver.api.routers.authenticate import UserWithAuthDependency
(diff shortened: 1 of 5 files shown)


carbonserver/carbonserver/api/routers/organizations.py
```diff
@@ -19,6 +19,8 @@
     OrganizationSumsUsecase,
 )
 
+from .authenticate import UserWithAuthDependency
```
SaboniAmine, line 22: lets try to avoid relative imports

SaboniAmine (approved):

inimaz (approved):
Looks good @prmths128 !

Issue #605, opened 2024-07-09
pkg_resources deprecated and moved to setuptools


imciner2:
* CodeCarbon version: 
* Python version: 3.12
* Operating System: Fedora

### Description

Running a simple script leads to the error trace:

```python
File ~/dev/python/codecarbon/.venv/lib64/python3.12/site-packages/codecarbon/__init__.py:6
      [1](https://file+.vscode-resource.vscode-cdn.net/home/imcinerney/dev/python/codecarbon/~/dev/python/codecarbon/.venv/lib64/python3.12/site-packages/codecarbon/__init__.py:1) """
      [2](https://file+.vscode-resource.vscode-cdn.net/home/imcinerney/dev/python/codecarbon/~/dev/python/codecarbon/.venv/lib64/python3.12/site-packages/codecarbon/__init__.py:2) The Carbon Tracker module. The following objects/decorators belong to the Public API
      [3](https://file+.vscode-resource.vscode-cdn.net/home/imcinerney/dev/python/codecarbon/~/dev/python/codecarbon/.venv/lib64/python3.12/site-packages/codecarbon/__init__.py:3) """
      [5](https://file+.vscode-resource.vscode-cdn.net/home/imcinerney/dev/python/codecarbon/~/dev/python/codecarbon/.venv/lib64/python3.12/site-packages/codecarbon/__init__.py:5) from ._version import __version__  # noqa
----> [6](https://file+.vscode-resource.vscode-cdn.net/home/imcinerney/dev/python/codecarbon/~/dev/python/codecarbon/.venv/lib64/python3.12/site-packages/codecarbon/__init__.py:6) from .emissions_tracker import (
      [7](https://file+.vscode-resource.vscode-cdn.net/home/imcinerney/dev/python/codecarbon/~/dev/python/codecarbon/.venv/lib64/python3.12/site-packages/codecarbon/__init__.py:7)     EmissionsTracker,
      [8](https://file+.vscode-resource.vscode-cdn.net/home/imcinerney/dev/python/codecarbon/~/dev/python/codecarbon/.venv/lib64/python3.12/site-packages/codecarbon/__init__.py:8)     OfflineEmissionsTracker,
      [9](https://file+.vscode-resource.vscode-cdn.net/home/imcinerney/dev/python/codecarbon/~/dev/python/codecarbon/.venv/lib64/python3.12/site-packages/codecarbon/__init__.py:9)     track_emissions,
[...]

PR #607, opened 2024-07-10
Python 3.12 support

benoit-cty:
Following #605 this PR goal is to remove dependency to `pkg_resources` to support Python 3.12.

We switch to native `importlib` and add conditional dependency for older version of Python.


SaboniAmine (approved):
LGTM, thanks!

PR #609, opened 2024-07-11
feat(api) allow creating user with id and no password

prmths128:
Used to create user from authentication server's data

One more element to the registration workflow.


inimaz (changes requested):
We can remove altogether the password even from the model, since it will be handled by fief I guess. Or is there a use case to keep saving the `hashed_password`?

prmths128:
Removing it means that having a compatible auth server will be the **only** way to create an account and log in.
That adds a pretty big external dependency on the app.

inimaz (approved):
LGTM!

PR #614, opened 2024-07-19
fix(api): user signup flow

inimaz:
* Remove endpoint `POST /users` in favor of `POST /users/signup`
* Add default organitation to user's organitations list

### DIFF
--- a/carbonserver/carbonserver/api/services/signup_service.py
+++ b/carbonserver/carbonserver/api/services/signup_service.py
@@ -1,12 +1,18 @@
+import logging
 from uuid import UUID
 
 from carbonserver.api.infra.repositories.repository_organizations import (
     SqlAlchemyRepository as OrganizationRepository,
 )
+from carbonserver.api.infra.repositories.repository_projects import (
+    SqlAlchemyRepository as ProjectRepository,
+)
 from carbonserver.api.infra.repositories.repository_users import (
     SqlAlchemyRepository as UserRepository,
 )
-from carbonserver.api.schemas import User, UserCreate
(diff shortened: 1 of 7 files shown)


carbonserver/carbonserver/api/services/signup_service.py
```diff
@@ -38,5 +44,31 @@ def subscribe_user_to_org(
             organization_id, organization_api_key
         )
         if key_is_valid:
-            self._user_repository.subscribe_user_to_org(user, organization_id)
+            user = self._user_repository.subscribe_user_to_org(user, organization_id)
         return user
+
+    def new_user_steps(self, user: User) -> User:
```
SaboniAmine, line 50: can this method be renamed into something more explicit ?
I have trouble naming this method, but here are some suggestions : 
- initialize_org_and_default_project
- register_user_with_default_project

SaboniAmine (approved):
the service / usecase architecture is to be reworked, however this is out of this PR's scope
besides naming it's good for me, thanks!

carbonserver/carbonserver/api/services/signup_service.py
prmths128: `new_user_setup` ?

prmths128 (approved):

carbonserver/carbonserver/api/services/signup_service.py
inimaz: I am more in favor of `new_user_setup`, because we don't know what other steps we might have to do in the future to a new user. What do you think @SaboniAmine ?

SaboniAmine: Yes, lgtm !

PR #622, opened 2024-07-26
Clean users management & add filters on routers

SaboniAmine:
This PR proposes a way to prevent users from accessing data not in their organization.

Current implementation implies that org id should be sent in the request, which prevents a logical check in db. We might want to enforce that, in a standard function applied at router level (and works with project token implementation).
How could we do that ?


inimaz:
Nice cleanup, thanks a lot @SaboniAmine!

About the `organization_id`:
* I think the organization id should not be provided in the request but calculated by the server, even if it costs us one more db call. To let the user send it in the request looks risky. Imagine this scenario
    * User1: Org org_id1 has projects= [project_id1] that has experiments = []
    * User2: Org org_id2 has projects= [project_id2] that has experiments = []

If we let the user send the organization id, then we would be allowing for instance User1 to send in the request `organization_id=org_id1 ` and add experiment to project `project_id2`. 

2 proposals that come to mind now: 

1. for each `project_token_id`, `experiment_id`, `run_id`, `emission_id`... we do a getter, that connects to the db, does the needed joins and retrieves the `organization_id` related to it. And then the logic that you have added to check if it is the right org. **Pros**: less code and we do not affect services/repositories. **Cons**: 1 more db call.
2. We pass the auth_user to each service and verify that it is the right org by doing joins and a filter `organization.id in auth_user.db_user.organizations` in each SQL call. **Pros**: We don't add a db call. **Cons**: a lot more code to add + many joins that might not be needed otherwise + what if we forget to do this in a future new sql model/service

What do you think?

prmths128:
I agree with Inigo and don't think adding organization_id to all calls is the way to go.
This should be managed by a permission service since we retrieve that from the DB.
I'm not 100% clear on the 2 options described but the 1st option seems  better and simpler. And easier to cache.

SaboniAmine:
Indeed, this is not really adapted! 
I would then prefer go with Inigo's first proposition, and extend it to a generic middleware, which could read either a jwt token, a cookie or a project token and return an approbation or not. Here is an example of a middleware, which reimplements OAuth2 (not our target) : https://github.com/code-specialist/fastapi-auth-middleware/blob/main/fastapi_auth_middleware/oauth2_middleware.py or another one with Keycloak : https://github.com/waza-ari/fastapi-keycloak-middleware/blob/main/fastapi_keycloak_middleware/middleware.py
We could try to fit in this middleware the main methods we could use : 
- Token / cookie / project token reading & interpretation. Currently @prmths128 's PR proposes a versatile reader, we could extend it to the project token & have only a single interceptor at the router level (which could be injected at server startup, as a middleware, and then be deactivated for testing or other purposes)
- Scope interpretation. Some filtering data could be exported / synced with fief, such as the organisation list maybe ? We wouldn't need a db call as it should be present in the jwt then.
I'll try to propose something.

prmths128:
Using fief *could* be a way. But that should be tested a bit before.
The simple and foolproof  way would be to check in DB.

We could also add the permission  id in the token.
If we forge the project token ourselves, we can use a JWT with the project id in the token and use that as a permission.
We wouldn't need to check the user/organization in DB anymore. And we would't need to add anything to fief.


As for using a common middleware for all auth. Maybe ? The project token is a bit different from the other permission checks. Should we keep that separate or not ?

PR #624, opened 2024-07-26
Feat: check project token

inimaz:
For each endpoint, it will read from the Headers the property `x-project-token`, retrieve the token from the db and check that is valid and has correct access (READ/WRITE...).

⚠️ I have not deployed it in local yet, nor tested it with a real db.
This is just to show how it looks until now.

We should merge this only **after** https://github.com/mlco2/codecarbon/pull/622 is merged.

- [] Once this flow works we should integrate it with the access_token check and do one or the other. Maybe if `x-project-token` is present, we do the project-token auth check flow?

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ Here is a sample for BibTeX:
                   Marc Alencon and
                   Michał Stęchły and
                   Christian Bauer and
-                  Lucas-Otavio and
+                  Lucas Otávio N. de Araújo and
                   JPW and
                   MinervaBooks},
   title        = {mlco2/codecarbon: v2.4.1},
--- a/carbonserver/carbonserver/api/domain/project_tokens.py
+++ b/carbonserver/carbonserver/api/domain/project_tokens.py
@@ -1,0 +1,17 @@
+import abc
+
+from carbonserver.api import schemas
+
+
+class ProjectTokens(abc.ABC):
+    @abc.abstractmethod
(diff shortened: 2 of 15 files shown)


prmths128:
It should check the project_token OR the web app auth.
Otherwise the user won't be able to access that from the webapp or API.
Unless I missed something

inimaz:
100% agree. I am waiting for the other PR to be merged to implement that OR.

carbonserver/carbonserver/api/routers/emissions.py
```diff
@@ -45,7 +46,16 @@ def add_emission(
     emission_service: EmissionService = Depends(
         Provide[ServerContainer.emission_service]
     ),
+    project_token_service: ProjectTokenService = Depends(
+        Provide[ServerContainer.project_token_service]
+    ),
+    x_project_token: str = Header(None),  # Capture the x-project-token from the headers
```
prmths128, line 52: `x_api_token` instead ?
it's a pretty standard header, we can use that instead of cooking our own variant 
the token's scope might be limited, it's still an api token

prmths128 (approved):

carbonserver/carbonserver/api/routers/emissions.py
inimaz: Done, see [ba08bd4](https://github.com/mlco2/codecarbon/pull/624/commits/ba08bd46a5d0d5c7f2e67e50a8a5a8c10213bbc3)

PR #626, opened 2024-08-01
fix: added the TDP of CPU Intel Core i7-1360P

loekadam:
Added the TDP of the CPU Intel Core i7-1360P to cpu_power.csv
Source: https://www.intel.com/content/www/us/en/products/sku/232155/intel-core-i71360p-processor-18m-cache-up-to-5-00-ghz/specifications.html


benoit-cty (approved):

benoit-cty:
Thanks for your contribution !

PR #589, opened 2024-06-21
Update emissions_tracker.py to fix issue #549

LeahChercham:
Update emissions_tracker.py to fix issue https://github.com/mlco2/codecarbon/issues/549 Divide by zero in start_task() by replacing time.time() by time.perf_counter()

Replaced all time.time() by time.perf_counter() in file emissions_tracker.py.

perf_counter gets the current time with higher precision.

(I had to recreate this pull request as I deleted my repo by mistake: https://github.com/mlco2/codecarbon/pull/570)

PR #628, opened 2024-08-04
release 2.5.1

inimaz:
Create a new release containing the fix of https://github.com/mlco2/codecarbon/pull/589 (that uses `time.perf_counter` instead of `time.time` to avoid divide by zero errors)


benoit-cty (approved):
Thanks !

PR #631, opened 2024-08-08
Added type hints and docstring to GPUDevice

MyGodItsFull0fStars:
I added type hints and a docstring for the `GPUDevice` dataclass.

To comply with support for `Python 3.8` I used the `typing` library instead of the type hints supported from `Python 3.9` onward.


benoit-cty (approved):
Thanks !

PR #634, opened 2024-08-13
Proposed refactoring changes to cpu.py

MyGodItsFull0fStars:
# Proposed Refactoring Changes
This pull request contains proposed changes to `cpu.py`.

## Docstrings

I added some docstrings to methods and classes that are publicly visible.

## CPU Details

In `IntelRAPL`, the variable `self._cpu_details` was initialised, but never used later on, as a new variant without the leading `_` called `self.cpu_details` was introduced. As public functions are available, the `self.cpu_details` mutations were changed to `self._cpu_details`. 

## Pylint
Refactored multiple sections with help of Pylint.

### W0201 - attribute-defined-outside-init

[Pylint source](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/attribute-defined-outside-init.html)

Added `self._windows_exec_backup = None` to `__init__()` of Intel Power Gadget (line 93), as it was defined within a function.

### W0707 - raise-missing-from

[Pylint source](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/raise-missing-from.html)

Changed code in:

```python
except PermissionError as e:
    raise PermissionError(
        "Unable to read Intel RAPL files for CPU power, we will use a constant for your CPU power."
        + " Please view https://github.com/mlco2/codecarbon/issues/244"
        + f" for workarounds : {e}"
    )
        + " for workarounds : %s",
        e,
    ) from e
```

Note, it might make sense to raise another error type instead of raising the same one again, as the `from e` is usually used to chain exceptions and provide a traceback.

### R1734 - use-list-literal

[Pylint source](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/use-list-literal.html)

Replaced `list()` with `[]` as this approach is faster, since the additional function call is omitted.

### R1735 - use-dict-literal

[Pylint source](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/use-dict-literal.html)

Replaced `dict()` with `{}` as this approach is faster, since the additional function call is omitted.

### W1201 - logging-not-lazy

[Pylint source](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-not-lazy.html)

Refactored logging functions to now use lazy string generation.

### DIFF
--- a/codecarbon/core/cpu.py
+++ b/codecarbon/core/cpu.py
@@ -207,7 +207,6 @@ class IntelPowerGadget:
         Placeholder method for starting the Intel Power Gadget monitoring.
         """
         # TODO: Read energy
-        pass
 
 
 class IntelRAPL:
@@ -314,9 +313,9 @@ class IntelRAPL:
                 cpu_details[rapl_file.name] = rapl_file.energy_delta.kWh
                 # We fake the name used by Power Gadget when using RAPL
                 if "Energy" in rapl_file.name:
-                    cpu_details[
-                        rapl_file.name.replace("Energy", "Power")
-                    ] = rapl_file.power.W
+                    cpu_details[rapl_file.name.replace("Energy", "Power")] = (
+                        rapl_file.power.W
+                    )
         except Exception as e:
             logger.info(
                 "Unable to read Intel RAPL files at %s\n \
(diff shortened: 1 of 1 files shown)


benoit-cty:
Thanks !
I would prefer to keep f-string as I find them more readable, and because of [PEP-0498](https://www.python.org/dev/peps/pep-0498). Can you add the option `logging-format-style=fstr` to PyInt ?

benoit-cty (approved):

MyGodItsFull0fStars:
Thanks for accepting the PR! 

To your previous comment, I also prefer the readability of f-strings but yes, this variant is (sadly) more robust.

PR #635, opened 2024-08-14
Removed redundant dangling return keywords from `rapl.py`

MyGodItsFull0fStars:
As they add no logic, I removed the redundant (dangling) `return` keywords in `rapl.py`.


benoit-cty (approved):
Nice catch.

PR #637, opened 2024-08-14
subprocess.Popen now uses a with-statement in powermetrics.py

MyGodItsFull0fStars:
Replaced the regular call from `process = subprocess.Popen(...)` to `with subprocess.Popen(...) as process` as this variant is the recommended approach for resource allocating processes.

By using 'with' the release of the allocated resources is ensured even in the case of an exception. Source: [Pylint R1732](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/consider-using-with.html).


benoit-cty (approved):
Thanks !

MyGodItsFull0fStars:
Thank you for merging my PR 🥳

PR #638, opened 2024-08-15
Added type hint co2 signal

MyGodItsFull0fStars:
Added type hints for constants as well as `get_emissions(...)` function return type.
Additionally added docstring to `get_emissions(...)` function.


benoit-cty (approved):
Thanks !

PR #643, opened 2024-08-27
init refacto for functional filtering in api

SaboniAmine:
This PR is a cleaner version of #622, handling web authentication flow to read personal resources.


prmths128 (approved):
tldr: lgtm

remarks: I'm not fond of removing the password hash and api user creation path.
That means we commit further to having fief as a hard dependency.
That's ok for now but I don't think that's a good idea long term.
That's mostly tomorrow's problem though, for now that still cleaner.

PR #649, opened 2024-09-02
Fixed duration values in task monitoring

Gommarlo:
Fixed a bug when measuring task duration: formerly it used epoch time as a start value, while every other instance of time measurement uses relative time. This resulted in inconsistent values.

Replacing time.time() with time.perf_counter() solves the issue.


benoit-cty:
Thanks for reporting this.
Do you think it will be better to also use it for https://github.com/mlco2/codecarbon/blob/f16c986aa7cb52f6b4e7439fbc100691a1339432/codecarbon/core/measure.py#L28 ?

Gommarlo:
My suggestion here is to not use time.time() at all for multiple reasons:

- time.time() suffers from changes of the system clock (daylight savings, sync with an external clock, leap seconds...)
- time.time() usually has less precison

Another thing to keep in mind, is that the fix I provided might not be the ideal one: time.perf_counter() is the highest precision you can get when measuring time (along with time.perf_counter_ns()) which implies a larger overhead when accessing the clock.

Another option is to use time.monotonic() which has significantly less overhead, does not have time.time()'s issues, but also has a lower precision.

Depending on your requirements you should use whichever you see fit. 

In my humble opinion, I believe that time.monotonic() should be the standard duration tracker due to Codecarbon's goal of being as lightweight as possible, but for tasks it might be helpful to have higher precision as I may want to run short energy-intensive code (or you could allow the user to choose which timer to use)

codecarbon/core/api_client.py
```diff
@@ -37,7 +37,6 @@ class ApiClient:  # (AsyncClient)
     """
 
     run_id = None
-    _previous_call = time.time()
```
benoit-cty, line 40: Deleted because I don't see usage of this.

benoit-cty (approved):
Thanks for your answer, I've updated all calls to time().

benoit-cty:
This will be the 2.7.0 release of CodeCarbon.

PR #657, opened 2024-09-08
fix(api) fix main user flows

prmths128:
- fix mandatory jwt token (=> optional)
- fix return type during signup crashing the first login and preventing membership


SaboniAmine (approved):
Thanks!

PR #660, opened 2024-09-09
fix(webapp) fix frontend after backend changes

prmths128:
switch some pages from server to client mode


SaboniAmine (approved):
Lets merge this temporary fix before rewriting the client / server setups

PR #661, opened 2024-09-10
doc(api): update local deployment procedure

SaboniAmine:
- Add env var example file for the API (no autoload from .env)
- Downgrade fastapi_pagination temporarily to avoid bug introduced in latest fastapi release (https://github.com/uriyyo/fastapi-pagination/issues/1273)


inimaz (approved):
Thanks a lot @SaboniAmine! This will solve the fastApi `dependant` issue in our master  branch.

PR #669, opened 2024-09-13
draft: docker compose UI + fief

inimaz:
# How to test this
```
docker compose up --build -d
```
## Edit it with your host name
Know your hostname by running in bash
```sh
hostname
```
Search inigopc in all the repo and change inigopc with your hostname.

## Add the backend URL to the redirect-URIs list of the admin client
Go to `http://YOUR_HOSTNAME.local:8000/admin`. Then search for Clients --> Check the client is the same as the clientID in `.env.fief`, edit it and add the redirectURI of the backend, `http://YOUR_HOSTNAME.local:8001/auth/login`. In my case
image

## Current Issue (Almost working!)
I am able to arrive to the main page, but the call of `Next.js` to `/organizations` returns `[]`. On the other hand, if I do the same call with the API (Try it [http://YOUR_HOSTNAME.local:8001/api/docs]()) it returns the good organization.
I see the cookie of user_session. The cookie is there!

But if I add logs inside `UserWithAuthDependency` all the tokens are `None` when the UI calls `/organizations`. So somehow it is not reading the cookie somewhere in the server.

@SaboniAmine @prmths128 any ideas?


prmths128:
forked and merged elsewhere
closing

PR #671, opened 2024-09-16
fix(webapp) fix app crashing when fetching organizations list without auth

prmths128:
note: this function is not going to work anyway since it has no authentication
I'm not going to fix it now so I'm adding a safe default return value in case of error, thus preventing the app from crashing (but not fixing the auto redirect to the default organization page when applicable - now that I think of it we might want to get rid of that later).


SaboniAmine (approved):
Thanks !

Issue #673, opened 2024-09-20
Is it right that the Carbon Intensity of Solar, Wind, ... is ignored in your calculations?


headscott:
* CodeCarbon version: 2.5.0
* Python version: 3.12.4
* Operating System: Linux

### Description

I am about to write a paper about sustainability of AI for my university. Therefore I wanted to know what Electricity mix (Percentages of Coal, Petroleum, Natural Gas, Solar, ...) your code is guessing and how it gets the values/where are the guessed values from?

Where can I get more current information about the german energy mix/Where did you get it from?

### What I Did

I saw in `/codecarbon/core/emissions.py` that there is a function called `_region_energy_mix_to_emissions_rate` which only uses coal, petroleum and natural gas. But on your website [CodeCarbon](https://mlco2.github.io/codecarbon/methodology.html) there is also mentioned emissions for Solar, Wind, ...

benoit-cty:
Hello,

The function you mention is only used in the case of an error DataSourceException, which is not supposed to happen anymore.

Our data came from [our_world_in_data](https://github.com/mlco2/codecarbon/blob/master/codecarbon/data/private_infra/our_world_in_data.ipynb) and you can see them in the file https://github.com/mlco2/codecarbon/blob/master/codecarbon/data/private_infra/global_energy_mix.json

We could also use ElectricityMaps to get the energy mix in realtime.
 
Let us know when your paper is out !

headscott:
Is there a way for me to do something like `get_regional_energy_mix(ISO_Code, Region)` and get the same results that you used to calculate the CO2eq of the power consumption? If so, what should I do?

I tried this code, but it didn't work:

from codecarbon.external.geography import GeoMetadata
from codecarbon.input import DataSourceException
from codecarbon import EmissionsTracker

def get_my_energy_mix():
    # create EmissionTracker Instance
    tracker = EmissionsTracker()

    # define GeoMetadata for your region
    geo = GeoMetadata(
        country_iso_code,
        country_name,
        region,
        latitude,
        longitude
    )

    try:
        country_emissions_data = tracker._data_source.get_country_emissions_data(
            geo.country_iso_code.lower()
        )

        if geo.region not in country_emissions_data:
            # TODO: Deal with missing data, default to something
            raise ValueError(
                f"Region: {geo.region} not found for Country "
                + f" with ISO CODE: {geo.country_iso_code}"
            )

        print("Energy Mix for country:", country_emissions_data)

    except DataSourceException:
[...]

PR #674, opened 2024-09-20
update python script to retrieve our word in data's data

MarionCoutarel:
Change the our_world_in_data script so the global_energy_mix.json now holds the last available data regarding the carbon intensity of electricity per country.

For exemple, as carbon intensity of electricity was not available for Albania in 2023, Albania was not in the former global-energy_mix.json. Albania now appears with its carbon intensity of electricity of 2022.

It allows to integrate 135 new coutries (or region).

5 out 135 with data older than 2022 : 2021 for overseas french regions (French Guiana, Guadeloupe, Martinique, Reunion) and 2009 for western Sahara


benoit-cty:
Thanks Marion !

A test is failing because it use JOR (Jordana) as a test and we know have data for it.

Can you check if we have another test for countries without data ? Ti know if we update the result for JOR or if we have to switch to another unknown country code ?

benoit-cty:
It may be interesting to have a look to https://github.com/mlco2/codecarbon/issues/673#issuecomment-2386704373

benoit-cty (approved):

PR #675, opened 2024-09-21
feat(cli,api) add auth to cli for new api

prmths128:
WIP

TODO: add http://localhost:8080& to fief redir uri
Needs public pkce oauth2 app on fief 

This will add fief auth to the cli. This allows the user to manage the projects.
And allow creating an api token to send emissions for a given project.

### DIFF
--- a/codecarbon/cli/main.py
+++ b/codecarbon/cli/main.py
@@ -3,7 +3,10 @@ from pathlib import Path
 from typing import Optional
 
 import questionary
+import requests
 import typer
+from fief_client import Fief
+from fief_client.integrations.cli import FiefAuth
 from rich import print
 from rich.prompt import Confirm
 from typing_extensions import Annotated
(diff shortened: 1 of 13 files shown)


codecarbon/cli/main.py
```diff
@@ -20,6 +23,10 @@
 from codecarbon.core.schemas import ExperimentCreate, OrganizationCreate, ProjectCreate
 from codecarbon.emissions_tracker import EmissionsTracker
 
+AUTH_CLIENT_ID = "pkqh9CiOkp4MkPqRqM_k8Xc3mwBRpojS3RayIk1i5Pg"
```
benoit-cty, line 26: Isn't it dangerous to put this code in public ?

prmths128: The secret yes (used in the backend on /auth/login), the id no. It's a public information
That being said we still have some work to streamlmine and secure the auth process.

SaboniAmine: This is a specific client for public usage of the authentication system.
This client ID is not secret, as it is only used to log in the user and after the user credentials will be stored in the credentials.json file, and use a different authentication workflow.
So it should be okay (and also is the recommended way to authenticate CLIs from the fief documentation : https://docs.fief.dev/integrate/python/cli/)

SaboniAmine (approved):

PR #685, opened 2024-10-10
fix: infinite load in project components

inimaz:
This is to fix the infinite loading of some project components


SaboniAmine (approved):
Thanks !!

PR #688, opened 2024-10-15
feat: hashed passwords + alembic migration

inimaz:
# 2 Flows
Create token flow:
1. Generate a key
2. Get a look_up value (to make it easier/faster to read from the db)
3. Store hashed key with lookup value

Verify token flow:
1. Get lookup value of the incoming token
2. Look up by that value all possible candidates + other params (e.g. project_id)
3. Hash the incoming token
4. Verify that the hash key is the same as the stored hashed key

Why use `bcrypt` instead of passlib? It seems passlib is abandoned, in FastApi they recommend using bcrypt ==>
https://github.com/fastapi/fastapi/discussions/11773#discussioncomment-10640267

# TODO
- [x] Test create project token + list + delete
- [x] Test post run + post emission (will use the verification of the token)

### DIFF
--- a/carbonserver/carbonserver/api/infra/database/sql_models.py
+++ b/carbonserver/carbonserver/api/infra/database/sql_models.py
@@ -183,14 +183,13 @@ class ProjectToken(Base):
     project = relationship("Project", back_populates="project_tokens")
     # Dates
     last_used = Column(DateTime, nullable=True)
-    expiration_date = Column(DateTime, nullable=False)
     # Permissions
     access = Column(Integer)
 
     def __repr__(self):
         return (
             f'<ApiKey(project_id="{self.project_id}", '
-            f'api_key="{self.hashed_token}", '
+            f'hashed_token="{self.hashed_token}", '
             f'name="{self.name}", '
             f'revoked="{self.revoked}", '
             f'last_used="{self.last_used}", '
--- a/carbonserver/carbonserver/api/infra/database/sql_models.py
+++ b/carbonserver/carbonserver/api/infra/database/sql_models.py
@@ -177,8 +177,10 @@ class ProjectToken(Base):
     id = Column(UUID(as_uuid=True), primary_key=True, index=True, default=uuid.uuid4)
     project_id = Column(UUID(as_uuid=True), ForeignKey("projects.id"))
     name = Column(String)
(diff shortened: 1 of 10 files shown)


carbonserver/carbonserver/api/infra/database/sql_models.py
```diff
@@ -177,17 +177,22 @@ class ProjectToken(Base):
...
     last_used = Column(DateTime, nullable=True)
+    expiration_date = Column(DateTime, nullable=False)
     # Permissions
     access = Column(Integer)
 
     def __repr__(self):
         return (
             f'<ApiKey(project_id="{self.project_id}", '
-            f'api_key="{self.token}", '
+            f'api_key="{self.hashed_token}", '
```
prmths128, line 193: f'hashed_token="{self.hashed_token}", '

prmths128 (approved):

carbonserver/carbonserver/api/infra/database/sql_models.py
```diff
@@ -177,17 +177,22 @@ class ProjectToken(Base):
...
     project_id = Column(UUID(as_uuid=True), ForeignKey("projects.id"))
     name = Column(String)
-    token = Column(String, unique=True)
+    hashed_token = Column(String,nullable=False)
+    lookup_value = Column(String,nullable=False) # This is the first 8 characters of the SHA-256 hash of the API key. Used for filtering faster
+    revoked = Column(Boolean, default=False)
     project = relationship("Project", back_populates="project_tokens")
+    # Dates
     last_used = Column(DateTime, nullable=True)
+    expiration_date = Column(DateTime, nullable=False)
```
SaboniAmine, line 186: lets not handle token validity by date, we have no need for this complexity

SaboniAmine (commented):
That seems great, thanks! I would remove the expiration date for the token, which is not yet a constraint for token generation.

carbonserver/carbonserver/api/infra/database/sql_models.py
inimaz: Removed! See [fe6dcfe](https://github.com/mlco2/codecarbon/pull/688/commits/fe6dcfe13b45a9915d8990a0fc780e4fc00e913f)

PR #691, opened 2024-10-19
fix: starlette version compatible with fastapi

inimaz:
Starlette was not letting fast-api deploy because incompatible versions.
In FastAPI they say not to fix the version of starlette https://fastapi.tiangolo.com/deployment/versions/#about-starlette
@benoit-cty  How can we handle this with `hatch`?


prmths128 (approved):

benoit-cty:
Hello, maybe https://github.com/jazzband/pip-tools/issues/1656#issuecomment-1194757900?

PR #699, opened 2024-10-28
Build with Hatch instead of `build`

benoit-cty:
- Build with Hatch instead of `build`
- Publish a new version of CodeCarbon


SaboniAmine (approved):
Thanks for the upgrade!

Issue #430, opened 2023-07-20
Fix Conda Build


benoit-cty:
The conda build fail because of a missing dependency : `ModuleNotFoundError: No module named 'prometheus_client'`

See https://github.com/conda-forge/codecarbon-feedstock/pull/16

benoit-cty:
The conda build failed also for CodeCarbon Channel:

```
Run uibcdf/action-build-and-upload-conda-packages@v1.3.0
Run echo "::group::Information of the trigger event"
Information of the trigger event
Run echo "::group::Checking if the file meta.yaml exists"
Checking if the file meta.yaml exists
Run echo "::group::Building conda package for host platform"
Building conda package for host platform
  conda build . --no-anaconda-upload --output-folder /tmp/compilation-ZRiiddAXIN --python 3.11
  usage: conda [-h] [-v] [--no-plugins] [-V] COMMAND ...
  conda: error: argument COMMAND: invalid choice: 'build' (choose from 'activate', 'deactivate', 'clean', 'compare', 'config', 'create', 'env', 'export', 'info', 'init', 'install', 'list', 'notices', 'package', 'remove', 'uninstall', 'rename', 'run', 'search', 'update', 'upgrade', 'content-trust', 'doctor', 'repoquery')
  Error: Process completed with exit code 2.
```

Maybe we could move to rattler-build that seems more stable.

inimaz:
Solved by #700

PR #700, opened 2024-10-30
[CI] Fix Conda build

benoit-cty:
This PR is for switching to Rattler-Build for faster and more reliable Conda build, which stop working a long time ago...

I forced the execution in CI and it works : https://github.com/mlco2/codecarbon/actions/runs/11596188289

See #430 

I also add explicit support for Python 3.13.


.github/workflows/build.yml
```diff
@@ -6,21 +6,67 @@ on:
     branches: [master]
 
 jobs:
-  build:
```
benoit-cty, line 9: I rename it to "python-test" as it best describe what this step does.

inimaz (approved):
💯 Thanks! Important to get codecarbon working in conda.

PR #704, opened 2024-11-06
release: 2.7.4

inimaz:


benoit-cty:
I will prefer to have only a 2.7.3

And maybe we could improve the message : ` f"Lock file already exists. Path: {LOCKFILE}. This usually means another instance of codecarbon is running."` to  `f"Lock file {LOCKFILE} already exists. This usually means another instance of codecarbon is running. You can safely delete it if you want or use allow_multiple_runs parameter to always bypass it."`

benoit-cty (approved):
Thanks changing the version number.

PR #706, opened 2024-11-10
fix: pin versions of hatch and hatchling

inimaz:
Known issue in hatchling does not allow hatch to build the package. 
Issue in hatchling ==> https://github.com/pypa/hatch/issues/1793

# Workaround
Fix the versions of hatch and hatchling (I have already done it in the other CI jobs, this is the only one remaining)


benoit-cty (approved):
bad luck you face that problem.

Issue #567, opened 2024-06-06
Adding support for CUDA_VISIBLE_DEVICES


lipengfeizju:
* CodeCarbon version: 2.4.2
* Python version: 3.9.19
* Operating System: Rocky Linux release 8.8 (Green Obsidian)

### Description

In our university's cluster, our goal is to measure the energy consumption of a deep learning model. The server uses SLURM system and we only get 1 A100 ( out of 8 GPUs). The GPU power measurement from codecarbon is about all 8 GPUs, instead of the GPU we have been allocated. 

Techinically speaking, I guess in line 184 of `codecarbon/core/gpu.py`, it queries all GPUs from nvml instead of focusing on the GPU we are actually using. To get a more accurate measurement, it would be better to only look up the power consumption related to `CUDA_VISIBLE_DEVICES`. 

Similar discussion about this topic can also be found [here](https://github.com/gpuopenanalytics/pynvml/issues/28) in pynvml.

So is it possible to add a new feature to support measurements focusing on `CUDA_VISIBLE_DEVICES`? I think this is important for deep learning applications, since the other non-visiable devices are usually unrelated to the power consumption of the DL applications. 

Thank you again for providing the code base for carbon measurement.

inimaz:
Hello @lipengfeizju! Thanks for using `codecarbon`.
I didn't know about this var `CUDA_VISIBLE_DEVICES`. If they end up implementing the function in `pynvml` it would be useful to use it indeed.

In the meantime, in `codecarbon` there is a way to filter the GPUs that are tracked if you provide their `gpu_id`.  https://github.com/mlco2/codecarbon/blob/master/codecarbon/emissions_tracker.py#L191
So maybe you can get the id via `nvidia-smi` of the ones you want to use and do
```python
EmissionsTracker(
             ...
             gpu_ids ="0,3,4"
)
```
Is this what you need?

lipengfeizju:
Sorry to reopen the issue again, is it possible to measure the power of several specific CPU cores? (Maybe just like we do for the GPU ids)

benoit-cty:
Maybe we could initialize `gpu_ids` with `os.environ("CUDA_VISIBLE_DEVICES")` ?

For CPU node could you open another issue and provide the codecarbon debug logs ? Because it's not possible yet but maybe we could imagine a way to do it.

lipengfeizju:
Thanks!

PR #712, opened 2024-11-15
Use CUDA_VISIBLE_DEVICES by default

benoit-cty:
Following https://github.com/mlco2/codecarbon/issues/567 it will be better for CodeCarbon to use the env variable CUDA_VISIBLE_DEVICES as a default value for the parameter gpu_ids.

I also drop Python 3.8 from CI as it is deprecated https://devguide.python.org/versions/ and use the same Python version everywhere in CI for consistency. I kept 3.8 in supported version to allow our users to continue to use CodeCarbon in 3.8 until it stop working.


codecarbon/external/hardware.py
```diff
@@ -97,14 +97,25 @@ def measure_power_and_energy(
...
+                if gpu_id in self.gpu_ids:
+                    gpu_ids.append(gpu_id)
+                else:
+                    logger.info(
+                        f"GPU number {gpu_id} will not be monitored, at your request."
+                    )
             gpu_ids = self.gpu_ids
-            assert set(gpu_ids).issubset(
-                set(range(self.num_gpus))
-            ), f"Unknown GPU ids {gpu_ids}"
```
benoit-cty, line 107: We stop raising an AssertionError that will break the user code. Instead, we just log a warning above.

tests/test_gpu.py
```diff
@@ -210,19 +210,24 @@ def test_gpu_metadata_total_power(self):
...
+        gpu1_energy3 = Energy.from_millijoules(energy_consumption[4])
+        # gpu2_energy1 = Energy.from_millijoules(energy_consumption[1])
+        gpu2_energy2 = Energy.from_millijoules(energy_consumption[3])
+        gpu2_energy3 = Energy.from_millijoules(energy_consumption[5])
+
+        # gpu2_power = Power.from_energies_and_delay(gpu1_energy1, gpu1_energy2, Time(5))
+        gpu2_power2 = Power.from_energies_and_delay(gpu1_energy2, gpu1_energy3, Time(5))
+        # gpu1_power = Power.from_energies_and_delay(gpu2_energy1, gpu2_energy2, Time(5))
+        gpu1_power2 = Power.from_energies_and_delay(gpu2_energy2, gpu2_energy3, Time(5))
+        expected_power = gpu1_power2 + gpu2_power2
```
benoit-cty, line 225: I need to do this because it seems the first calls to `nvmlDeviceGetTotalEnergyConsumption` are lost because of the new code of hardware.py:from_utils that made two calls to the class :
Previous code: `        return cls(gpu_ids=gpu_ids)`

New code:
```py
        gpus = cls(gpu_ids=gpu_ids)
        new_gpu_ids = gpus._get_gpu_ids()
        if len(new_gpu_ids) < gpus.num_gpus:
            logger.warning(
                f"You have {gpus.num_gpus} GPUs but we will monitor only {len(gpu_ids)} of them. Check your configuration."
            )
        return cls(gpu_ids=new_gpu_ids)
```

tests/test_emissions_tracker.py
```diff
@@ -95,7 +95,7 @@ def test_carbon_tracker_TWO_GPU_PRIVATE_INFRA_CANADA(
         self.assertGreaterEqual(
             mocked_get_gpu_details.call_count, 2
         )  # at least 2 times in 5 seconds + once for init >= 3
-        self.assertEqual(2, mocked_is_gpu_details_available.call_count)
+        self.assertEqual(3, mocked_is_gpu_details_available.call_count)
```
benoit-cty, line 98: I need to add one call because of the change in `from_utils` in `hardware.py`.

codecarbon/external/hardware.py
```diff
@@ -97,14 +97,25 @@ def measure_power_and_energy(
...
         :return: list of ids
         """
+        gpu_ids = []
         if self.gpu_ids is not None:
-            gpu_ids = self.gpu_ids
-            assert set(gpu_ids).issubset(
-                set(range(self.num_gpus))
-            ), f"Unknown GPU ids {gpu_ids}"
+            # Check that the provided GPU ids are valid
+            if not set(gpu_ids).issubset(set(range(self.num_gpus))):
```
inimaz, line 106:
```suggestion
            if not set(self.gpu_ids).issubset(set(range(self.num_gpus))):
```

inimaz (changes requested):
Thanks for this @benoit-cty !
Just small typo in one of the checks.

SaboniAmine (commented):
LGTM, thanks Benoît! I'll let Inigo approve when his change request is done.

codecarbon/emissions_tracker.py
```diff
@@ -382,7 +390,8 @@ def set_CPU_GPU_ram_tracking(self):
             self._conf["gpu_model"] = "".join(
                 [f"{i} x {name}" for name, i in gpu_names_dict.items()]
             )
-            self._conf["gpu_count"] = len(gpu_devices.devices.get_gpu_static_info())
+            if self._conf.get("gpu_count") is None:
+                self._conf["gpu_count"] = len(gpu_devices.devices.get_gpu_static_info())
```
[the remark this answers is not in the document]
benoit-cty: If pynvml fail in `get_gpu_static_info()`, the result will be an empty list so `self._conf["gpu_count"]` will be set to 0.

codecarbon/external/hardware.py
benoit-cty: Nice catch !

inimaz (approved):

PR #722, opened 2024-11-26
WIP feat(cli) add option to disable api in command line

prmths128:
Ask the user to enable/disable the api during config

### DIFF
--- a/codecarbon/cli/main.py
+++ b/codecarbon/cli/main.py
@@ -3,6 +3,7 @@ import sys
 import time
 from pathlib import Path
 from typing import Optional
+from uuid import uuid4
 
 import questionary
 import requests
@@ -16,6 +17,7 @@ from typing_extensions import Annotated
 from codecarbon import __app_name__, __version__
 from codecarbon.cli.cli_utils import (
     create_new_config_file,
+    get_api_enabled,
     get_api_endpoint,
     get_config,
     get_existing_local_exp_id,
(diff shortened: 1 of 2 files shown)


codecarbon/cli/main.py
```diff
@@ -177,13 +179,29 @@ def config():
...
     api_endpoint = get_api_endpoint(file_path)
     api_endpoint = typer.prompt(
-        f"Current API endpoint is {api_endpoint}. Press enter to continue or input other url",
+        f"Current API endpoint is at {api_endpoint}. Press enter to continue or input other url",
         type=str,
         default=api_endpoint,
     )
     overwrite_local_config("api_endpoint", api_endpoint, path=file_path)
+
+    print(f"Logging in to auth server at {AUTH_SERVER_URL}")
```
benoit-cty, line 202:
```suggestion
    print(f"Logging into the auth server at {AUTH_SERVER_URL}")
```

codecarbon/cli/main.py
```diff
@@ -2,6 +2,7 @@
 import time
 from pathlib import Path
 from typing import Optional
+from uuid import uuid4
```
benoit-cty, line 5: This import seems to not be used, right ?

benoit-cty (approved):
Thanks, it's a nice option.

codecarbon/cli/main.py
prmths128: I imported that to set a random experiment id, since I thought the experiment id might be needed even with no api.
Not sure though, I have to check.

benoit-cty (changes requested):
Hi @prmths128,
`fief_auth` doesn't exist.

And what do you think of using existing `save_to_api` instead of creating  `api_enabled` ?

inimaz:
Closed as already implemented

PR #725, opened 2024-12-03
Fix API calls

benoit-cty:
Since 6 month the API was not called at regular intervals, only at the end.


inimaz (approved):
Thanks for the fix @benoit-cty !

PR #726, opened 2024-12-03
feat: log where the config file is

inimaz:
# Example of output
```
[codecarbon INFO @ 10:47:11] Codecarbon is taking the configuration from the local file: /home/inigo/personal/codecarbon/.codecarbon.config
[codecarbon INFO @ 10:47:11] [setup] RAM Tracking...
[codecarbon INFO @ 10:47:11] [setup] GPU Tracking...
[codecarbon INFO @ 10:47:11] Tracking Nvidia GPU via pynvml
```

### DIFF
--- a/codecarbon/core/config.py
+++ b/codecarbon/core/config.py
@@ -3,6 +3,8 @@ import os
 from pathlib import Path
 from typing import List
 
+from codecarbon.external.logger import logger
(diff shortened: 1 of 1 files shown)


benoit-cty (approved):
Thanks

Issue #724, opened 2024-12-02
[UI] Wrong dates in emissions chart


inimaz:
![image](https://github.com/user-attachments/assets/0dda7d37-190c-4cbd-827e-9e1253762831)

The emissions charts is showing some weird dates. Goal is to show the real dates

PR #727, opened 2024-12-03
fix: scatter chart shows correct timestamp

inimaz:
Closes https://github.com/mlco2/codecarbon/issues/724.

Improvements in scatter chart:

- Sort data by timestamp
- Correct timestamp displayed
- Color
- Units

### DIFF
--- a/webapp/src/components/runs-scatter-chart.tsx
+++ b/webapp/src/components/runs-scatter-chart.tsx
@@ -1,12 +1,4 @@
-import {
-    ScatterChart,
-    Scatter,
-    XAxis,
-    YAxis,
-    Tooltip,
-    Label,
-    ResponsiveContainer,
-} from "recharts";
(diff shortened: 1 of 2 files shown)


inimaz:
This branch is deployed in dev.

benoit-cty:
Great !
Can you also change the color of the _energy_ trace because it quite invisible in grey on grey:
image

And add the hh:mm:ss in the tooltip ?
Because while looking at the timeseries it's important to know the time.

SaboniAmine:
> Great ! Can you also change the color of the _energy_ trace because it quite invisible in grey on grey: 

The tooltip & the color issues are fixed in a PR I didn't push yet, also changing the run scatter chart to a time series one similar to the emissions chart

SaboniAmine (approved):
Thanks!

PR #741, opened 2024-12-12
Add a ResourceTracker to have cleaner code

benoit-cty:
This PR just move BaseEmissionsTracker.set_CPU_GPU_ram_tracking to ResourceTracker.

PR #740, opened 2024-12-12
config(codecarbon) update cli auth settings

prmths128:

PR #745, opened 2024-12-14
Release 2.8.2

benoit-cty:
This is the version bump.

I also refactor the `CONTRIBUTING.md` because some sections was not in the right order/level.

## What’s new in 2.8.2 ?

* config(codecarbon) update cli auth settings (#740) @prmths128
* Add a ResourceTracker to have cleaner code (#741) @benoit-cty
* chore(carbonserver) more cors (#739) @prmths128
* chore(carbonserver) update cors config (#738) @prmths128
* Add SmolLM2 example (#734) @benoit-cty
* fix(ui): units are now kg and kWh (#728) @inimaz
* fix: scatter chart shows correct timestamp (#727) @inimaz


SaboniAmine:
Can we wait a bit for the release ? I still have issues with package -> production api setup and feel like we can ship the fix in this version.

SaboniAmine (approved):
We're all good !

PR #746, opened 2024-12-14
feat(api): add auth to project token endpoints

inimaz:
Add authentication to the project-token endpoints.

FE changes:
* remove `use server` from api tokens calls

BE changes
* add authentication and authorization checks on the project-token endpoints
* env variable CORS_ORIGIN
* Now we run the api unit tests in every PR


prmths128 (approved):

carbonserver/carbonserver/api/routers/authenticate.py
```diff
@@ -84,6 +84,10 @@ async def get_login(
 
         # check if the user exists in local DB ; create if needed
         if "id_token" not in res.json():
+            if "access_token" not in res.json():
```
inimaz, line 87: This is to add more logging when there is no access token in the response from fief

.github/workflows/build.yml
```diff
@@ -88,7 +88,7 @@ jobs:
     - name: Unit tests on api
       run: |
         pip install hatch==1.13.0 hatchling==1.25.0
-        # hatch run api:test-unit
+        hatch run api:test-unit
```
inimaz, line 91: We run the api unit tests now

carbonserver/main.py
```diff
@@ -94,6 +96,10 @@ def init_server(container):
     server.include_router(emissions.router)
     add_pagination(server)
 
+    # Add CORS from env variable
+    CORS_ORIGINS = os.environ.get("CORS_ORIGINS", "")
```
inimaz, line 100: To be able to add custom CORS_ORIGINS via env variables

carbonserver/carbonserver/api/routers/authenticate.py
```diff
@@ -84,6 +84,10 @@ async def get_login(
...
+                print("Access token not found in response. Response:")
```
SaboniAmine, line 88: Can we remove those print statements ?

SaboniAmine (approved):
LGTM, thanks!

carbonserver/carbonserver/api/routers/authenticate.py
inimaz: Done! --> [49e1d7f](https://github.com/mlco2/codecarbon/pull/746/commits/49e1d7f2287d4d9a07a3261b926dc2db734b57d4)

Issue #757, opened 2024-12-30
Organization is None after creation


FrejaThoresen:
* CodeCarbon version: v2.8.2
* Python version: 3.12.8
* Operating System: Linux Ubuntu

### Description

After creating a config and following the prompts, an error of None. The error happens after I have already run 

```
carboncode login
```

and successfully logged in. 

### What I Did

```
codecarbon config
Welcome to CodeCarbon configuration wizard
Existing global config file found :
Current configuration : 

Config file content : 
{'api_endpoint': 'https://api.codecarbon.io'}
No organization_id in config, follow setup instruction to complete your 
configuration file!
? Use existing global ~/.codecarbon.config to configure or create a new file some
where else ? Create New Config
Creating new config file
Where do you want to put your config file ? [~/.codecarbon.config]: 
Config file created at /home/thoresen/.codecarbon.config
Current API endpoint is https://api.codecarbon.io. Press enter to continue or input other url [https://api.codecarbon.io]: 
? Pick existing organization from list or Create new organization ? Create New Or
ganization
Organization name [Code Carbon user test]: 
Organization description [Code Carbon user test]: 
[codecarbon ERROR @ 18:45:30] ApiClient Error when calling the API on https://api.codecarbon.io/organizations with : {"name": "Code Carbon user test", "description": "Code Carbon user test"}
[codecarbon ERROR @ 18:45:30] ApiClient API return http code 403 and answer : {"detail":"User not found"}
Created organization : None
╭────────────────────── Traceback (most recent call last) ──────────────────────╮
│ /home/thoresen/Code/green_ai_bench/.venv/lib/python3.12/site-packages/codecar │
│ bon/cli/main.py:217 in config                                                 │
│                                                                               │
│   214 │   │   print(f"Created organization : {organization}")                 │
│   215 │   else:                                                               │
│   216 │   │   organization = [orga for orga in organizations if orga["name"]  │
│ ❱ 217 │   org_id = organization["id"]                                         │
│   218 │   overwrite_local_config("organization_id", org_id, path=file_path)   │
│   219 │                                                                       │
│   220 │   projects = api.list_projects_from_organization(org_id)              │
│                                                                               │
│ ╭───────────────────────────────── locals ──────────────────────────────────╮ │
│ │                 api = <codecarbon.core.api_client.ApiClient object at     │ │
│ │                       0x7b18290c0440>                                     │ │
│ │        api_endpoint = 'https://api.codecarbon.io'                         │ │
[...]

PR #763, opened 2025-01-19
fix: redirect user to dashboard after cli account creation

SaboniAmine:
To solve #757, which is caused by the fact that the applicative account is only created when visiting the dashboard for the first time, with a valid jwt, lets redirect the user to the home page of the dashboard after CLI account creation.
The config CLI function should redirect to the dashboard / create the applicative user account when used with no prior visit to the dashboard, but this can be done in a separate PR

### DIFF
--- a/carbonserver/carbonserver/api/routers/projects.py
+++ b/carbonserver/carbonserver/api/routers/projects.py
@@ -92,6 +92,7 @@ def read_project_detailed_sums(
     project_id: str,
     start_date: Optional[datetime] = None,
     end_date: Optional[datetime] = None,
+    auth_user: UserWithAuthDependency = Depends(MandatoryUserWithAuthDependency),
     project_global_sum_usecase: ProjectSumsUsecase = Depends(
         Provide[ServerContainer.project_sums_usecase]
     ),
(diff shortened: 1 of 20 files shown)


inimaz (approved):
LGTM. Thanks!

carbonserver/carbonserver/api/routers/projects.py
```diff
@@ -91,6 +91,7 @@ def read_project_detailed_sums(
     project_id: str,
     start_date: Optional[datetime] = None,
     end_date: Optional[datetime] = None,
+    auth_user: UserWithAuthDependency = Depends(MandatoryUserWithAuthDependency),
```
prmths128, line 94: => Optionnal, check access in the service

prmths128 (approved):

SaboniAmine:
OK so the redirect response from starlette doesn't work with the fief auth success page. Lets force directly from the config a call to an endpoint which will trigger the user applicative account creation

carbonserver/carbonserver/api/routers/projects.py
SaboniAmine: lets rebase the PR when finished

PR #769, opened 2025-01-22
Improve arguments consistency in track_emissions decorator

LucasHaug:
This PR was motivated by the fact that, at the moment, it is not possible to set the `tracking_mode` for the emissions tracker when using the `track_emissions` decorator.

When I was fixing it, I realized that other arguments were also missing, such as `experiment_name`, `on_csv_write`, `logger_preamble`, and `country_2letter_iso_code`.

To make it easier to check for missing arguments in the future, I changed the order of the arguments in the decorator to match the order in the `BaseEmissionsTracker` and `OfflineEmissionsTracker`.


benoit-cty:
Nice catch ! Thanks a lot for fixing this.

benoit-cty (approved):

PR #771, opened 2025-01-24
fix/add_expid_offline_mode

### LINKED ISSUE #770: Argument 'experiment_id' not taken into account when 'offline' is set to False in track_emissions decorator.
### Problem 

When running:

```py
@track_emissions(offline=True, experiment_id=experiment_id, country_iso_code=os.getenv("CODECARBON_COUNTRY_ISO_CODE"))
def myfunc():
    pass
```

then the experiment id is **not taken into account** in the results. Though it is when `offline=False`.

This is du to the `OfflineEmissionTracker` not taking experiment_id as an input parameter, though the BaseEmissionTracker abstract class takes it.

### Expected behavior

Argument `offline` would not influence te behavior of `experiment_id` and the later would be taken into account in both cases

MathieuCiancone:
closes #770

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -60,8 +60,8 @@ class BaseEmissionsTracker(ABC):
     """
     Primary abstraction with Emissions Tracking functionality.
     Has two abstract methods, `_get_geo_metadata` and `_get_cloud_metadata`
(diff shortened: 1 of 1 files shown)


benoit-cty:
Thanks, but as said in the issue we suggest to use `project_name` and keep `experiment_id` for the API.

PR #772, opened 2025-01-26
ci: deploy app from ci

inimaz:
Deploy to PRD when push to master

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,6 +14,9 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v4
+      with:
+        fetch-depth: 0 # Same as git fetch --unshallow
+        clean: true    # Remove any untracked files or modifications
 
     - name: Install Clever Tools CLI
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,41 @@
...
+      env:
+        CLEVER_TOKEN: ${{ secrets.CLEVER_TOKEN }}
+      run: clever login --token ${{ secrets.CLEVER_TOKEN }} --secret ${{ secrets.CLEVER_SECRET }}
+
+    - name: Deploy backend to Clever Cloud (Staging)
+      env:
+        CLEVER_APP_ID: ${{ secrets.BE_CLEVER_APP_ID_STAGING }}
+      run: |
+        clever link $CLEVER_APP_ID
+        clever deploy
```
SaboniAmine, line 34: Lets add the "-f" tag, otherwise the history will be hard to maintain (this will prevent us from manually deploy the app to prod)

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,41 @@
...
+        PATH=${PATH}:$(pwd)/clever-tools-${CC_VERSION}_linux
+
+    - name: Login to Clever Cloud
+      env:
+        CLEVER_TOKEN: ${{ secrets.CLEVER_TOKEN }}
+      run: clever login --token ${{ secrets.CLEVER_TOKEN }} --secret ${{ secrets.CLEVER_SECRET }}
+
+    - name: Deploy backend to Clever Cloud (Staging)
+      env:
+        CLEVER_APP_ID: ${{ secrets.BE_CLEVER_APP_ID_STAGING }}
```
SaboniAmine, line 31: Can we target only the prod env with the CI ? And keep manual deploy for the dev env ?

SaboniAmine (commented):
Thanks for this CI !!

.github/workflows/deploy.yml
inimaz: Sure! I wanted to test the deploy first towards staging in order not to crash anything

.github/workflows/deploy.yml
inimaz: Done!

.github/workflows/deploy.yml
inimaz: The deployments to staging worked!

SaboniAmine (approved):

PR #777, opened 2025-01-30
Add an Ollama and a command line tool example

benoit-cty:
Demonstrate how to use CodeCarbon to measure the emissions of:
- a Ollama API running locally.
- a run of a command line tool


inimaz (approved):
Thanks for this @benoit-cty !

PR #782, opened 2025-02-10
CLI can be called in offline mode without and API endpoint

EvanKomp:
I was having trouble getting `codecarbon monitor` to work in an offline environment. It seems like OfflineEmissionsTracker is only accessible in the python api. I changed the main CLI script to allow the user to start an offline mode.

Side note, I also could not use `codecarbon config` in this environment - the second prompt, for the api endpoint gave me `FiefAuthNotAuthenticatedError`

The tool could be pretty useful for researchers using HPC clusters and submitting jobs for training/and or other non deep learning hgh compute jobs like simulations, if a clear offline CLI was available

### DIFF
--- a/codecarbon/cli/main.py
+++ b/codecarbon/cli/main.py
@@ -1,9 +1,9 @@
 import os
+import signal
 import sys
 import time
 from pathlib import Path
 from typing import Optional
-import signal
 
 import questionary
 import requests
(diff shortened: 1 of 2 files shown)


inimaz:
Hello @EvanKomp, thanks for using `codecarbon`!

there is the option of running `codecarbon monitor --no-api` that launches codecarbon in offline mode. Have you tried it/Does that not work for you?

EvanKomp:
It probably could with a few changes - when using the no-api flag it still uses the EmissionTracker class, which defaults to global emissions when it cannot figure out the geo metadata. A mechanism to explicitly pass a country/region would be helpful, like in the pull. 

Even more idea would be a way to configure for a particular energy mix. That may be too niche a problem. I am on a campus which is mostly solar and differs from my local grid, so being able to modify the config file to overwrite this somehow would be super.

benoit-cty (approved):

Issue #768, opened 2025-01-22
Saving file not where it is supposed to be


Lux-aura:
Hello ! :) 

Everything works well but I can't find the .csv file where it is supposed to be. 

![Image](https://github.com/user-attachments/assets/20354f1a-b478-4a1a-8ec0-85a681c7bb3a)

inimaz:
Hello @Lux-aura, 

thanks for using `codecarbon`!
In order to help further, can you tell us more how did you launch it? Is it by calling `EmissionsTracker` or via the cli `codecarbon monitor...`. 
As well provide an example of the params you used to initialize it. Thanks!

Lux-aura:
It is by the cli codecarbon monitor 
Thanks ! :)

inimaz:
Hard to tell! Can you provide the logs after you run `codecarbon monitor`? + the content of the .codecarbon.config file you might have used.

Lux-aura:
Please find the log here but can't drop a .config file. 

![Image](https://github.com/user-attachments/assets/a22ab70a-a977-49d0-a77b-ed87eab357c9)

![Image](https://github.com/user-attachments/assets/3709d838-3bdc-425e-ba91-8c952426e425)

![Image](https://github.com/user-attachments/assets/c0501eee-6dd4-4a23-b483-10c2e65b1b96)

I don't know how to send you ?

Ucodia:
Same exact issue using codecarbon 2.8.3 with the decorator `@track_emissions`

```
[codecarbon INFO @ 11:44:47] Saving emissions data to file C:\Users\ucodia\code\stylegan3\emissions.csv
```

No file to be found in this directory.

See commit on stylegan3 fork: https://github.com/Ucodia/stylegan3/commit/18e4ccc4657688ffd67d93ef9194ccc2a060d20f

felixkosmalla:
For me it only works with 2.7.1

inimaz:
Thanks for reporting it. I see the issue. 
The current log says `Saving ...` whereas the saving is done somewhere else down the line.  So if there is an error in between, nothing was saved even though the log was already printed.
Now changing the log to `Emissions data (if any) will be saved to file ...`, see -->https://github.com/mlco2/codecarbon/pull/794/files

PR #785, opened 2025-02-13
Fix SIGTERM handling to allow cleanup code execution

Ucodia:
The current implementation uses `os._exit(1)` in the Lock class's signal handler, which immediately terminates the process without allowing any cleanup code to execute. This prevents important cleanup code in `finally` blocks from being executed.

This PR replaces `os._exit(1)` with `raise SystemExit(1)`, which allows Python to handle the termination gracefully. This ensures that:
- `try/finally` blocks complete their execution
- Context managers can clean up resources
- `atexit` handlers are called
- Other cleanup code has a chance to run

This is particularly important for users who rely on cleanup code in decorators like `@track_emissions` to properly close resources and save data when their application receives a SIGTERM signal.

Fixes issue #768

### DIFF
--- a/codecarbon/lock.py
+++ b/codecarbon/lock.py
@@ -38,7 +38,7 @@ class Lock:
         """Ensures the lock file is removed when the script is interrupted."""
         logger.debug(f"Signal {signum} received. Releasing lock and exiting.")
         self.release()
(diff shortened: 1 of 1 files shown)


inimaz:
Indeed, thanks for this @Ucodia !

inimaz (approved):

Issue #790, opened 2025-02-26
CPU on Raspberry Pi 5 not recognized


FrejaThoresen:
Dear CodeCarbon team,

I am running codecarbon on a Raspberry Pi 5, and the CPU (and Hailo AI acceleration device) are not recognized. 

`[codecarbon INFO @ 16:09:26] Codecarbon is taking the configuration from global file: /home/pi/.codecarbon.config
[codecarbon INFO @ 16:09:26] [setup] RAM Tracking...
[codecarbon INFO @ 16:09:26] [setup] CPU Tracking...
[codecarbon WARNING @ 16:09:26] No CPU tracking mode found. Falling back on CPU constant mode. 
 Linux OS detected: Please ensure RAPL files exist at \sys\class\powercap\intel-rapl to measure CPU

[codecarbon WARNING @ 16:09:26] We saw that you have a Cortex-A76 but we don't know it. Please contact us.
[codecarbon INFO @ 16:09:26] CPU Model on constant consumption mode: Cortex-A76
[codecarbon INFO @ 16:09:26] [setup] GPU Tracking...
[codecarbon INFO @ 16:09:26] No GPU found.
[codecarbon INFO @ 16:09:26] >>> Tracker's metadata:
[codecarbon INFO @ 16:09:26]   Platform system: Linux-6.6.62+rpt-rpi-2712-aarch64-with-glibc2.36
[codecarbon INFO @ 16:09:26]   Python version: 3.11.2
[codecarbon INFO @ 16:09:26]   CodeCarbon version: 2.8.3
[codecarbon INFO @ 16:09:26]   Available RAM : 7.860 GB
[codecarbon INFO @ 16:09:26]   CPU count: 4
[codecarbon INFO @ 16:09:26]   CPU model: Cortex-A76
[codecarbon INFO @ 16:09:26]   GPU count: None
[codecarbon INFO @ 16:09:26]   GPU model: None
[codecarbon INFO @ 16:09:29] Saving emissions data to file emissions.csv
[codecarbon INFO @ 16:09:29] 
Graceful stopping: collecting and writing information.
Please wait a few seconds...
[codecarbon INFO @ 16:09:29] Energy consumed for RAM : 0.000000 kWh. RAM Power : 2.9473571777343754 W
[codecarbon INFO @ 16:09:29] Energy consumed for all CPUs : 0.000000 kWh. Total CPU Power : 42.5 W
[...]

PR #791, opened 2025-02-28
[Version 3] Raspberry CPU support

benoit-cty:
Following https://github.com/mlco2/codecarbon/issues/790 I've added the support for Rpi 4 and 5 CPU.

Source : https://gadgetversus.com/processor/broadcom-bcm2712-vs-broadcom-bcm2711/


codecarbon/data/hardware/cpu_power.csv
```diff
@@ -1163,6 +1163,8 @@ B75,95
 B77,95
 B97,95
 B99,95
+Cortex-A72,7.5
```
inimaz, line 1166: I see in the reference that this one is 8W
image

inimaz:
@benoit-cty you have some conflicts there, can you check? It would be interesting to merge this thanks!

Issue #778, opened 2025-01-31
Error when using codecarbon config


robitaillej:
I'm trying to link my project to codecarbon ui using `codecarbon login` and `codecarbon config`. I am able to connect correctly to my online account, which properly generates the credentials.json file. In this account, I manually added from the ui the Nautilo projec. But I have errors with code 500 when running `codecarbon config`  and the result is that the experiment is not set in the ui. It doesnt seem like its able to properly talk with the API endpoint. Here is the full log of the command prompt:
```
$ codecarbon config
Welcome to CodeCarbon configuration wizard
Existing global config file found :
Current configuration : 

Config file content : 
{
    'api_endpoint': 'https://api.codecarbon.io',
    'organization_id': '4ef0bf6c-481a-4cc8-b295-57b3d1b3a8c0',
    'project_id': '101ee3ca-7a1c-4542-9425-6b3a26110d15',
    'experiment_id': '57d3eadc-84e5-4094-aec2-64d0a00c1512',
    'api_key': 'key'
}
[codecarbon ERROR @ 15:27:59] ApiClient Error when calling the API on https://api.codecarbon.io/projects/101ee3ca-7a1c-4542-9425-6b3a26110d15
[codecarbon ERROR @ 15:27:59] ApiClient API return http code 500 and answer : {"detail":"Generic error"}
[codecarbon ERROR @ 15:27:59] ApiClient Error when calling the API on https://api.codecarbon.io/experiments/57d3eadc-84e5-4094-aec2-64d0a00c1512
[codecarbon ERROR @ 15:27:59] ApiClient API return http code 500 and answer : {"detail":"Generic error"}

Experiment :
None

Project :
None

Organization :
{'name': 'p4zHfGogV2PS6BUIyVGe', 'description': 'Default organization', 'id': '4ef0bf6c-481a-4cc8-b295-57b3d1b3a8c0'}
? Use existing global ~/.codecarbon.config to configure or create a new file somewhere else ? Create New Config
Creating new config file
Where do you want to put your config file ? [~/.codecarbon.config]: ./codecarbon.config
Config file created at codecarbon.config
Current API endpoint is https://api.codecarbon.io. Press enter to continue or input other url [https://api.codecarbon.io]: 
? Pick existing organization from list or Create new organization ? p4zHfGogV2PS6BUIyVGe
? Pick existing project from list or Create new project ? Nautilo
[...]

PR #795, opened 2025-03-08
fix(api): auth context can read project

inimaz:
This is related to https://github.com/mlco2/codecarbon/issues/778.
Get one project did not work because the call to know if the user had enough permissions did not work.


SaboniAmine (approved):
Great, thanks !

prmths128 (approved):

PR #797, opened 2025-03-09
feat(codecarbon) add raspberry support

prmths128:
Add support for rpi power metrics (from rpi tools, using `vcgencmd pmic_read_adc`)


inimaz:
🙌  This would be very nice to have @prmths128, does it work?

inimaz:
Hello @prmths128. Can this be merged? See also #791 if there is any overlap. @benoit-cty can you check?

PR #804, opened 2025-03-15
[Version 3] New heuristic for RAM

benoit-cty:
Breaking change in RAM power estimation.

    In V3, we need to improve the accuracy of the RAM power estimation.
    Because the power consumption of RAM is not linear with the amount of memory used,
    for example, in servers you could have thousands of GB of RAM but the power
    consumption would not be proportional to the amount of memory used, but to the number
    of memory modules used.
    But there is no way to know the memory modules used in the system, without admin rights.
    So we need to build a heuristic that is more accurate than the previous one.
    For example keep a minimum of 2 modules. Execept for ARM CPU like rapsberry pi where we will consider a 3W constant.
    Then consider the max RAM per module is 128GB and that RAM module only exist in power of 2 (2, 4, 8, 16, 32, 64, 128).
    So we can estimate the power consumption of the RAM by the number of modules used.

    1. **ARM CPU Detection**:
    - Added a `_detect_arm_cpu` method that checks if the system is using an ARM architecture
    - For ARM CPUs (like Raspberry Pi), a constant 3W will be used as the minimum power

    2. **DIMM Count Estimation**:
    - Created a `_estimate_dimm_count` method that intelligently estimates how many memory modules might be present based on total RAM size
    - Takes into account that servers typically have more and larger DIMMs
    - Assumes DIMM sizes follow powers of 2 (4GB, 8GB, 16GB, 32GB, 64GB, 128GB) as specified

    3. **Scaling Power Model**:
    - Base power per DIMM is 2.5W for x86 systems and 1.5W for ARM systems
    - For standard systems (up to 4 DIMMs): linear scaling at full power per DIMM
    - For medium systems (5-8 DIMMs): decreasing efficiency (90% power per additional DIMM)
    - For large systems (9-16 DIMMs): further reduced efficiency (80% power per additional DIMM)
    - For very large systems (17+ DIMMs): highest efficiency (70% power per additional DIMM)

    4. **Minimum Power Guarantees**:
    - Ensures at least 5W for x86 systems (assuming 2 DIMMs at minimum)
    - Ensures at least 3W for ARM systems as requested

    ### Example Power Estimates:

    - **Small laptop (8GB RAM)**: ~5W (2 DIMMs at 2.5W each)
    - **Desktop (32GB RAM)**: ~10W (4 DIMMs at 2.5W each)
    - **Small server (128GB RAM)**: ~18.6W (8 DIMMs with efficiency scaling)
    - **Large server (1TB RAM)**: ~44W (using 16x64GB DIMMs with high efficiency scaling)

    This approach significantly improves the accuracy for large servers by recognizing that RAM power consumption doesn't scale linearly with capacity, but rather with the number of physical modules. Since we don't have direct access to the actual DIMM configuration, this heuristic provides a more reasonable estimate than the previous linear model.

    The model also includes detailed debug logging that will show the estimated power for given memory sizes, helping with validation and fine-tuning in the future.


benoit-cty:
For the record, here is the proposition of DeepSeek R1 :
To improve the V3 heuristic for RAM power estimation while accounting for older systems with smaller modules, the following approach can be used:
Heuristic Adjustments:

    ARM Exception: For ARM-based systems (e.g., Raspberry Pi), use a fixed 3W.

    Minimum Modules: Assume a minimum of 2 modules for non-ARM systems to avoid underestimation in small configurations.

    Dynamic Max Module Size:

        Use a lower assumed maximum module size (e.g., 32GB) instead of 128GB. This increases the estimated number of modules for older systems, improving accuracy without admin data.

        Calculate modules as:
        Number of Modules = max(ceil(Total RAM / 32GB), 2)
        This balances newer systems (which may use larger modules) and older ones (which use smaller ones).

    Per-Module Power: Assign 3W per module, reflecting typical power usage regardless of module size.

Example Calculations:

    16GB System:
    16GB / 32GB = 0.5 → ceil(0.5) = 1 → 2 modules (min enforced) → 6W.

    64GB System:
    64GB / 32GB = 2 → 2 modules → 6W (aligns with newer systems using 32GB modules; older systems with 8GB modules would have 8 modules → 24W, but heuristic estimates conservatively).

    128GB System:
    128GB / 32GB = 4 → 4 modules → 12W.

Trade-offs:

    Older Systems: Closer estimates for configurations using smaller modules (e.g., 8GB/module) by assuming 32GB/module.

    Newer Systems: Slight overestimation for large modules (e.g., 64GB/module), but more accurate than the linear model.

This approach prioritizes accuracy for older systems while maintaining simplicity and avoiding reliance on admin-only data.

docs/examples.html
```diff
@@ -157,40 +160,50 @@ <h2>Using the Context Manager<a class="headerlink" href="#using-the-context-mana
...
 </pre></div>
 </div>
 </section>
-<section id="using-the-decorator">
-<h2>Using the Decorator<a class="headerlink" href="#using-the-decorator" title="Link to this heading"></a></h2>
+<section id="using-the-explicit-object">
+<h2>Using the Explicit Object<a class="headerlink" href="#using-the-explicit-object" title="Link to this heading"></a></h2>
+<p>This is the recommended way to use the CodeCarbon tracker in a Notebook : you instantiate the tracker and call the <cite>start()</cite> method at the begginning of the Notebook. You call the stop() method at the end of the Notebook to stop the tracker and get the emissions.</p>
+<p>If not in an interractive Notebook, always use a <cite>try…finally</cite> block to ensure that the tracker is stopped even if an error occurs during training.
+This is important to ensure the CodeCarbon scheduler is stopped. If you don’t use <cite>try…finally</cite>, the scheduler will continue running in the background after your computation code has crashed, so your program will never finish.</p>
```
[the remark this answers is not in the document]
benoit-cty: done

codecarbon/core/cpu.py
```diff
@@ -66,14 +66,17 @@ def is_rapl_available() -> bool:
 def is_psutil_available():
     try:
         nice = psutil.cpu_times().nice
-        if nice > 0.1:
+        if nice > 0.0001:
```
benoit-cty, line 69: Because on huge CPU, it could be above 0 but under 0.1.

codecarbon/core/units.py
```diff
@@ -61,7 +63,10 @@ class Energy:
 
     @classmethod
     def from_power_and_time(cls, *, power: "Power", time: "Time") -> "Energy":
-        return cls(kWh=power.kW * time.hours)
+        assert isinstance(power.kW, float)
```
SaboniAmine, line 66: Quel est l'intérêt de l'assert ici ? ça ne risque pas de casser le programme à un endroit non désiré ?

codecarbon/external/ram.py
```diff
@@ -0,0 +1,343 @@
...
+            base_power_per_dimm = 1.5  # Watts
+            # Minimum 3W for ARM as requested
+            min_power = 3.0
+        else:
+            # x86 systems
+            base_power_per_dimm = RAM_SLOT_POWER_X86  # Watts
+            # Minimum 5W for x86 as requested (2 sticks at 2.5W)
+            min_power = base_power_per_dimm * 2
+
+        # Estimate power based on DIMM count with decreasing marginal power per DIMM as count increases
```
SaboniAmine, line 167: ça pourrait être intéressant de faire des statistiques sur les setups les plus courants. Peut être via les données remontées par l'API

SaboniAmine (approved):
Merci Benoît !

codecarbon/core/units.py
benoit-cty: Cette endroit est la base du calcul, si jamais il n'y a pas la bonne unité il y a plein d'effets de bord.
Ca m'est arrivé lors du dev et j'ai mis du temps à trouver la source du problème. Je me suis dit qu'en le vérifiant ici, cela évitera des problèmes à l'avenir.

PR #809, opened 2025-03-25
fix: update nextjs to latest to prevent known vulnerability

inimaz:
* Vulnerability: https://zhero-web-sec.github.io/research-and-things/nextjs-and-the-corrupt-middleware

We are not impacted by it, but it is always good to be on the safe side.


SaboniAmine (approved):
merci Iñigo!

PR #812, opened 2025-03-25
chores: split github workflows

inimaz:
# Goal

Split the workflows so that only some jobs are triggered depending on the path of the files updated.

e.g.  an edit on the server will not trigger the jobs of the UI, an edit on the package will not trigger a deploy of the server...


benoit-cty (approved):
Really nice !!!

I didn't know it was possible.

PR #814, opened 2025-03-28
Prevent to always create a run

benoit-cty:
The `ApiClient` class is creating a run in init. But when we are using it only to get data it create useless entry in the database.

And now that the new API is in production, we could remove the warning.


inimaz (changes requested):
Thanks @benoit-cty! Maybe we can remove the call to `self._create_run()` in the `__init__`? I see that we create the run anyhow if not present in the `add_emissions`  See [here](https://github.com/mlco2/codecarbon/pull/814/files#diff-cebb815d4048a12df07aea4bc8dd17ac158de6b5d2c8163e2ccc220dd48d954eL212)/

codecarbon/core/api_client.py
```diff
@@ -39,33 +39,29 @@ class ApiClient:  # (AsyncClient)
...
     def __init__(
         self,
-        # endpoint_url="https://api.codecarbon.io",
-        endpoint_url="https://dash-dev.cleverapps.io/api",  # beta API
+        endpoint_url="https://api.codecarbon.io",
         experiment_id=None,
         api_key=None,
         access_token=None,
         conf=None,
+        do_not_create_run=False,
```
inimaz, line 47: Maybe we can rename the variable `do_not_create_run` to `create_run`? Because the (my) brain is too bad with double-negations. Otherwise we can merge it

inimaz (approved):
Thanks @benoit-cty ! With this, the API client will be in better shape. I mentioned a minor comment

SaboniAmine (approved):
Thanks!

codecarbon/core/api_client.py
benoit-cty: True. I switch to `create_run_automatically=True`

PR #825, opened 2025-04-20
fix(carbonserver) validate fief token

prmths128:
Add missing token validation for cli use case

### DIFF
--- a/carbonserver/carbonserver/api/services/auth_service.py
+++ b/carbonserver/carbonserver/api/services/auth_service.py
@@ -66,7 +66,7 @@ class UserWithAuthDependency:
         """
         self.error_if_not_found = error_if_not_found
 
-    def __call__(
+    async def __call__(
         self,
         auth_user_cookie: Optional[FiefUserInfo] = Depends(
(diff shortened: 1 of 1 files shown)


carbonserver/carbonserver/api/services/auth_service.py
```diff
@@ -85,6 +85,9 @@ def __call__(
                 algorithms=["HS256", "RS256"],
             )
         elif bearer_token is not None:
+            if settings.environment != "develop":
+                await fief.validate_access_token(bearer_token.credentials)
```
inimaz, line 89: This throws an exception if not valid, see in its docs --> https://github.com/fief-dev/fief-python/blob/main/fief_client/client.py#L655-L662

Maybe we can catch it and throw a 401?

SaboniAmine (commented):
See Inigo's review

carbonserver/carbonserver/api/services/auth_service.py
prmths128: yeah, good idea

inimaz (approved):
Looks good!

Issue #821, opened 2025-04-15
Request to add i9-14900HX to supported CPU list


RafiullahOmar:
Hey, could you please add the Intel Core i9-14900HX to the list of CPUs supported by CodeCarbon? Its TDP is 55W. Thanks!

PR #827, opened 2025-04-24
Add i9-14900HX to supported CPU list #821

IamLRBA:
For more accurate energy and emissions estimation for users with this CPU, avoiding fallback values and improving the precision of CodeCarbon's tracking, I added support for the i9-14900HX to supported CPU list.

Fixes https://github.com/mlco2/codecarbon/issues/821


codecarbon/data/hardware/cpu_power.csv
```diff
@@ -3007,6 +3007,7 @@ Intel Pentium T4400,35
 Intel Pentium T4500,35
 Intel Pentium U5400,18
 Intel Pentium U5600,18
+Intel(R) Core(TM) i9-14900HX,55
```
benoit-cty, line 3010:
```suggestion
Intel Core i9-14900HX,55
```

benoit-cty (changes requested):
Sorry for the delay, may I suggest you include the change in the part of the file where other i9 are ?
To keep alphabetical order.

PR #834, opened 2025-05-05
Add missing empty line before code block

blegat:
It currently does not display as a code block : https://mlco2.github.io/codecarbon/methodology.html#ram


inimaz (approved):
Nice! Thanks @blegat !

Issue #845, opened 2025-05-23
[Bug] Carbonboard doesn't support new dash app version


nona-e:
``Dash App`` has a new version v3.0.4 which changes `app.run_server` to `app.run` in carbonboard.py:

```
def viz(filepath: str, port: int = 8050, debug: bool = False) -> None:
    df = pd.read_csv(filepath)
    app = render_app(df)
    app.run_server(port=port, debug=debug)
```

To fix this, you could change it to use `app.run()` or install an old version of dash `conda install -c conda-forge dash=2.18.2`.

PR #847, opened 2025-05-23
Fix/845: Change to use app.run()

nona-e:
The Dash App v3.0+ switched from using app.run_server() to app.run().

Fixes mlco2/codecarbon#845


benoit-cty:
Thank you for contributing.

If you want to fix the pre-commit error, you could look at https://github.com/mlco2/codecarbon/blob/master/CONTRIBUTING.md#coding-style--linting to install it locally.

benoit-cty (approved):

benoit-cty:
Thanks for the contribution !

PR #850, opened 2025-05-27
fix(cli): auth error when doing codecarbon login

inimaz:
There was an issue after the login via `codecarbon` cli, the wrong token was being checked.


inimaz:
This is the output of the response now:
```bash
codecarbon login
⠇ Please complete authentication in your browser.
Successfully authenticated Getting a token...
```

PR #852, opened 2025-06-02
851 emissions energy consumed is zero when tracking tasks

benoit-cty:
Measure always works but sometime the returned energy is 0.


benoit-cty:
Duplicate with #853

Issue #874, opened 2025-06-19
[Dashboard] Add more info to the login page


benoit-cty:
The login page of https://dashboard.codecarbon.io/ is missing some more information on what the user are going to find here.

We need to explain that this is only for the users of the API and link to https://codecarbon.io/ and https://github.com/mlco2/codecarbon

paulCoulombel:
Hi! I'm a newbie in the open-source world (no PR done yet), but I really like your project and want to contribute to it. Is this issue still relevant? If it is, I can try to develop something.

PR #875, opened 2025-06-20
feat(webapp): add more info to main login page

inimaz:
Related to #874 
Add more info to the login page:
image


benoit-cty (approved):
I think we could do more, like adding texts and screenshots of the dashboard, but it is a well improvement while waiting to do better later :+1:

PR #891, opened 2025-07-10
chores: create ossf scorecard workflow

inimaz:
## Description

Goal is to obtain a score from the static analysis of OSSF. More info can be found here ==> https://github.com/marketplace/actions/ossf-scorecard-action
This can help us adopt best practices.

Once merged, the following badge will show our score, and we can use it in our README somewhere

[OpenSSF Scorecard](https://scorecard.dev/viewer/?uri=github.com/mlco2/codecarbon)


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,78 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
```
SaboniAmine, line 9: What is the purpose of this rule ? Shouldn't it reference at least one value ?

SaboniAmine (approved):
Thanks Inigo!

.github/workflows/scorecard.yml
inimaz: Empty seems to be the value by default. We can update it later if needed

Issue #870, opened 2025-06-17
Fix the security issues detected by Code Scanning


benoit-cty:
See https://github.com/mlco2/codecarbon/security/code-scanning

PR #908, opened 2025-07-15
chores: tensorflow dependencies in examples

inimaz:
## Description
There is a dependency to `tensorflow` in the examples. Fixing it to a version to avoid flagging it as a vulnerability


inimaz:
Related to #870

PR #910, opened 2025-07-15
Potential fix for code scanning alert no. 12: Clear-text storage of sensitive information

inimaz:
Potential fix for [https://github.com/mlco2/codecarbon/security/code-scanning/12](https://github.com/mlco2/codecarbon/security/code-scanning/12)

To address the issue, sensitive data such as `settings.fief_admin_password` should be encrypted before being written to files. The `cryptography` library can be used to encrypt the sensitive data. A symmetric encryption key can be generated and stored securely, and the sensitive data can be encrypted using this key before being written to the `.env` files. The decryption process should only occur when the data is needed.

Steps to fix:
1. Import the `cryptography` library and set up a symmetric encryption mechanism.
2. Generate a secure encryption key and store it securely (e.g., in a secure key management system or environment variable).
3. Encrypt sensitive data (e.g., `settings.fief_admin_password`) before writing it to the `.env` files.
4. Update the `replace` function to handle encrypted values appropriately.

---

### DIFF
--- a/deploy/deploy.py
+++ b/deploy/deploy.py
@@ -2,12 +2,12 @@
 import dataclasses
 import re
 import secrets
-from cryptography.fernet import Fernet
 import subprocess
 import time
 
 import requests
 import typer
+from cryptography.fernet import Fernet
 from pydantic_settings import BaseSettings, SettingsConfigDict
 from typing_extensions import Annotated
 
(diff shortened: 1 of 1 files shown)


prmths128:
Not sure that's the way to go. Seems like a way to complicate the setup without adding real security.
If you use a cloud deployment you should use a secret manager to manage them instead of writing to .env files.
If you use a local setup with .env file you *will* have sensitive information there, there's no way around that.

inimaz:
I agree, I think it gets flagged as vulnerability just because later in the code one can print it for instance. But closing this as not needed for now, thanks!

PR #939, opened 2025-09-24
fix: pip install require-hashes in dockerfile

inimaz:
Fix https://github.com/mlco2/codecarbon/security/code-scanning/29 vulnerability


SaboniAmine (approved):
Thanks!

cianc:
FYI, I think this broke the Docker build, see https://github.com/mlco2/codecarbon/pull/962

PR #962, opened 2025-11-07
Added missing PIP requirement.txt hashes

cianc:
## Description
Docker build currently fails due to missing hashes in requirements/requirements.txt 

## Related Issue
Please link to the issue this PR resolves: 961

## Motivation and Context

https://github.com/mlco2/codecarbon/pull/939 added the `--require-hashes` flag to `pip install` in the Dockerfile. This now fails with:

```
ERROR: Hashes are required in --require-hashes mode, but they are missing from some requirements. Here is a list of those requirements along with the hashes their downloaded archives actually had. Add lines like these to your requirements files to prevent tampering. (If you did not enable --require-hashes manually, note that it turns on automatically when any package has a hash.)
    alembic==1.17.1 --hash=sha256:cbc2386e60f89608bb63f30d2d6cc66c7aaed1fe105bd862828600e5ad167023
```

## How Has This Been Tested?
The following it succeeds in the branch with this PR, but fails with the above error in the master branch:

```
pip install -r requirements/requirements-api.txt --require-hashes
```

## Types of changes

## Checklist:


SaboniAmine:
Hello @cianc, thanks for your contribution!
I feel like we missed something on the #939 PR. The docker build isn't used for the API deployment on our infra, it is only used for local development setup. For this purpose, I don't think the hashes are needed, and for clarity I would like to avoid having hashes in the requirements file used for our own deployment. 
The solution might then be to remove the `--require-hash` argument from the Dockerfile rather than compiling dependencies with this option. Are you ok with this @inimaz ?

cianc:
Just checking in if we know what we want to do here yet? I'm happy to revert this if @inimaz wants to drop `--require-hash`

inimaz:
Yes, sorry. It was a security issue spotted by our security bot and I thought it was an easy fix, sorry for the mess. The best way will be to just ignore it since we do not use it for deployment other than in local. What do you think @cianc , @SaboniAmine ?

cianc:
Sounds good to me, thanks @inimaz

PR #971, opened 2025-11-16
Doc : Discord contact and Privacy Policy draft

benoit-cty:
## Description

Change ways to contact us.

Draft a Privacy Policy.

## Motivation and Context

Readme information was old.


benoit-cty:
> Thanks @benoit-cty! Much clearer Readme. We can maybe move the privacy_policy file into the `/webapp` folder or the `/docs`? Or you prefer to have it at the root?

I prefer to keep it here, and move it in other PR with better integration.

SaboniAmine:
Should we add the privacy policy as a new front end page ?
Do we need to add more details on the processing registry, in the dashboard ?

PR #977, opened 2025-11-19
fix: #976 Inconsistent behavior for RAM and CPU Monitoring

benoit-cty:
## Description

RAM and CPU process mode do not handle sub-process the same way.

## Related Issue
Please link to the issue this PR resolves: #976

## Motivation and Context

For RAM we look at all sub-process, but for CPU we just read first-child.

## How Has This Been Tested?

A new test has been added.

## Types of changes
What types of changes does your code introduce? Put an `x` in all the boxes that apply:
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
Go over all the following points, and put an `x` in all the boxes that apply.
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **[CONTRIBUTING.md](https://github.com/mlco2/codecarbon/blob/master/CONTRIBUTING.md)** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #975, opened 2025-11-19
Prometheus total Energy export


Matyro:
**Is your feature request related to a problem? Please describe.**
Prometheus exporter does only provide current power and energy consumption between each push. As far as I know the prometheus push gateway only caches a single values. Therefore if prometheus querys not often enough it is possible to miss values.

**Describe the solution you'd like**
Export the total energy since start of the application

cianc:
@inimaz if you'd like I can take this. I think it should be a fairly straightforward addition of a counter.

Matyro:
Not sure if this was the best solution but I implemented this in MR #984

williamacostalora:
Hi! I'm a student contributor in CodePath AI301 and I'd like to work on this issue. My plan is to add a cumulative energy counter to the Prometheus exporter so total energy since start is always available, regardless of scrape frequency. Happy to start reproducing this week — is there a preferred Prometheus client library version I should target?

PR #984, opened 2025-11-25
Tracking of multiple pids

Matyro:
## Description
Added the functionality to track several process ids and its children
Modified Prometheus output to report total energy and cleanup after termination

## Related Issue
-

## Motivation and Context
This allows to run codecarbon as a deamon logging slurm jobs directly from the slurm prolog and cleanup after jobs end

## How Has This Been Tested?
Changes are currently run in our cluster

## Types of changes

## Checklist:

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -176,7 +176,7 @@ class BaseEmissionsTracker(ABC):
             str
         ] = _sentinel,  # Deprecated, use electricitymaps_api_token
         tracking_mode: Optional[str] = _sentinel,
-        tracking_pids: Optional[int] = _sentinel,
+        tracking_pids: Optional[List[int]] = _sentinel,
         log_level: Optional[Union[int, str]] = _sentinel,
         on_csv_write: Optional[str] = _sentinel,
         logger_preamble: Optional[str] = _sentinel,
@@ -1182,7 +1182,7 @@ def track_emissions(
         str
     ] = _sentinel,  # Deprecated, use electricitymaps_api_token
     tracking_mode: Optional[str] = _sentinel,
-    tracking_pids: Optional[int] = _sentinel,
+    tracking_pids: Optional[List[int]] = _sentinel,
     log_level: Optional[Union[int, str]] = _sentinel,
     on_csv_write: Optional[str] = _sentinel,
     logger_preamble: Optional[str] = _sentinel,
--- a/tests/test_pid_tracking.py
+++ b/tests/test_pid_tracking.py
@@ -57,7 +57,7 @@ class TestPIDTracking(unittest.TestCase):
             output_dir=self.emissions_path,
             output_file=self.emissions_file + "_global.csv",
             tracking_mode="process",
-            tracking_pids=os.getpid(),
+            tracking_pids=[os.getpid()],
         )
 
         tracker_pid.start()
@@ -76,4 +76,10 @@ class TestPIDTracking(unittest.TestCase):
         assert isinstance(emissions_pid, float)
 
         self.assertNotEqual(emissions_pid, 0.0)
-        self.assertAlmostEqual(emissions_self, emissions_pid, 2)
+
+        # Compare emissions from both trackers, should be less than 10% difference
+        diff = abs(emissions_pid - emissions_self)
+        avg = (emissions_pid + emissions_self) / 2
+        percent_diff = (diff / avg) * 100
+        print(f"Percent difference: {percent_diff}%")
+        self.assertLessEqual(percent_diff, 10.0)
--- a/codecarbon/external/ram.py
+++ b/codecarbon/external/ram.py
@@ -50,7 +50,7 @@ class RAM(BaseHardware):
                                       total RAM used. Defaults to True.
             tracking_mode (str, optional): Whether to track "machine" or "process" RAM.
                                           Defaults to "machine".
-            tracking_pids (int, optional): Process id to track RAM usage for "process"
+            tracking_pids ([int], optional): Process id to track RAM usage for "process"
                                             tracking_mode. Defaults to None.
             force_ram_power (int, optional): User-provided RAM power in watts. If provided,
                                            this value is used instead of estimating RAM power.


benoit-cty:
Thanks, it seems better than https://github.com/mlco2/codecarbon/pull/977

Why did you add tracking_pids to the API ? You need to keep the tracked pids in the database ?

Matyro:
> Thanks, it seems better than #977
> 
> Why did you add tracking_pids to the API ? You need to keep the tracked pids in the database ?

Does not really make sense, you are right. 
I removed tracking_pids from the SQL and API

codecarbon/emissions_tracker.py
```diff
@@ -806,7 +810,7 @@ def _prepare_emissions_data(self) -> EmissionsData:
             longitude=self._conf.get("longitude"),
             latitude=self._conf.get("latitude"),
             ram_total_size=self._conf.get("ram_total_size"),
-            tracking_mode=self._conf.get("tracking_mode"),
+            tracking_mode=self._conf.get("tracking_mode"),
```
benoit-cty:
```suggestion
            tracking_mode=self._conf.get("tracking_mode"),
```

benoit-cty:
Thanks.
Do you think you could add unit test ?
And documentation in `methodology.rst` ?
Be aware that you changed `.rst.txt` but they will be overwritten, you have to change only `.rst` files for documentation.

benoit-cty:
I also encourage you to run `uv run task precommit-install` to heve linting on commit.
See https://github.com/mlco2/codecarbon/blob/master/CONTRIBUTING.md#coding-style--linting

Matyro:
I setup the precommit tasks and hopefully fixed every file I touched. 
Result of `uv run python -m unittest` not sure if they were caused by me and how


```
======================================================================
ERROR: tests.test_viz_data (unittest.loader._FailedTest.tests.test_viz_data)
----------------------------------------------------------------------
ImportError: Failed to import test module: tests.test_viz_data
  File "/usr/lib/python3.12/unittest/loader.py", line 394, in _find_test_path
    module = self._get_module_from_name(name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/unittest/loader.py", line 337, in _get_module_from_name
    __import__(name)
  File "/root/codecarbon/tests/test_viz_data.py", line 6, in <module>
    from dash import dash_table as dt
ModuleNotFoundError: No module named 'dash'


======================================================================
FAIL: test_empty_conf (tests.test_config.TestConfig.test_empty_conf)
----------------------------------------------------------------------
  File "/root/codecarbon/tests/test_config.py", line 151, in test_empty_conf
    self.assertDictEqual(conf, target)
AssertionError: {} != {'allow_multiple_runs': 'True'}
- {}
+ {'allow_multiple_runs': 'True'}

======================================================================
FAIL: test_parse_env_config (tests.test_config.TestConfig.test_parse_env_config)
----------------------------------------------------------------------
  File "/usr/lib/python3.12/unittest/mock.py", line 1843, in _inner
    return f(*args, **kw)
           ^^^^^^^^^^^^^^
  File "/root/codecarbon/tests/test_config.py", line 57, in test_parse_env_config
    self.assertDictEqual(
AssertionError: {'codecarbon': {'test': 'test-VALUE', 'test_key': 'this_other_value'}} != {'codecarbon': {'allow_multiple_runs': 'True', 'test': 'te[38 chars]ue'}}
- {'codecarbon': {'test': 'test-VALUE', 'test_key': 'this_other_value'}}
+ {'codecarbon': {'allow_multiple_runs': 'True',
+                 'test': 'test-VALUE',
+                 'test_key': 'this_other_value'}}

======================================================================
FAIL: test_read_confs (tests.test_config.TestConfig.test_read_confs)
----------------------------------------------------------------------
  File "/root/codecarbon/tests/test_config.py", line 96, in test_read_confs
    self.assertDictEqual(conf, target)
AssertionError: {'no_overwrite': 'path/to/somewhere', 'loca[103 chars]lue'} != {'allow_multiple_runs': 'True', 'no_overwri[134 chars]lue'}
+ {'allow_multiple_runs': 'True',
- {'local_new_key': 'cool value',
? ^

+  'local_new_key': 'cool value',
? ^

   'local_overwrite': 'SUCCESS:overwritten',
   'no_overwrite': 'path/to/somewhere',
   'syntax_test_key': 'no/space= problem2'}

----------------------------------------------------------------------
Ran 109 tests in 243.240s

FAILED (failures=3, errors=1, skipped=1)
```

Matyro:
Implemented #975 

grafik

codecarbon/output_methods/metrics/prometheus.py
```diff
@@ -70,19 +78,30 @@ def generate_gauge(metric_doc: MetricDocumentation):
 gpu_energy_gauge = generate_gauge(gpu_energy_doc)
 ram_energy_gauge = generate_gauge(ram_energy_doc)
 energy_consumed_gauge = generate_gauge(energy_consumed_doc)
+energy_consumed_total_gauge = generate_gauge(energy_consumed_total_doc)
```
cianc: Should this be a counter instead? Total energy consumed should never go down.

Matyro: Counter only has the "inc" routine instead of "set" , I would prefer to use the already accumulated data instead of maintaining a separate cumulative sum

cianc: But don't we already have the incremental value in `carbon_emission["energy_consumed"]`?

It would be nice to use the proper Prometheus semantics if possible to prevent confusion to future consumers of the data. For example, I would naivley assume I couldn't calculate a rate over this value because it's a gauge, whereas it would actually be fine.

codecarbon/output_methods/metrics/prometheus.py
```diff
@@ -124,7 +143,12 @@ def add_emission(self, carbon_emission: dict):
         # Send the new metric values
         push_to_gateway(
             self.prometheus_url,
-            job="codecarbon",
+            job=self.jobname,
             registry=registry,
             handler=self._auth_handler,
         )
+
+        # Now set the total energy consumed
```
cianc: Should this be moved above the push_to_gateway() call?

codecarbon/output_methods/metrics/prometheus.py
Matyro: "energy_consumed" is only the energy in the last time slice, for total energy the total_emission dict is need, which is present in the "out" function but not natively in the add_emission function. To avoid changing several interfaces I reworked some of the changes.

The problem originates that values in the push gateway are overwritten. If the codecarbon update time is shorter than the Prometheus query time you will lose values and not be able to correctly accumulate them in Prometheus.
If I did not misunderstood the pushgateway

cianc: I think I'm missing something.
If "energy_consumed" is the energy in the last time slice, then why can't total energy consumed be a counter and we call `energy_consumed_total_counter.inc(carbon_emission["energy_consumed"])` ?

Is there some reason that the sum of the "energy_consumed" won't be equal to the total energy consumed?

Matyro: I never looked this deep into the code. 
For this to work PrometheusOutput.out(...) needs to be called for every internal measurement update or at least with averages, It might be implemented this way ¯\_(ツ)_/¯

If it is implemented that way, then my next question is: Why is a total object passed, and is it supposed to be used?

If, for any reason, the Prometheus update rate is decoupled from the internal update rate, this could potentially lead to issues.

cianc: With the caveat that I'm not an expert in this code base and it would be good to get a maintainer to confirm:

Output handlers get called from here: https://github.com/mlco2/codecarbon/blob/36dcc227c6984fb5ecc8e8a05139ff238d421664/codecarbon/emissions_tracker.py#L701

and total and delta emissions are calculated here:
https://github.com/mlco2/codecarbon/blob/36dcc227c6984fb5ecc8e8a05139ff238d421664/codecarbon/emissions_tracker.py#L678

Note that the delta emissions calculation is based on the current and previously stored total emissions:
https://github.com/mlco2/codecarbon/blob/36dcc227c6984fb5ecc8e8a05139ff238d421664/codecarbon/emissions_tracker.py#L816

So using the delta emissions to increment the prometheus total emissions counter seems safe.
Alternatively you could store the total emissions on every call and calculate the delta yourself, effectively reproducing the logic in https://github.com/mlco2/codecarbon/blob/36dcc227c6984fb5ecc8e8a05139ff238d421664/codecarbon/emissions_tracker.py#L816

As to why are total emissions passed to `PrometheusOutput.out()`? Well the out() method of all output handlers need to match because they get called like this: https://github.com/mlco2/codecarbon/blob/36dcc227c6984fb5ecc8e8a05139ff238d421664/codecarbon/emissions_tracker.py#L702 - that means they all have to have total and delta params.

Of course that doesn't mean that the actual total needs to be passed, only that something is. You can see some other methods are more explicit about ignoring total, eg: https://github.com/mlco2/codecarbon/blob/36dcc227c6984fb5ecc8e8a05139ff238d421664/codecarbon/output_methods/http.py#L66

Matyro: I changed the type to counter in [856c5a5](https://github.com/mlco2/codecarbon/pull/984/commits/856c5a5fb6f3baa638755fe0873dda28e1f78749)

cianc: Nice, thank you!

Matyro:
Whats failing is the test I put in, how should this be fixed.
On my machine this did work and I do not know if I was lucky or something is wrong

The relevant part is found in tests/test_pid_tracking.py 

#

tests/test_pid_tracking.py
```diff
@@ -0,0 +1,79 @@
...
+        if os.path.isfile(self.emissions_file_path):
+            os.remove(self.emissions_file_path)
+
+        for proc in self.process:
+            proc.terminate()
+            proc.wait()
+
+    def test_carbon_pid_tracking_offline(self):
+
+        # Subprocess PIDs are childs, therefore booth should be equal
```
cianc: s/childs/children/
s/booth/both/

cianc:
Based on the test output it does look like the path that should be testing automatic inclusion of child pids is not based on how small its emissions are:

```
FAILED tests/test_pid_tracking.py::TestPIDTracking::test_carbon_pid_tracking_offline - AssertionError: 6.22814233144833e-05 != 0.005424063623566096 within 2 places (0.005361782200251613 difference)
```

Can you confirm on your machine that those children are definitely being included and it's not being hidden by the `assertAlmostEqual?`

Might be worth adding some debug logging to figure out what's going on with the child process accounting.

tests/test_pid_tracking.py
Matyro: Fixed

Matyro:
> Based on the test output it does look like the path that should be testing automatic inclusion of child pids is not based on how small its emissions are:
> 
> ```
> FAILED tests/test_pid_tracking.py::TestPIDTracking::test_carbon_pid_tracking_offline - AssertionError: 6.22814233144833e-05 != 0.005424063623566096 within 2 places (0.005361782200251613 difference)
> ```
> 
> Can you confirm on your machine that those children are definitely being included and it's not being hidden by the `assertAlmostEqual?`
> 
> Might be worth adding some debug logging to figure out what's going on with the child process accounting.

I changed the comparison to relative values, as I use a 256 Core machine power consumption is different compared to the unit test machine

Edit 1:
For whatever reasons GPUs are traced as well, this dominates power consumption. I am not sure why they are included as I do not activate GPU tracing and cuda_visible_devices is empty

Edit 2:
The problem is that I misunderstood the implementation. By default on my machine its using the rapl implementation and completely skips _get_power_from_cpu_load
I will modify the test somehow and check why the cpu_load functionality is not used with rapl. 
As rapl should only give results for hardware somewhere the transfer to processes must be done, currently not sure where

Matyro:
I may have found the issue:

https://github.com/mlco2/codecarbon/blob/b86a854c2bb4fe22ee236e05eac8268b8b2d42d7/codecarbon/core/resource_tracker.py#L64-L76

with power gadget, rapl and powermetrics tracking mode and pids are not passed. It defaults to "machine" which does not seem to be correct in my understanding

inimaz:
Thanks a lot for this @Matyro. Can I suggest to split this PR into 2 PRs?
-  Create a new PR with just the prometheus enhancement. 
- Leave this one for the multiple pids tracking.
Like this we discuss the 2 features separately, I am guessing the multiple pids is going to take more work

Matyro:
> Thanks a lot for this @Matyro. Can I suggest to split this PR into 2 PRs?
> 
>     * Create a new PR with just the prometheus enhancement.
> 
>     * Leave this one for the multiple pids tracking.
>       Like this we discuss the 2 features separately, I am guessing the multiple pids is going to take more work

https://github.com/mlco2/codecarbon/pull/1007

Matyro:
I intended to overwrite the history of this branch and remove all the prometheus stuff, instead it created the new branch https://github.com/mlco2/codecarbon/tree/pr-984

benoit-cty:
Thanks @Matyro, can you do the rebase and fix the conflict with mlco2:master ?

Also the test is flooding the output : https://github.com/mlco2/codecarbon/actions/runs/20552994440/job/59033265603?pr=984

Matyro:
Further discussion in https://github.com/mlco2/codecarbon/pull/1025
History was cleaned up to remove the Prometheus stuff

Issue #972, opened 2025-11-16
[Tests] Add unit tests to the output methods


inimaz:
For all the output methods listed in [here](https://github.com/mlco2/codecarbon/tree/master/codecarbon/output_methods) it will be interesting to add unit tests.

cianc:
I'm happy to work on this unless someone else wants it.

inimaz:
Go ahead! Thanks @cianc !

PR #988, opened 2025-11-27
Add unit tests for logfire and http output methods.

cianc:
## Description
Add unit tests for logfire and http output methods.
This requires adding logfire as a dev dependency so we can import it in tests. 
Also make it clearer which `EmissionsData` params are unnecessary for the `out` and `live_out` methods.

## Related Issue
Please link to the issue this PR resolves: #972 

## Motivation and Context
Increases test coverage.

## How Has This Been Tested?
Ran all existing and new tests.

## Types of changes

## Checklist:

### DIFF
--- a/codecarbon/output_methods/file.py
+++ b/codecarbon/output_methods/file.py
@@ -68,10 +68,10 @@ class FileOutput(BaseOutput):
                 # No entries
                 return True
             dict_from_csv = dict(csv_entries_list[0])
-            list_of_column_names = list(dict_from_csv.keys())
-            return list(data.values.keys()) == list_of_column_names
+            list_of_column_names = sorted(dict_from_csv.keys())
+            return sorted(data.values.keys()) == list_of_column_names
 
-    def out(self, total: EmissionsData, _: EmissionsData):
+    def out(self, total: EmissionsData, _):
         """
         Save the emissions data from a whole run to a CSV file.
 
--- a/codecarbon/output_methods/http.py
+++ b/codecarbon/output_methods/http.py
@@ -19,7 +19,7 @@ class HTTPOutput(BaseOutput):
     def __init__(self, endpoint_url: str):
         self.endpoint_url: str = endpoint_url
 
-    def out(self, total: EmissionsData, delta: EmissionsData):
+    def out(self, total: EmissionsData, _: EmissionsData):
         try:
             payload = dataclasses.asdict(total)
             payload["user"] = getpass.getuser()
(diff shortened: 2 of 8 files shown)


benoit-cty:
@cianc , can you confirm that every changes was merged by https://github.com/mlco2/codecarbon/pull/989 ?
Or did I do something wrong ?

cianc:
Not sure what happened here, maybe I messed pu my branches.
Looks like https://github.com/mlco2/codecarbon/pull/989/files contained all the files from this PR.

cianc:
One thing: I see this failure https://github.com/mlco2/codecarbon/actions/runs/19781662666/job/56682713399 complaining:

```
E   ModuleNotFoundError: No module named 'logfire'
```

This works in my local environment
```
cianc@fedora:~/src/codecarbon$ .venv/bin/python3
Python 3.14.0 (main, Oct 17 2025, 00:00:00) [GCC 15.2.1 20251022 (Red Hat 15.2.1-3)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import logfire
>>> logfire.__doc__
'**Logfire** is the observability tool focused on developer experience.'
```

I did add logfire to to dev dependencies in [pyproject.toml](https://github.com/mlco2/codecarbon/pull/989/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711), but maybe that wasn't enough?

inimaz:
Interesting issue!
Digging a bit, here there is a summary:
- The test [passed in the branch](https://github.com/mlco2/codecarbon/actions/runs/19765382107/job/56636773195) but failed in [master](https://github.com/mlco2/codecarbon/actions/runs/19781662666/job/56682713399).
- This is because in the master script we install `codecarbon` [directly from the wheel]( https://github.com/mlco2/codecarbon/blob/master/.github/workflows/package.yml#L58) so it does not have all the optional packages.  We did this to test how the package would behave when the user installs it via pypi/conda.

**TLDR**:  the test should fail in master as expected because there is no logfire unless the user installs it😄 Thanks @cianc for spotting this!

Solutions:

1. We can maybe edit the test to try to import logfire and if not present in the python modules skip the test?
2. We can flag what tests to not run when running directly the wheel
3. Any other ideas?

There is a similar issue with `prometheus` for instance and any other `output_method` that needs an external library. In the prometheus case we imported it into the `codecarbon` package but in retrospective I do not think it is the best way. It does not scale, if we implement 30 ouput_methods, we will not want the 30 libraries inside `codecarbon` when each end user just needs one or two.

cianc:
Another option: can we add logfire to this line where we're already installing `pandas`?
https://github.com/mlco2/codecarbon/blame/d0ff5e47eb5c7576774dfa38d097ae0cf78a7b25/.github/workflows/package.yml#L62

benoit-cty:
> There is a similar issue with `prometheus` for instance and any other `output_method` that needs an external library. In the prometheus case we imported it into the `codecarbon` package but in retrospective I do not think it is the best way. It does not scale, if we implement 30 ouput_methods, we will not want the 30 libraries inside `codecarbon` when each end user just needs one or two.

I totally agree : we need to have minimal dependencies for default install (reduce dependency hell and lower attack surface) and provide optional dependencies for output method.
And implement try..except in modules.

PR #995, opened 2025-12-03
Fix EmptyDataError when reading empty CSV files

isztldav:
## Description
This PR fixes a bug where `pandas.errors.EmptyDataError` is raised when the FileOutput class attempts to read an existing but empty CSV file. The fix adds error handling in the `out()` method to catch this exception and treat empty files as if they don't exist, creating a new file with the current emissions data.

## Related Issue
I have not opened an issue.

## Motivation and Context
This bug occurs when CSV emission files exist but are empty, which can happen if:
- A previous write operation was interrupted
- The file was manually created but never populated
- File system errors occurred during a write operation

Without this fix, the application crashes with `pandas.errors.EmptyDataError: No columns to parse from file` when trying to append or update emissions data, preventing users from tracking their carbon emissions.

## How Has This Been Tested?
1. **Unit Tests**: Added two new test cases in `tests/output_methods/test_file.py`:
   - `test_file_output_out_append_empty_file_exists()`: Verifies append mode handles empty files correctly
   - `test_file_output_out_update_empty_file_exists()`: Verifies update mode handles empty files correctly

2. **Manual Testing**:
   - Created empty CSV files and verified the fix handles them gracefully
   - Confirmed existing functionality remains unchanged with non-empty files

3. **Pre-commit Hooks**: Ran all linting and formatting checks

4. **Test Suite**: Ran the full test suite to ensure no regressions:
   ```bash
   uv run python -m unittest tests.output_methods.test_file.TestFileOutput
   ```

N/A - This is a bug fix with no UI changes

## Types of changes

## Checklist:

### DIFF
--- a/codecarbon/output_methods/file.py
+++ b/codecarbon/output_methods/file.py
@@ -88,6 +88,11 @@ class FileOutput(BaseOutput):
 
         """
         file_exists: bool = os.path.isfile(self.save_file_path)
+        if file_exists and os.path.getsize(self.save_file_path) == 0:
+            logger.warning(
+                f"File {self.save_file_path} exists but is empty. Treating as new file."
+            )
+            file_exists = False
         if file_exists and not self.has_valid_headers(total):
             logger.warning("The CSV format has changed, backing up old emission file.")
             backup(self.save_file_path)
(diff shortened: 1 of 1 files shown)


cianc:
If we changed line 90 from

```
file_exists: bool = os.path.isfile(self.save_file_path)
```

to something like
```
file_exists_and_non_empty = (os.path.isfile(self.save_file_path)) and (os.path.getsize(self.save_file_path) != 0)
```
then I think we could rely on the existing logic without adding more conditionals? Personally I already find the existing logic hard to follow.

isztldav:
good idea, done

PR #998, opened 2025-12-07
deploy: use uv in the docker image

inimaz:
## Description
The Dockerfile was not using uv, this meant that it was desynchronized with the changes in the `pyproject.toml` file. 
After this PR, the requirements-api.txt is only used in the CI to deploy to Clever Cloud. Since it is generated via the CI, we no longer need it in the repo, we delete it.


SaboniAmine (approved):
Thanks!

Issue #861, opened 2025-06-11
Migrate from Fief ?


benoit-cty:
The Dashboard use Fief for authentication and it will stop to be maintained : https://github.com/orgs/fief-dev/discussions/543

It put our user at risk in the long term, maybe we have to switch to another tool ?

digicosmos86:
Is there any update on this issue? fief-client's `httpx` version requirement is in conflict with other packages that we are using

benoit-cty:
Thanks for letting us know about this problem. Indeed Fief-client specify `httpx >=0.21.3,<0.28.0`.

I think the best way is to move `fief-client[cli]` from our main dependencies to a new optional dependency `codecarbon[api]`. So we have to rename actual `api `to `api-server`.

But this will break all our existing API users.

@SaboniAmine do you have a better idea ?

@digicosmos86 can you give more info on your context ?
A possible immediate solution is to create a `constraints.txt` file:
`httpx>=0.29.0`

Install dependencies with the constraint:

`pip install -r requirements.txt -c constraints.txt`

digicosmos86:
Thank you @benoit-cty for looking into this. Here's a bit more extra context: we are building something that uses both the Gemini API and the Whisper model, and in the container that runs app, we use codecarbon to calculate the carbon emission. The Gemini SDK is the one that depends on higher versions of httpx, which conflicts with `fief-client`.

We are using a similar solution inspired by your temporary solution: we put codecarbon and the gemini sdk in separate extra groups, and tell `uv` that there's a conflict, so `uv` resolves these groups separately. Fortunately in our use case the gemini sdk can be completely separate from `codecarbon` though. I wonder if there is a way to force the higher version of `httpx` and override what `fief-client` requires in `codecarbon`'s dependency definitions?

PR #999, opened 2025-12-07
feat: allow any authentication provider + fief OIDC

inimaz:
## Description
Instead of using directly Fief, we can pass an auth_provider interface that uses OIDC and points to Fief. 

Two main gains:
1. Like this migrating to other auth providers will be easier. If they use OIDC, update the env vars. If they do not use OIDC they will need to implement the `auth_provider` interface.
1. We can use this interface as well to provide a no-authentication provider 😄 . See 
`carbonserver/carbonserver/api/services/auth_providers/no_auth_provider.py`, for testing purposes. This will allow the UI in local to be easier to test/develop.

Related to https://github.com/mlco2/codecarbon/issues/861

### DIFF
--- a/carbonserver/tests/api/integration/test_auth_cookie_flow.py
+++ b/carbonserver/tests/api/integration/test_auth_cookie_flow.py
@@ -1,0 +1,158 @@
+"""
+End-to-end auth flow against a running API using Fief + fastapi-oidc.
+
+Required environment variables (test is skipped if any is missing):
+    CODECARBON_API_URL        Base API URL (e.g. https://api.dev/codecarbon/api)
+    FIEF_TEST_USER_EMAIL      Login email for the test user
+    FIEF_TEST_USER_PASSWORD   Login password for the test user
+    CODECARBON_TEST_ORG_ID    Organization ID in which to create the project/experiment
+
+To run this:
+    export $(cat ./carbonserver/tests/api/integration/.env | xargs) && uv run pytest carbonserver/tests/api/integration/test_auth_fief_cookie_flow.py -vv
+"""
+
+import os
+import uuid
+from html.parser import HTMLParser
+from urllib.parse import urljoin
+
+import pytest
+import requests
+
+SESSION_COOKIE_NAME = "user_session"
+
+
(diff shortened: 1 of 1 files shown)


prmths128:
do we need multiple auth providers ?
the short term issue is not the fief server but the reliance on custom libs instead of generic oidc 
i'd suggest  instead keeping the fief server for the moment but switching the fief libs to  fastapi_oidc on the back and next-auth on the front might 
and then using another oidc server

inimaz:
> do we need multiple auth providers ?

Thanks for the comment @prmths128 ! I agree we won't need 2 auth providers ever at the same time. On the other hand if we can implement the fastapi_oidc so that the server is injected and easily plugable? Like that for local dev we inject a mock and it is easier to work with it. What do you think?

carbonserver/carbonserver/config.py
```diff
@@ -6,9 +6,29 @@ class Settings(BaseSettings):
...
-    fief_client_id: str = Field("", env="FIEF_CLIENT_ID")
-    fief_client_secret: str = Field("", env="FIEF_CLIENT_SECRET")
-    fief_url: str = Field("https://auth.codecarbon.io/codecarbon-dev", env="FIEF_URL")
+    # Authentication provider settings
+    auth_provider: str = Field(
+        "oidc", env="AUTH_PROVIDER"
+    )  # Options: 'oidc', 'fief' (deprecated), 'none'
+
+    # OIDC settings (with backward compatibility for Fief environment variables)
+    oidc_client_id: str = ""
```
SaboniAmine, line 15: lets maybe source from env vars with default values ?

inimaz: It is in the config part. I did it like this because we need to still read the `FIEF_...` env vars.

So in this configuration:
https://github.com/mlco2/codecarbon/pull/999/changes#diff-24a79c91033797e992d55fd4eddbcda1cff587119f27fe3cd7d2c739bb616ab8R43

The precedence is:

- First checks OIDC_CLIENT_ID
- If not found, then checks FIEF_CLIENT_ID
- If neither found, uses the default value (empty string "")

PR #1002, opened 2025-12-11
fix(api): update uv lock + requirements-api.txt

inimaz:
## Description
Update uv lock + requirements-api.txt


inimaz:
Not needed anymore, see #1006

PR #1004, opened 2025-12-12
feat: codecarbon monitor -- any_command

inimaz:
## Description

Add to the cli the ability to run codecarbon with **any other tool**.
```
codecarbon run -- bash -c 'echo "Processing..."; sleep 20; echo "Done!"'
```

# Output in my console
```bash
~/: uv run codecarbon run -- bash -c 'echo "Processing..."; sleep 30; e
cho "Done!"'

🌱 CodeCarbon: Starting emissions tracking...
   Command: bash -c echo "Processing..."; sleep 30; echo "Done!"

Processing...
Done!

============================================================
🌱 CodeCarbon Emissions Report
============================================================
   Command: bash -c echo "Processing..."; sleep 30; echo "Done!"
   Emissions: 0.0317 g CO2eq
   Saved to: /home/inigo/personal/codecarbon/emissions.csv
   ⚠️  Note: Measured entire machine (includes all system processes)
============================================================
```
## Motivation and Context

Non-python users will be able to run their commands and measure the machine emissions while running those commands.

## Basic Usage

```bash
codecarbon run -- <command> [arguments...]
```

The `--` separator is required to separate CodeCarbon options from your command.

## Simple Examples

- Track a bash one-liner:
```bash
codecarbon run -- bash -c 'echo "Processing..."; sleep 3; echo "Done!"'
```

- Track a CPU-intensive calculation:
```bash
codecarbon run -- bash -c 'for i in {1..1000}; do echo "scale=100; a(1)*4" | bc -l > /dev/null; done; echo "Calculation complete!"'
```

- Track a Node.js script:
```bash
codecarbon run -- node my_script.js
```

- Track a Python script (although better to use the EmissionTracker directly):
```bash
codecarbon run -- python examples/mnist.py
```
- Track file compression:
```bash
codecarbon run -- tar -czf archive.tar.gz large_directory/
```

- Track image processing (requires ImageMagick):
```bash
codecarbon run -- convert input.jpg -resize 50% output.jpg
```

- Track video encoding (requires ffmpeg):
```bash
codecarbon run -- ffmpeg -i input.mp4 -c:v libx264 output.mp4
```

- Any command you can run on the cli
```
codecarbon run -- YOUR_COMMAND_HERE
```
## Types of changes
- [x ] New feature (non-breaking change which adds functionality)

## Checklist:
- [x ] My code follows the code style of this project.
- [ x] My change requires a change to the documentation.

### DIFF
--- a/codecarbon/cli/monitor.py
+++ b/codecarbon/cli/monitor.py
@@ -79,12 +79,9 @@ def run_and_monitor(
     process = None
     try:
         # Run the command, streaming output to console
-        process = subprocess.Popen(
-            command,
-            stdout=sys.stdout,
-            stderr=sys.stderr,
-            text=True,
-        )
+        # Let the child inherit the parent's std streams so Click's
+        # `CliRunner` can capture output (don't pass StringIO objects).
+        process = subprocess.Popen(command, text=True)
 
         # Wait for completion
         exit_code = process.wait()
--- a/tests/test_cli_main.py
+++ b/tests/test_cli_main.py
@@ -1,0 +1,81 @@
+"""Tests for the CodeCarbon CLI main function."""
+
+from typer.testing import CliRunner
+
+from codecarbon.cli import main as cli_main
+
+
+class FakeApiClient:
+    def __init__(self, endpoint_url=None):
+        self.endpoint_url = endpoint_url
+        self.token = None
+
+    def set_access_token(self, token):
+        self.token = token
+
+    def get_list_organizations(self):
+        return [{"id": "1", "name": "fake-org"}]
+
+
+def fake_get_access_token():
+    return "fake-token"
+
+
+def test_version_flag():
+    runner = CliRunner()
+    result = runner.invoke(cli_main.codecarbon, ["--version"])
+    assert result.exit_code == 0
+    assert cli_main.__app_name__ in result.output
+    assert str(cli_main.__version__) in result.output
+
+
+def test_api_get_calls_api_and_prints(monkeypatch):
+    runner = CliRunner()
+    monkeypatch.setattr(cli_main, "ApiClient", FakeApiClient)
+    monkeypatch.setattr(cli_main, "_get_access_token", fake_get_access_token)
+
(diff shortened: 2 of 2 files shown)


benoit-cty:
Nice !
It will close https://github.com/mlco2/codecarbon/issues/382

Did you have a look to https://github.com/mlco2/codecarbon/pull/431 ?

Do we need to add a warning about code execution ? Maybe not as the command is for this purpose and won't change user privilege ?

inimaz:
Ah sorry I was not at all aware of https://github.com/mlco2/codecarbon/pull/431! Checking it now

inimaz:
> Do we need to add a warning about code execution ? Maybe not as the command is for this purpose and won't change user privilege ?

@benoit-cty it launches exactly what the user wants to launch, so I guess no need for a warning? If it is unsafe code it is unsafe with or without `codecarbon run`

benoit-cty:
@copilot  Please add a documentation in `docs/edit/usage.rst` for this new feature.

benoit-cty:
I was about to merge this PR in the 3.2.2 release but with all my modification for sub-process it's more a 3.3.

I think it's better to have sub-process monitor for this PR as it make sense to monitor only the program in parameter.

What do you think ?

inimaz:
Sounds good @benoit-cty , does this mean we can monitor now exactly the process and the other processes do not affect the measurement?
Another topic, just about semantics. I proposed `codecarbon run --`  but now thinking about it, should we use instead `codecarbon monitor --`? Since codecarbon is just monitoring what is being run and not editing anything...

benoit-cty:
> Sounds good @benoit-cty , does this mean we can monitor now exactly the process and the other processes do not affect the measurement?

Process monitoring works for RAM and CPU, not for GPU.

> Another topic, just about semantics. I proposed `codecarbon run --` but now thinking about it, should we use instead `codecarbon monitor --`? Since codecarbon is just monitoring what is being run and not editing anything...

Yes, I'm OK with that proposition 👍

inimaz:
Sounds good, will take a look at it tmw

PR #1018, opened 2026-01-03
Remove fief lib from frontend

prmths128:
## Description
Removes fief lib from frontend.
=> Auth only uses the session cookie from the auth server.
=> Might impact some minor features (like the user profile page)
=> Could break all pages rendered server side (none?)


webapp/src/middleware.ts
```diff
@@ -22,5 +12,5 @@ export async function middleware(request: NextRequest) {
         return;
     }
 
-    return authMiddleware(request);
+//     return authMiddleware(request);
```
SaboniAmine: to remove

PR #1025, opened 2026-01-12
Multiple pid tracking

Matyro:
## Description
This is the followup of !984

Added the functionality to track several process ids and its children
This was implemented by isolating CPU relevant parts from external/hardware into cpu/hardware and modifying the energy and power functions to scale there return value with CPU Utilization.

## Related Issue
-

## Motivation and Context
This enables emission data logging of other processes. In my case its the automatic tracking and reporting of energy consumption to users in a slurm cluster environment.

Currently if given pids are childs of each other statistics will be double counted

## How Has This Been Tested?
Changes will be texted on a testcluster

## Types of changes

## Checklist:


Matyro:
https://github.com/mlco2/codecarbon/blob/c8582bae212887259fa2586312632a1dff3eea4e/codecarbon/emissions_tracker.py#L925-L927

Should this be always machine metrics or depending on tracking mode?

davidberenstein1957:
Hi @Matyro, apologies for the long silence on this one.

Since it was opened, #1004 rewrote `_get_power_from_cpu_load` and added child-process CPU aggregation via `cpu_times()` deltas, so master now covers part of what this does, and the branch conflicts across most of the files it touches. The CI logs from the last run have aged out of retention, so we cannot even tell what the 3.13/3.14 failures were.

Closing to keep the queue honest rather than because the idea was wrong. The piece we would still genuinely like is tracking arbitrary external PIDs, which master still cannot do — it only walks the children of the current process. Given the overlap, a fresh, smaller PR adding just that on top of the existing child-walking code would likely be less work than rebasing this one, and much easier to review. Very happy to look at it if you are still interested.

Issue #994, opened 2025-12-03
[Co2 Signal] compute emissions on a regular basis


benoit-cty:
**Is your feature request related to a problem? Please describe.**

By default the carbon emissions are computed at the exit. 
When using CO2 Signal with a process running for many hours you only get the carbon intensity at the end. So we loose the variations. 

**Describe the solution you'd like**

Always compute emissions in the same place we call our API.

PR #1050, opened 2026-02-04
Compute emissions based on ongoing carbon intensity

cianc:
## Description
Compute emissions based on ongoing carbon intensity samples, rather than just taking the latest value and applying it to all energy consumption every time.

## Related Issue
Please link to the issue this PR resolves: #994

## Motivation and Context
Improves emission accuracy over long runs.

## How Has This Been Tested?
Added a new test and ran all existing tests.

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_emissions_tracker.py
+++ b/tests/test_emissions_tracker.py
@@ -657,8 +657,6 @@ class TestCarbonTracker(unittest.TestCase):
         self.assertIn("gpu_model", hardware_info)
         self.assertIn("gpu_ids", hardware_info)
 
-
-class TestVariableEmissions(unittest.TestCase):
     @mock.patch("codecarbon.emissions_tracker.EmissionsTracker._get_geo_metadata")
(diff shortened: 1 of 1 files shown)


benoit-cty:
Thank you !
Because some people are using CodeCarbon with short tasks, the potential problem I see is the time it will take to do it. Maybe it will be better to have something like a cache with a 5 minutes duration, what do you think ?

cianc:
> Thank you ! Because some people are using CodeCarbon with short tasks, the potential problem I see is the time it will take to do it. Maybe it will be better to have something like a cache with a 5 minutes duration, what do you think ?

Could you elaborate on this, I think I'm missing something?  Are you referring to the time it takes to call `electricitymaps_api.get_emissions`? I don't think this PR  changes the number of these calls, it just changes the accounting.

benoit-cty:
Sorry, you are absolutely right, the `def _update_emissions` is just calling the same function as before.

Perf improvement is another subject.

benoit-cty (approved):

PR #1055, opened 2026-02-08
fix(api): remove logfire from api server

inimaz:
There is an error in the API server that keeps popping up.
# The error

```
2026-02-08T14:54:01.571Z
Caught an internal error in Logfire. Your code should still be running fine, just with less telemetry. This is just logging the internal error.

...

AttributeError: module 'pydantic' has no attribute 'RootModel'
```
# What it means
Logfire is using `pydantic v2` whereas the `carbonserver` is using `pydantic v1`. They are not compatible.

# The solution
Remove `logfire` from the installed packages (it was not running anyhow). We will reactivate it when we migrate to pydantic v2.

### DIFF
--- a/carbonserver/carbonserver/api/services/signup_service.py
+++ b/carbonserver/carbonserver/api/services/signup_service.py
@@ -1,17 +1,8 @@
 import logging
-from contextlib import nullcontext
 from uuid import UUID
 
 import jwt
 from fastapi import HTTPException
-
-try:
-    import logfire
-
-    LOGFIRE_AVAILABLE = True
-except (ImportError, AttributeError) as e:
-    LOGFIRE_AVAILABLE = False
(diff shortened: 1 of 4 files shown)


SaboniAmine:
Thanks, logfire free plan was also deprecated. However I think we could migrate to pydantic v2 without breaking change, could you test it from this PR ?

inimaz:
Info about the migration to pydantic v2 --> https://github.com/mlco2/codecarbon/issues/1056

SaboniAmine (approved):

SaboniAmine:
OK lets do it in a separate PR!

inimaz:
Just saw your comment after @SaboniAmine sorry!

> logfire free plan was also deprecated

 then I can remove logfire entirely n?

PR #1061, opened 2026-02-15
fix: starlette to compatible version

inimaz:
## Description
Starlette needs to follow the available version of Fastapi. Otherwise it breaks


SaboniAmine (commented):
we need to pin it also in the pyproject.toml
and maybe upgrade fastapi / starlette (the 1.0.0 version is soon to be published, currently rc1 is in progress: https://github.com/Kludex/starlette/pull/3145)

prmths128 (approved):

PR #1063, opened 2026-02-15
chore: remove fief lib from cli

prmths128:
switch to auth lib

### DIFF
--- a/codecarbon/cli/main.py
+++ b/codecarbon/cli/main.py
@@ -1,15 +1,20 @@
+import json
 import os
 import signal
 import sys
 import time
+import webbrowser
+from http.server import BaseHTTPRequestHandler, HTTPServer
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


codecarbon/cli/main.py
```diff
@@ -393,17 +493,11 @@ def monitor(
             )
             raise typer.Exit(1)
 
-        tracker_args = {**tracker_args, "save_to_api": api}
-
-    # If extra args are provided (e.g. `codecarbon monitor -- my_script.py`), delegate to `run_and_monitor`
-    if getattr(ctx, "args", None):
-        return run_and_monitor(ctx, **tracker_args)
```
inimaz, line 400: Watch out! You are removing the `codecarbon monitor -- myscript.py` here. Maybe do not update the `monitor` command

Issue #820, opened 2025-04-13
API client to throw error instead of None


inimaz:
I think we can merge this one and create a separate issue to reply in the client `raise_for_status()` instead of `None` if there is an error. Like this we can call the API client and give a more meaningful error as @SaboniAmine suggested! What do you think?

_Originally posted by @inimaz in https://github.com/mlco2/codecarbon/issues/798#issuecomment-2800035595_

bcmyguest:
Would be happy to take this one if it's still relevant :)

bcmyguest:
Noticed a couple of other things when trying to get env setup. 

- project could benefit from using `uv` for dependency resolution (on install and pip compile)

eg. (not sure this is complete, more info [here](https://juftin.com/hatch-pip-compile/#configuration-options)

```
# pyproject.toml

[tool.hatch.envs.default]
# pip-compile automate the export of environments.txt
type = "pip-compile"
installer = "uv"
pip-compile-resolver = "uv"
```

- python3.8 tests do not work as installing `urllib==2.5.0` is not possible on py3.8
- test deps don't seem to be separated from build deps
- setting `CODECARBON_ALLOW_MULTIPLE_RUNS` env variable could be done with `pytest-dotenv` setting in `pytest.ini` and adding a `.env.test` file 

All except the second one are somewhat debatable, so not sure the best place to put them - happy to elaborate.

inimaz:
Hello @bcmyguest It is still relevant! Feel free to take it/ask for any questions.

- About the use of `uv`, I agree. I see @benoit-cty already started [a branch to migrate ](https://github.com/mlco2/codecarbon/pull/815) maybe it can be discussed in that PR?
- About test deps not separated from build deps. I think this is only true for the env `api`. Thanks for pointing out. We use it to deploy our API, we could separate it into `api-test` and `api`.
- And about the `pytest.ini`, indeed go ahead!

benoit-cty:
Hello,
Yes, we would like to switch to UV. I stopped my PR has I don't know how to migrate some part, which I do not write done :-(. I can do the rebase tomorrow and check what is left to do.

Thanks for contributing !

benoit-cty:
Hello @bcmyguest I've made the rebase if you want to have a look at https://github.com/mlco2/codecarbon/pull/815/

sankalpsthakur:
I'd like to take this — will make the API client raise on HTTP errors and catch in the emissions tracker.

PR #1089, opened 2026-02-27
fix(api_client): raise HTTPError instead of returning None on API errors

AbdelhamidKhald:
## Summary

Relates to #820

## Problem

`ApiClient` methods silently return `None`, `False`, or `[]` when API calls fail. This makes it impossible for callers to distinguish between "API returned an error" and "no data exists", leading to confusing `TypeError` exceptions downstream (e.g., iterating over `None`).

## Solution

Added `response.raise_for_status()` at the end of `_log_error()`, so all API methods now raise `requests.exceptions.HTTPError` after logging the error details. Removed the now-unreachable `return None/False/[]` statements.

### Safety

All existing callers already handle exceptions properly:
- `CodeCarbonAPIOutput.live_out()` / `.out()` — wrapped in `try/except Exception`
- `_create_run()` — has `except requests.exceptions.ConnectionError` and `except Exception`
- `add_emission()` — has `except Exception`
- CLI commands — all go through `main()` which has a global `try/except Exception`

## Changes

- `codecarbon/core/api_client.py`: Added `response.raise_for_status()` to `_log_error()`; removed 13 unreachable return statements and 1 dead `None` guard
- `tests/test_api_call.py`: Added `test_call_api_error` verifying `HTTPError` is raised on 500/400 responses

## Testing

- All 11 existing + new tests pass
- `black` formatting: clean
- `ruff` linting: clean
- Verified all caller exception handling paths

## Checklist

- [x] Tests pass locally
- [x] Linter passes
- [x] Follows contribution guidelines
- [x] Changes are minimal and focused


SaboniAmine (approved):
Hello, thanks for your contribution!
This new behavior seems a good addition.
Ideally, the error raised would be useful for the user to understand what went wrong. In this case, I would have expected that the raised error would be 404 in no orgnanization are found server-side, for instance. How is your contribution dealing with error codes ?

sankalpsthakur:
Heads-up: opened a refreshed attempt in #1277 for the same issue (#820), including EmissionsTracker try/except so the tracker doesn't break when the API client raises (called out in the issue). Happy to close #1277 if you'd rather revive this PR instead — just say the word.

benoit-cty:
Thanks, closing in favor of #1277

Issue #1090, opened 2026-02-27
Add type annotations


ArmaanjeetSandhu:
CodeCarbon currently ships without type annotations and without a `py.typed` marker, which means static type checkers cannot provide type checking, autocompletion, or inline documentation for CodeCarbon code. Adding type support would be a meaningful quality-of-life improvement.

### What I'm proposing

I'd like to contribute inline type annotations directly to CodeCarbon's source files, following [PEP 484](https://peps.python.org/pep-0484/) and [PEP 561](https://peps.python.org/pep-0561/). Concretely, this would involve:

- Annotating function signatures and class attributes across CodeCarbon's public API, starting with the most commonly used surfaces.
- Adding an empty `codecarbon/py.typed` marker file, which signals to type checkers that the package supports typing.
- Ensuring the annotations are verified by running mypy and/or pyright in CI, so they don't silently drift out of sync with the implementation.

I plan to use `stubgen` (from mypy) to generate an initial skeleton, then refine the signatures by hand, particularly for the more complex parts of the API that involve union types or overloads.

A separate `types-codecarbon` package on PyPI is an alternative, but inline annotations are generally the preferred long-term solution because the types live alongside the code and stay in sync naturally with each release. A separate package creates a maintenance burden where type stubs can lag behind the library's actual API.

### Before I start

I wanted to open this issue first to check a few things:

1. Are the maintainers open to this change in principle?
2. Is there a preference for inline annotations vs. a separate stub package?
3. Are there any parts of the codebase that might need special treatment?

Happy to start with a focused PR covering just the core public API and expand from there based on feedback. Please do let me know your thoughts!

inimaz:
Hello @ArmaanjeetSandhu this will be very helpful indeed!
Some answers:
>Is there a preference for inline annotations vs. a separate stub package?

inline annotations

> Are there any parts of the codebase that might need special treatment?

I suggest you go first and add the CI mypy check that ignores everything in a first PR. And then create subsequent PRs unignoring the files you added types to. Like this you do not have to handle a big PR and will be faster to merge + less conflicts to handle on your side. What do you think?

inimaz:
Reopening this as I think it is a good idea

ArmaanjeetSandhu:
> Hello [@ArmaanjeetSandhu](https://github.com/ArmaanjeetSandhu) this will be very helpful indeed! Some answers:
> 
> > Is there a preference for inline annotations vs. a separate stub package?
> 
> inline annotations
[...]

PR #1093, opened 2026-03-01
chore(types): add py.typed marker

ArmaanjeetSandhu:
Adds a [PEP 561](https://peps.python.org/pep-0561/) `py.typed` marker file to the package, in reference to #1090 

To confirm the marker is present in the installed package, run:
```bash
uv run python -c "import codecarbon; import pathlib; p = pathlib.Path(codecarbon.__file__).parent / 'py.typed'; print(p, p.exists())"
```


benoit-cty (approved):
Thanks, I did not know about `py.typed`.

PR #1098, opened 2026-03-05
chore(api): separate api and package pyproject

inimaz:
## Description
The main `pyproject.toml` contains references to the `carbonserver` deps (the API).
This PR aims to separate both dependencies. Like this any restrictions on the package do not apply in the server and viceversa.


SaboniAmine (approved):
Thanks Inigo!

PR #1103, opened 2026-03-13
fix: fallback to backup API when geojs returns empty region

### LINKED ISSUE #1101: Add fallback IP to region providers
**Is your feature request related to a problem? Please describe.**

The current IP to region provider `geojs.io` is unable to return a region for some IPs, like in this example where `region` is missing: https://app.geojs.io/s/206.12.125.2

```bash
curl https://get.geojs.io/v1/ip/geo.json
{
  "accuracy": 1000,
  "area_code": "0",
  "asn": 11105,
  "continent_code": "NA",
  "country": "Canada",
  "country_code": "CA",
  "country_code3": "CAN",
  "ip": "206.12.125.2",
  "latitude": "43.6319",
  "longitude": "-79.3716",
  "organization": "AS11105 Simon Fraser University",
  "organization_name": "Simon Fraser University",
  "timezone": "America\/Toronto"
}
```

**Describe the solution you'd like**

Offer a fallback IP to region provider such as `ipinfo.io` which could return when `geojs.io` failed to do so

```bash
curl ipinfo.io
{
  "ip": "206.12.125.2",
  "city": "Pitt Meadows",
  "region": "British Columbia",
  "country": "CA",
  "loc": "49.2212,-122.6897",
  "org": "AS11105 Simon Fraser University",
  "postal": "V3Y",
  "timezone": "America/Vancouver",
  "readme": "https://ipinfo.io/missingauth"
}
```

**Describe alternatives you've considered**

I have considered going to MaxMind who provides data to `geojs.io` but that does not seem scalable if every unmapped IP ranges have to do the same when other IP to region provider provider already good data.

**Additional context**

One thing to consider is that CodeCarbon needs the 3 letter country code to properly map things internally and I could not find any other provider that could give the 3 letter code in a single call. Therefore it would seem we would have to consider mapping from 2 letters to 3 letters ourselves in case it is missing.

Additionally, if this is something the CodeCarbon project would consider to add as a feature, I volunteer to implement it.

ucodia:
## Description
Updated `GeoMetadata.from_geo_js()` to handle responses where the `geojs` API returns a successful HTTP 200 response but with an empty [region](cci:1://file:///Users/ucodia/code/codecarbon/codecarbon/external/geography.py:26:8-31:64) field. When this happens, it now logs a warning and falls through to the existing `ip-api.com` backup chain instead of returning incomplete metadata.

## Related Issue
Resolves #1101

## Motivation and Context
For certain IP ranges, `get.geojs.io` returns a successful lookup but with no region field. Because no exception was raised, from_geo_js() previously returned this incomplete data immediately. Downstream, this caused `get_region_emissions()` to perform lookups with an empty string, triggering `KeyError: ''` and falling back to country-level emissions with incorrect geographic coordinates logged.

## How Has This Been Tested?
- Added `GEO_METADATA_CANADA_EMPTY_REGION` test fixtures in tests/testdata.py.
- Added a new unit test test_geo_metadata_empty_region_falls_back_to_backup in to verify that an empty region correctly triggers the fallback.
- Ran `uv run pytest tests/test_geography.py -v` successfully.
- Verified that all other existing test cases continue to pass unaffected.

N/A

## Types of changes

## Checklist:


ucodia:
@benoit-cty after testing this in my production environment I discovered a few issues with the existing fallback mechanism.

### Issue 1

The fallback API will always fail when using the HTTPS endpoint

```
curl https://ip-api.com/json/
{"status":"fail","message":"SSL unavailable for this endpoint, order a key at https://members.ip-api.com/"}%
```

Using the HTTP endpoint is not great because this could expose the application location to eavesdropper and be vulnerable to man-in-the-middle.

### Issue 2

If the fallback API call fails, it reports on the wrong URL on the exception: https://github.com/ucodia/codecarbon/blob/ebc8d1a3f5dcae9013dcbe6883d05e711ad6ce28/codecarbon/external/geography.py#L136

### Proposal

After doing some research I found that `ipinfo.io` offers compliant IP geolocation
- Provides region information
- HTTPS without API key
- Allowed in commercial software

Similarly to previous fallback provider it does not provide the 3 letter country code but instead of calling yet another API, I suggest we add a dependency on `pycountry`. Also since this is static data that practically never changes we did not need an API for that in the first place.

If you are ok with the suggested new path, I can close this PR and open a new issue to properly track this.

ucodia:
@benoit-cty created a new PR for the proposed changes and updated the original issue with context: https://github.com/mlco2/codecarbon/pull/1104

Issue #404, opened 2023-05-05
Too many variables when initializing the EmissionsTracker


inimaz:
### Description
The BaseEmissionsTracker accepts 22 variables when initialized (and growing). This could lead to potential issues.  And it is a known code smell.

### What I Propose
Refactor the input variables to accept a config dict object as input.

So instead of
```
EmissionsTracker(project_name: "My project", measure_power_secs=10, output_dir="./temp")
```
codecarbon could do
```
config_options = {
    project_name: "My project",
    measure_power_secs=10,
    output_dir="./temp",
}
EmissionsTracker(config_options)
```

benoit-cty:
Or maybe we could use [kwargs](https://book.pythontips.com/en/latest/args_and_kwargs.html) ?

essp-lbg:
@inimaz im interested in making this contribution.  do you have any example config which I could use as guidance?

PR #1161, opened 2026-04-16
Potential fix for #404 Too many variables when initializing the EmissionsTracker

Jmilk3:
## Description
Modifies EmissionsTracker __init__ functions to use **kwargs to address the large number of input arguments as suggested in issue #404. The implementation seeks to avoid changing anything about how the code actually runs while clearing the code smell.

## How Has This Been Tested?
All existing tests still pass with no modifications after the changes to EmissionsTracker were implemented. All pre-commit linters/autoformatters passed.

## Types of changes

## AI Usage Disclosure

- [x] ♻️ No AI used. Car analogy : you drive the car.

## Checklist:

## Notes:
I am unsure if/how the documentation should be updated to reflect the changes made to the argument input format. I would be happy to make those changes if needed.

This is my first attempt at contributing to an open source project, so any feedback would be greatly appreciated!

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -12,7 +12,7 @@ import uuid
 from abc import ABC, abstractmethod
 from datetime import datetime
 from functools import wraps
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional
 
 import psutil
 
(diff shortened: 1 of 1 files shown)


davidberenstein1957:
Closing this one on scope rather than on the code.

#404 is about the number of parameters `EmissionsTracker.__init__` takes, and the fix for it is a design decision — most likely a config dataclass or similar — that needs to be agreed before an implementation can be reviewed on its merits. A refactor of the constructor evaluated in isolation is hard for a reviewer to say yes or no to, because the question is not whether this particular shape works but which shape we want.

If you would like to drive that, a comment on #404 proposing the approach would be the more useful next step, and the implementation follows easily once there is agreement.

PR #1164, opened 2026-04-19
docs: explain electricitymaps_api_token and query behaviour

### LINKED ISSUE #403: Documentation is unclear about co2_signal_api_token functionality
### Description

In the CodeCarbon documentation, you can see that you can add a `co2_signal_api_token` to your `.codecarbon.config` (on the "Parameters" page). However, I have found no indication of what effect this has and why I would want to do this in the entire documentation.

My assumption is the following:
If I don't add an API key, CodeCarbon will use estimates for the carbon intensity of a particular country. If I do add an API key, CodeCarbon will query the current (more accurate) carbon intensity via the API.

**Question 1:** Is this assumption correct?
**Question 2:** If it is correct, how often is the API queried? For example, if I start tracking in a Jupyter Notebook with `tracker.start()`, then train a model for 12 hours, then stop tracking with `tracker.stop()`, will CodeCarbon query the carbon intensity once at the init, or will it query the intensity once per hour and then average the values over the entire run? Sometimes models are trained for multiple days or even weeks, and if the carbon intensity is only queried once *not* using the API may actually be more accurate.

hamza-mobeen:
Closes #403

The `electricitymaps_api_token` parameter was mentioned in config examples but never documented. This PR adds a dedicated section to `configuration.md` answering the questions raised in the issue:

- What the token does: real-time carbon intensity vs static annual averages
- How often the API is queried: at stop, and every ~2 minutes during long runs
- Where to get a free token
- Notes that `co2_signal_api_token` is the deprecated predecessor

### DIFF
--- a/docs/how-to/configuration.md
+++ b/docs/how-to/configuration.md
@@ -84,6 +84,44 @@ Yields attributes:
     which relies on the [INI
     syntax](https://docs.python.org/3/library/configparser.html#supported-ini-file-structure).
 
+## Electricity Maps API Token
+
+By default, CodeCarbon estimates carbon intensity using annual country-level
+averages from Our World in Data. Providing an `electricitymaps_api_token`
+upgrades this to **real-time carbon intensity** data from the
+[Electricity Maps API](https://api.electricitymaps.com), giving more accurate
+emissions figures — especially useful if your grid's energy mix varies
+throughout the day.
+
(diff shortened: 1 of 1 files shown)


inimaz (approved):
Nice @hamza-mobeen  thanks for this addition!

PR #1169, opened 2026-04-28
Fix: Project dashboard not updating emissions data

Mounesh-13:
Problem
The Project dashboard does not update emissions data dynamically, while the Home dashboard reflects updates correctly.

Root Cause
The Home tab uses useSWR, which provides automatic polling and revalidation.
The Project tab relied on useEffect, resulting in stale data unless the page was manually refreshed.

Solution

Replaced useEffect-based data fetching with useSWR
Added refreshInterval (60s) for automatic updates
Used dynamic SWR keys (projectId + date range) to ensure proper re-fetching
Kept aggregation logic in the frontend using useMemo for performance

Improvements

Added null-safe reducers (?? 0) to prevent NaN issues from malformed API data
Introduced a user-facing error state for failed data fetching
Aligned Project dashboard behavior with Home dashboard for consistency

Testing

Verified automatic refresh (updates within 60 seconds)
Verified date range filter triggers re-fetch
Verified data updates after switching browser tabs (focus revalidation)
Verified correct behavior for empty projects (no NaN values)
Confirmed no console errors or UI regressions

This change ensures the Project dashboard stays in sync with real-time emissions data and improves overall reliability.

### DIFF
--- a/webapp/src/app/(dashboard)/[organizationId]/projects/[projectId]/page.tsx
+++ b/webapp/src/app/(dashboard)/[organizationId]/projects/[projectId]/page.tsx
@@ -8,16 +8,14 @@ import {
     getEquivalentTvTime,
 } from "@/helpers/constants";
 import { getDefaultDateRange } from "@/helpers/date-utils";
-import {
-    getExperiments,
-    getProjectEmissionsByExperiment,
-} from "@/server-functions/experiments";
-import { getOneProject } from "@/server-functions/projects";
+import { fetcher } from "@/helpers/swr";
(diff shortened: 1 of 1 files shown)


inimaz:
Hello @Mounesh-13 thanks for taking care of this, and sorry for the late review.
I see there are some conflicts with master, could you pull them in and resolve them?
Thanks again!

davidberenstein1957:
Closing after no response to the conflict-resolution request from early August — happy to reopen if you pick it back up.

If the dashboard genuinely is not refreshing emissions data, that is worth fixing, so if you are still around please do reopen with the branch rebased, or file an issue describing the reproduction and we can take it from there.

PR #1201, opened 2026-05-19
fix: amdsmi import when libamd_smi is missing (macOS / no ROCm)

davidberenstein1957:
## Summary

On systems where the `amdsmi` Python package is installed but ROCm / `libamd_smi.so` is not available (common on macOS), importing CodeCarbon could crash with `KeyError` during `import amdsmi` inside `gpu_amd.py`, breaking the CLI and any code path that loads GPU support.

This change broadens the existing `AttributeError` handler to also catch `OSError` and `KeyError`, sets `amdsmi = None` and `AMDSMI_AVAILABLE = False`, and logs a clear warning so AMD GPU metrics are skipped instead of failing import.

## Test plan

- [ ] On macOS (or any host without ROCm) with `amdsmi` installed: `codecarbon --help` and `python -c "import codecarbon"` succeed with a warning, no traceback.
- [ ] On a ROCm system with working `amdsmi`: AMD GPU metrics still work as before.

### DIFF
--- a/codecarbon/core/gpu_amd.py
+++ b/codecarbon/core/gpu_amd.py
@@ -28,12 +28,14 @@ except ImportError:
             "Please install amdsmi to get GPU metrics."
         )
     AMDSMI_AVAILABLE = False
-except (AttributeError, OSError, KeyError) as error:
+except (AttributeError, OSError, KeyError) as e:
     amdsmi = None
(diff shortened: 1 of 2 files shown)


benoit-cty (approved):

benoit-cty:
@copilot can you improve coverage by updating the tests ?

benoit-cty (approved):

PR #1203, opened 2026-05-19
feat: add FastAPI middleware for per-request emissions tracking

davidberenstein1957:
## Summary

- Adds optional `codecarbon[fastapi]` integration: middleware to track CO₂ emissions per HTTP request on FastAPI/Starlette apps
- Measurement runs **after** the response is sent, so clients are not blocked on hardware sampling
- Includes `create_codecarbon_lifespan` for a shared app-level tracker, route include/exclude filters, optional `save_to_api`, and opt-in `response_headers`
- Rebased on **v3.2.9** (latest release)
- Measured on a real HF embedder workload ([`paraphrase-MiniLM-L3-v2`](https://huggingface.co/sentence-transformers/paraphrase-MiniLM-L3-v2)): default deferred middleware ~**+4 ms** per request; sync headers cost more. See [docs/how-to/fastapi.md](docs/how-to/fastapi.md).

## Test plan

- [x] `uv run pytest tests/integrations/ -v`
- [ ] Manual smoke: `uv run --extra fastapi uvicorn examples.fastapi_middleware:app --reload` then `curl localhost:8000/predict`

## Usage

```python
from fastapi import FastAPI
from codecarbon.integrations.fastapi import add_codecarbon_middleware, create_codecarbon_lifespan

app = FastAPI(lifespan=create_codecarbon_lifespan(app, project_name="my-api"))
add_codecarbon_middleware(app)
```

Full guide: [docs/how-to/fastapi.md](docs/how-to/fastapi.md)


benoit-cty:
Hello, thanks for this. There is a problem with your branch : there are many changes that are already merged. Can you do a rebase ?

davidberenstein1957:
@benoit-cty I have reached out to some people at FastAPI, if they would be interested in a quick review :) For visibility, we could also consider deploying it as a standalone integration, but let's see if people like it.

codecarbon/integrations/fastapi/_headers.py
```diff
@@ -0,0 +1,119 @@
...
+
+from starlette.requests import Request
+from starlette.responses import Response
+
+from codecarbon.output_methods.emissions_data import EmissionsData
+
+HeaderConfig = Union[bool, str, Sequence[str], Mapping[str, str], None]
+HeaderFormatter = Callable[[EmissionsData, Request], Mapping[str, str]]
+
+FIELD_UNITS: dict[str, str] = {
```
SaboniAmine: Can this be transformed in an Enum ? I have the feeling it could be leveraged elsewhere for labels and it could be maintained from the core package, not only the fastapi integration. Thoughts ?

codecarbon/integrations/fastapi/_headers.py
```diff
@@ -0,0 +1,119 @@
...
+    "ram_power": "w",
+    "cpu_utilization_percent": "percent",
+    "gpu_utilization_percent": "percent",
+    "ram_utilization_percent": "percent",
+    "ram_used_gb": "gb",
+    "pue": "ratio",
+    "wue": "l-per-kwh",
+}
+
+HEADER_PRESETS: dict[str, dict[str, str]] = {
```
SaboniAmine: Same, a collection of Enums here would make sense no ?

codecarbon/integrations/fastapi/_headers.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def _auto_header_name(field: str) -> str:
+    unit = FIELD_UNITS.get(field, "")
+    title = "-".join(part.capitalize() for part in field.split("_"))
+    suffix = f"-{unit}" if unit else ""
+    return f"X-CodeCarbon-{title}{suffix}"
+
+
+def resolve_header_mapping(config: HeaderConfig) -> dict[str, str]:
```
SaboniAmine: Pretty sure this defensive method could be converted in a way more straightforward (and clear) one, if the unit tests demonstrate resilience to each corner case :)

codecarbon/integrations/fastapi/_routing.py
```diff
@@ -0,0 +1,135 @@
...
+        "/redoc",
+        "/openapi.json",
+        "/health",
+        "/healthz",
+        "/ready",
+        "/live",
+    }
+)
+
+HTTP_METHODS = frozenset(
```
SaboniAmine: Same, this could benefit from being an Enum

codecarbon/integrations/fastapi/_routing.py
```diff
@@ -0,0 +1,135 @@
...
+    return any(
+        matches_include(pattern, endpoint_key, endpoint_path) for pattern in include
+    )
+
+
+def build_task_name(
+    request: "Request",
+    formatter: Callable[["Request"], str] | None = None,
+) -> str:
+    """Derive a stable label like ``GET /items/{item_id}`` for task-scoped tracking.
```
SaboniAmine: This could lead to task override if resource is being accessed by 2 parallel or sequential requests. Maybe we could add some random here if the goal of this method is to provide a name for the codecarbon task concept

codecarbon/integrations/fastapi/middleware.py
```diff
@@ -0,0 +1,235 @@
+"""FastAPI/Starlette middleware for per-request emissions tracking."""
+
+from __future__ import annotations
+
+import asyncio
+from collections.abc import Awaitable, Callable, Iterable
+from typing import Any
+
+try:
```
SaboniAmine: This defensive import isn't needed, Starlette is installed with fastapi as a requirement for this plugin to work, you can assume that it will be present when end user will execute this code.

codecarbon/integrations/fastapi/middleware.py
```diff
@@ -0,0 +1,235 @@
...
+        tracker = EmissionsTracker(
+            project_name=self.project_name, **self.tracker_kwargs
+        )
+        tracker.start()
+        return tracker
+
+    async def _start_request_tracker(self) -> EmissionsTracker:
+        return await asyncio.to_thread(self._create_and_start_tracker)
+
+    async def _stop_request_tracker(
```
SaboniAmine: Stopping the tracker asynchronously will have an impact on measured value. Indeed, reading the value before emitting the OutputData object is made synchronously. Lets discuss this live I faced the same issue when dealing with the vLLM integration

codecarbon/integrations/fastapi/middleware.py
```diff
@@ -0,0 +1,235 @@
...
+        app: Any,
+        *,
+        project_name: str = "codecarbon-fastapi",
+        tracking_mode: str = "request",
+        include: Iterable[str] | None = None,
+        exclude: Iterable[str] | None = None,
+        response_headers: HeaderConfig | None = None,
+        include_emissions_header: bool = False,
+        header_formatter: HeaderFormatter | None = None,
+        task_name_formatter: Callable[[Request], str] | None = None,
```
SaboniAmine: This is a good idea!

docs/how-to/fastapi.md
```diff
@@ -0,0 +1,191 @@
...
+## `tracking_mode`: `request` vs `app`
+
+| Mode | Behavior |
+|------|-----------|
+| **`request`** (default) | Creates a short-lived `EmissionsTracker` per HTTP request. Safe under concurrency; each request is isolated. |
+| **`app`** | Reuses one tracker on `app.state` and uses `start_task` / `stop_task` per request (with an asyncio lock). Lower overhead; measurements for concurrent requests are serialized. |
+
+Use **`request`** unless you have measured a need for a shared tracker. For production APIs, prefer **`app`** mode with a lifespan handler and `save_to_file=False` to avoid per-request tracker startup cost.
+
+## Performance
```
SaboniAmine: Could we have a benchmark vs raw FastAPI, and with another telemetry tool ? 
Yes I'm thinking about Logfire but could be any other package injected as a middleware

examples/fastapi_middleware.py
```diff
@@ -0,0 +1,44 @@
...
+)
+
+_tracker_kwargs = {
+    "save_to_file": False,
+    "save_to_api": False,
+}
+
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
```
SaboniAmine: I'm not sure about explicit lifespan explicit declaration, could it be merged with the add_codecarbon_middleware method ?
My understanding is that the developer implementing the webserver leveraging the FastAPI would like to keep ownership of this lifespan function, and I'm not sure how to stack multiple constraints with this formalism. Do we have examples about how it's managed for similar libraries ?

SaboniAmine (commented):
Thanks David this a great PR! Left a few questions, and saw that you have already prepared a benchmark script, do you have any numbers / graph already computed to share ?

davidberenstein1957:
@inimaz This is ready for review. Rebased on v3.2.9 and benchmarked on a real HF embedder workload — middleware adds about 3 ms per request (~24 ms → ~27 ms). Details in the updated PR description and [docs/how-to/fastapi.md](https://github.com/mlco2/codecarbon/blob/feat/add-fastapi-middleware/docs/how-to/fastapi.md).

codecarbon/integrations/fastapi/_headers.py
davidberenstein1957: Good call — moved to `EmissionMetricField` / `HeaderPreset` in `codecarbon/core/emission_fields.py` so these labels can be reused outside the FastAPI integration. `_headers.py` now re-exports backward-compatible `FIELD_UNITS` / `HEADER_PRESETS` dicts.

codecarbon/integrations/fastapi/_headers.py
davidberenstein1957: Done — `HeaderPreset` enum + `preset_header_mapping()` live in core; FastAPI keeps the same string preset names (`emissions`, `default`, `energy`, `power`, `full`).

codecarbon/integrations/fastapi/_headers.py
davidberenstein1957: Refactored with `match` type dispatch + a single `_preset_mapping()` helper. Existing header unit tests still cover the corner cases (bool/None/presets/list/mapping/unknown).

codecarbon/integrations/fastapi/_routing.py
davidberenstein1957: Agreed — added `HttpMethod` in `codecarbon/core/emission_fields.py`; routing now uses `HttpMethod.__members__` for pattern parsing.

codecarbon/integrations/fastapi/_routing.py
davidberenstein1957: Good catch. We kept route labels stable (`GET /predict`) for logs/dashboard readability, and uniqueness lives at the measurement layer: with `create_codecarbon_lifespan`, `mark_http_request_start` → `_resolve_http_task_name()` appends a UUID when the name is already active (`HttpRequestBaseline`). Added a concurrent same-route test + docs note under `task_name_formatter`.

codecarbon/integrations/fastapi/middleware.py
davidberenstein1957: Removed the defensive Starlette import from `middleware.py`. The install hint now lives once at package import in `codecarbon.integrations.fastapi.__init__`.

codecarbon/integrations/fastapi/middleware.py
davidberenstein1957: Agreed this matters (same class of issue as vLLM). Current design: response is sent first, then finalize runs on a dedicated tracker **worker thread**; `_measure_power_and_energy()` is synchronous on that worker before `EmissionsData` / `on_request_complete`. Documented under “Measurement model” in `docs/how-to/fastapi.md` and added a regression test that measure precedes the callback. Happy to sync live if you want to align patterns with the vLLM integration.

codecarbon/integrations/fastapi/middleware.py
davidberenstein1957: Thanks — kept as-is.

docs/how-to/fastapi.md
davidberenstein1957: Added. Docs Performance table now includes raw FastAPI baseline, empty ASGI middleware, Logfire instrumentation (`send_to_logfire=False`), and CodeCarbon. Benchmark script supports `--with-logfire` (noop + Logfire scenarios). On the HF embedder workload CodeCarbon default is still ~3 ms overhead (~24 → ~27 ms), in the same ballpark as Logfire.

examples/fastapi_middleware.py
davidberenstein1957: We intentionally did **not** merge lifespan into `add_codecarbon_middleware` — FastAPI allows only one lifespan handler, and apps need to own DB/cache startup. Added `compose_lifespans()` so stacking is a one-liner, plus docs (“Combining with other lifespans”) and a commented stacked example in `examples/fastapi_middleware.py`.

davidberenstein1957:
@SaboniAmine Thanks again for the thorough review — I've addressed the inline threads in the latest pushes:

- **Core enums** for emission fields / header presets / HTTP methods (`codecarbon/core/emission_fields.py`)
- **Simpler** `resolve_header_mapping` + package-level Starlette import guard
- **Concurrent task naming** documented/tested (stable route labels + UUID uniqueness on the lifespan path)
- **Measurement model** docs + regression test (sync measure on tracker worker before callback)
- **`compose_lifespans()`** for stacking with user-owned startup/shutdown
- **Benchmarks** vs raw FastAPI + Logfire (`--with-logfire`); ~3 ms CodeCarbon overhead on the HF embedder workload

Local `uv run task test-package`: **552 passed**. Ready for another look when you have a moment.

davidberenstein1957:
Ponytail cleanup on this PR:

- **Removed** unused `_headers.py` / core `emission_fields` enums (never wired to middleware)
- **Added** opt-in `response_headers=` (sync measure → `X-CodeCarbon-*`; costs client latency)
- **Added** `include_background_tasks` (default `True`; set `False` to finalize at end-of-body)
- **Docs:** BackgroundTasks are included by default; WebSockets stay unsupported (no implementation)

See `docs/how-to/fastapi.md` Limitations + config section.

davidberenstein1957:
Added a **sync `response_headers`** row to the Performance section (and `--with-headers` on the benchmark script):

| Mode | Rough client latency (mocked 20 ms measure) |
|------|-----------------------------------------------|
| Deferred default | ~+3 ms (embedder table) |
| `response_headers=True`, c=1 | ~55 ms (~+25 ms vs ~30 ms baseline) |
| `response_headers=True`, c=4 | ~95 ms (tracker worker serializes sync measures) |

Reproduce: `uv run --extra fastapi python scripts/benchmark_fastapi_middleware.py --quick --with-headers`

davidberenstein1957:
@inimaz @SaboniAmine — ready for another look when you have time.

This commit finishes the HTTP perf work:
- REQUEST lane for `mark_http_request_start` (no thread-pool hop)
- Cached cloud metadata + emissions template on HTTP finalize
- Docs updated with **live** HF embedder benchmarks only (`--realistic --with-headers`)

Deferred middleware stays in the same ballpark as baseline on the live benchmark (~30–32 ms vs ~42 ms baseline mean on Darwin arm64).

benoit-cty:

## Verdict: 🔧 Request changes

Splitting each sampling window's energy across the requests in flight, weighted by overlap, is a nice design, and the attribution unit tests (sum invariant, concurrency) are good. But the PR has drifted: the branch was rewritten into a single commit (e293e0cd), and the description and earlier review threads describe code that no longer exists: `create_codecarbon_lifespan`, `compose_lifespans`, `response_headers`, `save_to_api`, `include_background_tasks`, the `codecarbon[fastapi]` extra, `docs/how-to/fastapi.md`, and the benchmark script. It also **conflicts with master** in `codecarbon/emissions_tracker.py`.

**Must fix:**
1. **The documented usage crashes at startup.**
   - `docs/how-to/examples.md` calls `add_codecarbon_middleware(app, …)` inside `lifespan`. Starlette builds the middleware stack before lifespan runs, so this raises `RuntimeError: Cannot add middleware after an application has started` (`middleware.py:118`). I reproduced this with FastAPI 0.141.1 / Starlette 1.6.0.
   - Fix: document adding the middleware at module level, with the tracker started and stopped in `lifespan`. The middleware could look up the tracker lazily from `app.state`. Please add a test for exactly the documented pattern.
2. **Memory grows without limit.** `attribution.py` `begin()` adds to `_in_flight`, and entries are removed only by `on_window()` (a tracker sample) or `close()`. If the tracker was never started or has stopped, and nothing calls `close()` (nothing does so automatically), every request leaks one `_InFlight`. Don't record requests when the tracker isn't running, or evict finished requests when no window is pending, and call `close()` on lifespan shutdown.
3. **Heavy work and a race on the scheduler thread.**
   - For each finished request, in each window, `_resolved` calls `tracker.http_request_emissions()` → `_prepare_emissions_data()` → `_update_emissions()` (`emissions_tracker.py:1016`, `middleware.py:91`). That runs inside `_measure_power_and_energy` on the scheduler thread. At high request rates, that is thousands of full `EmissionsData` builds per window, which delays sampling.
   - It also mutates `_total_emissions` / `_last_energy_covered` **without a lock**, racing with `flush()` / `stop()` on the main thread, so a delta can be counted twice.
   - Fix: compute each window's carbon intensity once, and derive request emissions as `energy_share × intensity` without touching tracker totals.
   - `_notify_energy_window_observers` also iterates `_window_observers` while `close()` can remove from it on another thread. Iterate over a copy.
4. **No optional extra.** FastAPI and httpx are only in the `dev` group (`pyproject.toml:98-99`), and `middleware.py:8` imports starlette at the top of the module with no hint. Please add a `codecarbon[fastapi]` extra, and raise an `ImportError` that tells users to install it.

**Tests:**
5. The single end-to-end test relies on `time.sleep` with a 0.5 s window and a real `EmissionsTracker` that does an online geo lookup, so it is likely to be flaky. Please use an offline tracker, or mocks. Add tests for `close()`, a tracker that was never started, the app raising (the status stays 500), and the lifespan pattern from point 1.

**Housekeeping:**
6. Please rebase, rewrite the description, and re-request review. @SaboniAmine's earlier threads (enums, lifespan, measuring an async stop) were answered against code that has since been removed.

**Nits:**
- `add_codecarbon_middleware` defines a subclass on every call just to get the instance onto `app.state`.
- It reads the private `tracker._total_energy`.
- The default `log_request` logs at INFO on every request; DEBUG would be better.
- `http_request_emissions()` puts a web-framework concept into the core tracker. It would sit better in the integration module.

davidberenstein1957:
Made the changes in 6819b504: module-level middleware with lifespan tracker, no leak, per-window intensity, fastapi extra, offline tests. Merged master (not rebased, to avoid a force-push) and rewrote the description.
Not done: the attributor still reads the private `tracker._total_energy` / `_start_time`; a public accessor felt like more core API than this PR needs.

codecarbon/integrations/fastapi/middleware.py
davidberenstein1957: Update in 6819b504: brought back a guarded import, because Starlette is now only in the new `codecarbon[fastapi]` extra and the error tells you to install it.

examples/fastapi_middleware.py
davidberenstein1957: Update in 6819b504: the middleware is now added at module level and finds the tracker on `app.state.codecarbon_tracker`. The lifespan only starts and stops the tracker, and pending requests are settled on shutdown automatically.

PR #1209, opened 2026-05-20
fix(cli): warn when auth validation falls back to API

### LINKED ISSUE #1208: CLI auth falls back to API on auth-server network errors
### Problem

`codecarbon/cli/auth.py` currently treats any `requests.RequestException` while validating the local access token as a successful validation path:

- `_validate_access_token()` returns `True` if the discovery document or JWKS fetch fails due to a network error.
- `get_access_token()` then reuses the stored access token and lets the API perform the final decision.

This is intentional for offline / flaky-network support, but it is worth documenting because it creates a clear caveat: when the auth server cannot be reached, the CLI does not locally prove that the token is still valid. A stale token, a provider migration, or a revoked token may still be accepted by the CLI until the API rejects the request.

### Why this matters

The behavior is easy to misread as local token validation. In practice, the CLI is using a best-effort check and then deferring trust to the server. That is acceptable for availability, but it should be called out explicitly so users and maintainers understand that:

- the CLI may proceed while offline without proving token freshness;
- failure to reach the auth server is not equivalent to token valid;
- security-sensitive changes in the auth flow should preserve this distinction.

### Suggested follow-up

- Document the fallback behavior in the CLI auth code and/or user-facing docs.
- Consider logging a warning when the CLI skips local validation because the auth server is unreachable.
- If stricter behavior is ever desired, add a separate mode for must validate locally versus best-effort offline fallback.

### Reference

Observed while reviewing PR #1199, which migrates JOSE validation from Authlib to Joserfc without changing this fallback path.

puneetdixit200:
## Description
Log a warning when local CLI access-token validation is skipped because the auth server cannot be reached. This preserves the existing offline/flaky-network fallback while making it explicit that token freshness was not proven locally and the API will make the final authorization decision.

## Related Issue
Fixes #1208

## Motivation and Context
The existing fallback is useful for availability, but silent success makes the path look like a completed local validation. A warning keeps the behavior unchanged while surfacing the caveat from the issue.

## How Has This Been Tested?
Red before implementation:

```sh
uv run --group dev python -m pytest -q tests/cli/test_cli_auth.py::TestAuthMethods::test_validate_access_token_network_error_returns_true
```

Green after implementation:

```sh
uv run --group dev python -m pytest -q tests/cli/test_cli_auth.py::TestAuthMethods::test_validate_access_token_network_error_returns_true
uv run --group dev python -m pytest -q tests/cli/test_cli_auth.py tests/cli/test_cli.py tests/cli/test_cli_main.py
uv run --group dev black --check codecarbon/cli/auth.py tests/cli/test_cli_auth.py
uv run --group dev ruff check codecarbon/cli/auth.py tests/cli/test_cli_auth.py
git diff HEAD^ --check
```

N/A

## Types of changes

## AI Usage Disclosure

OpenAI GPT-5 generated the small code/test change and PR text under explicit human-directed issue-fix constraints. I reviewed the touched auth path, preserved existing fallback behavior, and verified the change with the commands listed above.

- [x] 🟠 AI-generated: Car analogy : the car drive by itself, you are inside and give instructions.

## Checklist:

### DIFF
--- a/codecarbon/cli/auth.py
+++ b/codecarbon/cli/auth.py
@@ -6,6 +6,7 @@ PKCE), credential storage, JWKS validation, and transparent refresh.
 """
 
 import json
+import logging
 import os
 import webbrowser
 from http.server import BaseHTTPRequestHandler, HTTPServer
@@ -31,6 +32,7 @@ AUTH_SERVER_WELL_KNOWN = os.environ.get(
 _REDIRECT_PORT = 8090
 _REDIRECT_URI = f"http://localhost:{_REDIRECT_PORT}/callback"
 _CREDENTIALS_FILE = Path("./credentials.json")
+logger = logging.getLogger(__name__)
 
 
 # ── OAuth callback server ───────────────────────────────────────────
@@ -112,7 +114,12 @@ def _validate_access_token(access_token: str) -> bool:
         claims = jose_jwt.decode(access_token, keyset)
         claims.validate()
         return True
-    except requests.RequestException:
+    except requests.RequestException as exc:
+        logger.warning(
+            "Skipping local access token validation because the auth server is "
+            "unreachable; the API will validate the token. Error: %s",
+            exc,
+        )
         return True  # Can't reach auth server — let the API handle it
     except Exception:
         return False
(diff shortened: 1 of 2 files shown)


puneetdixit200:
Rebased this branch onto the current `master` head.

Local verification after the rebase:

```text
UV_CACHE_DIR=/private/tmp/codecarbon-1209-uv-cache uv run python -m pytest tests/cli/test_cli_auth.py
# 17 passed, 1 warning

git diff --check upstream/master...HEAD
# passed
```

GitHub currently reports no check rows for this branch yet.

inimaz (approved):
Nice addition thanks @puneetdixit200 !

PR #1213, opened 2026-05-28
fix(api): raise HTTPError instead of returning None on API errors

### LINKED ISSUE #820: API client to throw error instead of None

MD-Mushfiqur123:
Resolves #820

Changes:
- Replaced `_log_error()` + `return None` pattern with `r.raise_for_status()` in all ApiClient methods
- Removed unused `_log_error` method
- Removed None-check in `check_organization_exists` (method now raises on HTTP errors)
- `add_emission` and `_create_run` already have try/except wrappers so they continue to handle errors gracefully (logging + returning False/None)
- CodeCarbonAPIOutput (used by EmissionsTracker) already wraps `add_emission` in try/except
- Updated tests to expect `requests.HTTPError` instead of None returns


benoit-cty:
Thanks for your contribution. Can you improve the test coverage ?

PR #1250, opened 2026-06-21
chore: docker compose runs without auth

inimaz:
## Description
So far it was hard to test in local the webapp because of the authentication mechanism. The goal of this is to provide a simple way to deploy in local the server without auth. This makes it simple for devs to test the frontend + backend + db in one shot.

# Try it

```
docker compose up -d --build
```
Then go to `localhost:3000` and you should be able to test everything with just the auth mocked.

## AI Usage Disclosure

### DIFF
--- a/carbonserver/carbonserver/api/routers/authenticate.py
+++ b/carbonserver/carbonserver/api/routers/authenticate.py
@@ -23,9 +23,7 @@ AUTHENTICATE_ROUTER_TAGS = ["Authenticate"]
 LOGGER = logging.getLogger(__name__)
 OAUTH_SCOPES = ["openid", "email", "profile"]
 SESSION_COOKIE_NAME = "user_session"
-DEFAULT_REDIRECT_URL = (
-    f"{(settings.frontend_url or 'http://localhost:3000').rstrip('/')}/home"
-)
+
 
 router = APIRouter()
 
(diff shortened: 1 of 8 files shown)


carbonserver/carbonserver/api/routers/authenticate.py
```diff
@@ -81,7 +95,7 @@ async def get_login(
     login and redirect to frontend app with token
     """
     if auth_provider is None:
-        raise HTTPException(status_code=501, detail="Authentication not configured")
+        return RedirectResponse(get_redirect_url(request))
```
SaboniAmine: CAN it be computed at startup of server?
Redirect url shouldn't change during app lifetime

carbonserver/carbonserver/api/routers/authenticate.py
```diff
@@ -133,7 +147,7 @@ async def logout(
     Logout user by clearing session and removing cookie
     """
     if auth_provider is None:
-        raise HTTPException(status_code=501, detail="Authentication not configured")
+        return RedirectResponse(get_redirect_url(request))
```
SaboniAmine: Same

SaboniAmine (commented):
Thanks Inigo! This will be very useful!!

carbonserver/carbonserver/api/routers/authenticate.py
inimaz: Good catch! We don't even need to read the redirect from the request, changing it here [da71fbd](https://github.com/mlco2/codecarbon/pull/1250/commits/da71fbd393346fe74fdb236d182d5bf03fa603ab)

SaboniAmine (approved):

PR #1253, opened 2026-06-21
perf: defer API run creation until first emission upload (3/4)

davidberenstein1957:
## Summary

Part **3/4** of the tracker performance stack. **Depends on #1251 → #1252.**

cc @inimaz — small, isolated API write-path change from the #1246 split.

Defers the `POST /runs` API call until the first emission is actually uploaded, so online trackers with API output enabled no longer block on run creation during startup:

- `CodeCarbonAPIOutput` uses `create_run_automatically=False`
- `_ensure_api_run()` creates the run lazily on first `live_out` / `out` call

## Benchmarks (measured locally, offline Mac ARM, 2026-06-21)

Mocked geo/cloud/API network; measures `EmissionsTracker(save_to_api=True)` cold `start()`:

| Metric | #1251 + #1252 | **+ this PR** | Notes |
|--------|---------------|---------------|-------|
| API tracker `start()` avg | 407 ms | **369 ms** | ~9% faster — avoids blocking `POST /runs` at startup |
| Tracker `__init__` / warm lifecycle | unchanged | unchanged | No change expected — run creation deferred to first upload |

## Stack

1. #1251 — lazy initialization & import slimming
2. #1252 — hardware detection cache & warm-path reuse
3. **This PR** — lazy API run creation
4. #1254 — CLI monitor startup

## CI / quality

- [x] `pre-commit run` on all changed files — passed
- [x] `pytest tests/output_methods/test_http.py` — 9 passed
- [x] Full stacked suite (#1254 tip) — 535 passed locally

## Test plan

- [ ] Reviewer: online tracker with API output still uploads emissions correctly

## Replaces

Split from #1246.

PR #1252, opened 2026-06-21
perf: cache hardware detection and optimize warm-path reuse (2/4)

davidberenstein1957:
## Summary

Part **2/4** of the tracker performance stack. **Depends on #1251.**

cc @inimaz — this is the hardware-cache layer of the #1246 split.

Adds process-level hardware reuse and faster detection so repeat tracker runs in the same Python process skip repeated probing:

- New `hardware_cache.py` with `HardwareKind` enum and per-process setup cache
- Reuse CPU/GPU/RAM detection across tracker instances (`get_or_run_setup`)
- Cache probe results via `@lru_cache` (Power Gadget, PowerMetrics, NVIDIA, ROCm)
- Platform-aware CPU backend order (Mac ARM prefers fast `cpu_load` before PowerMetrics)
- Parallel CPU/GPU setup with `ThreadPoolExecutor`
- Global `cpu_percent` prime once per process for `cpu_load` mode
- PowerMetrics sudo check timeout (3 s)
- FAQ note on warm hardware reuse within one process

## Benchmarks (measured locally, offline Mac ARM, 2026-06-21)

Cumulative with #1251; fresh subprocess for cold metrics.

| Metric | master | #1251 only | **#1251 + this PR** | Δ vs master |
|--------|--------|------------|---------------------|-------------|
| Tracker `__init__` p50 | 190 ms | 1.9 ms | **1.2 ms** | ~99% faster |
| Cold lifecycle p50 (`init→start→stop`) | 1,685 ms | 1,767 ms | **408 ms** | **~76% faster (~4×)** |
| Warm lifecycle best (same process) | 1,565 ms | 1,561 ms | **4.8 ms** | **~325× faster** |
| CLI monitor subprocess p50 | 850 ms | 786 ms | 694 ms | ~18% faster |

## Stack

1. #1251 — lazy initialization & import slimming
2. **This PR** — hardware detection cache & warm-path reuse
3. #1253 — lazy API run creation
4. #1254 — CLI monitor startup

## CI / quality

- [x] `pre-commit run` on all changed files — passed
- [x] `pytest tests/test_hardware_cache.py tests/test_resource_tracker.py tests/test_cpu.py tests/test_powermetrics.py tests/test_gpu.py` — 106 passed
- [x] Full stacked suite (#1254 tip) — 535 passed locally

## Test plan

- [ ] Reviewer: second tracker in same process should reuse cached hardware, not re-probe

## Replaces

Split from #1246.

PR #1251, opened 2026-06-21
perf: defer tracker initialization and slim import path (1/4)

davidberenstein1957:
## Summary

Part **1/4** of the tracker performance stack (replaces #1246).

cc @inimaz — split per your review feedback into smaller, reviewable PRs.

Defers expensive work until it is actually needed so tracker construction and `import codecarbon` stay fast:

- Lazy-load static reference data in `DataSource` (CSV/JSON on first access, not at import)
- Defer hardware probing, geo validation, and emissions engine until first use
- Lazy-import output handlers and geography modules
- Skip the 1 Hz power monitor when `output_methods=[]`
- Avoid redundant measurement on `stop()` when a sample was just taken (`_measure_power_and_energy_if_stale`)
- Lazy-import `cpuinfo`, `arrow`, and direct `output_methods` imports

## Benchmarks (measured locally, offline Mac ARM, 2026-06-21)

Fresh Python subprocess per sample; baseline = `master`.

| Metric | master | After this PR | Δ |
|--------|--------|---------------|---|
| Tracker `__init__` p50 | 190 ms | **1.9 ms** | **~99% faster (~100×)** |
| Cold lifecycle p50 (`init→start→stop`) | 1,685 ms | 1,767 ms | ~same — hardware still probed on first start |
| Warm lifecycle best (same process) | 1,565 ms | 1,561 ms | ~same — cache lands in #1252 |
| CLI monitor subprocess p50 | 850 ms | 786 ms | ~8% faster (full CLI win in #1254) |

**What this PR delivers:** near-instant tracker construction. The big cold `start()` win comes in #1252.

## Stack

1. **This PR** — lazy initialization & import slimming
2. #1252 — hardware detection cache & warm-path reuse
3. #1253 — lazy API run creation
4. #1254 — CLI monitor startup

## CI / quality

- [x] `pre-commit run` on all changed files — passed (autoflake, isort, black, flake8)
- [x] Targeted tests — 45 passed
- [x] Full stacked suite (#1254 tip) — 535 passed locally

## Test plan

- [ ] Reviewer: offline tracker with `output_methods=[]` still measures on start/stop

## Replaces

Split from #1246.

PR #1235, opened 2026-06-11
perf(file): read only header row instead of entire CSV in has_valid_headers

vishali-mp:
## Description
Optimized FileOutput.has_valid_headers() to read only the CSV header row instead of parsing the entire file into memory.

**Before**: csv.DictReader read all rows into dicts (list(csv_reader)) just to extract column names from the first row's keys.
**After**: csv.reader reads just the first line via next(reader) and compares sorted headers directly — no unnecessary I/O or memory allocation for data rows.

Also optimized the append path in out() to use pd.DataFrame.to_csv(mode="a") instead of reading the full CSV, concatenating, and rewriting.

## Related Issue

https://github.com/mlco2/codecarbon/issues/1220

## Motivation and Context
On large CSV files, has_valid_headers() was O(n) in both time and memory for every out() call. This made append-mode writes progressively slower as the file grew. The fix makes header validation O(1) and append-mode writes O(1) per row.

## How Has This Been Tested?
All 22 existing tests in tests/output_methods/test_file.py pass, along with the full test suite (420 passed, 2 skipped).

## Types of changes
What types of changes does your code introduce? Put an `x` in all the boxes that apply:
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## AI Usage Disclosure

Please refer to [docs/how-to/ai-policy.md](https://docs.codecarbon.io/latest/how-to/ai_policy/) for detailed guidelines on how to disclose AI usage in your PR. Accurately completing this section is mandatory.

- [ ] 🟥 AI-vibecoded: You cannot explain the logic. Car analogy : the car drive by itself, you are outside it and just tell it where to go.
- [ ] 🟠 AI-generated: Car analogy : the car drive by itself, you are inside and give instructions.
- [x] ⭐ AI-assisted. Car analogy : you drive the car, AI help you find your way.
- [ ] ♻️ No AI used. Car analogy : you drive the car.

## Checklist:
Go over all the following points, and put an `x` in all the boxes that apply.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **[docs/how-to/contributing.md](https://docs.codecarbon.io/latest/how-to/contributing/)** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

[...]

PR #1261, opened 2026-06-29
release: v3.2.9

inimaz:
## Description
Create new release of the package


docs/how-to/contributing.md
```diff
@@ -317,8 +317,6 @@ git push --force-with-lease
 - Create a PR bumping the version with `uv run bumpver update --patch`. For a release candidate, use `uv run bumpver update --set-version 3.0.0_rc1`.
 - Run `uv run python .github/pyproject_versions.py -c` to check version consistancy.
 - Update the dependencies with `uv sync --upgrade`
-- Update the API dependencies with `cd carbonserver && uv sync --upgrade && cd ..`
-- Export API requirements with `uv run task carbonserver-api-requirements`
```
benoit-cty, line 321: Can you explain why you remove these lines ?

benoit-cty (commented):
Thanks Inigo,
What would be included in this release ?

benoit-cty:
> Thanks Inigo, What would be included in this release ?

Found it:

* perf: defer API run creation until first emission upload (3/4) (#1253) @davidberenstein1957
* perf: cache hardware detection and optimize warm-path reuse (2/4) (#1252) @davidberenstein1957
* fix: requirements-api.txt tested in ci (#1258) @inimaz
* chore: docker compose runs without auth (#1250) @inimaz
* perf: defer tracker initialization and slim import path (1/4) (#1251) @davidberenstein1957
* fix(UI): align date picker month arrows (#1233) @inimaz
* perf(file): read only header row instead of entire CSV in has_valid_headers (#1235) @vishali-mp
* fix(API): runs sums excludes the filtered runs (#1229) @inimaz
* fix: error when hover emissions chart (#1230) @inimaz

docs/how-to/contributing.md
inimaz: Because there is no need to update the `carbonserver` anymore. When we release we just update the package and client. Now they have 2 separated pyproject.toml and they do not depend to each other. 

We did this on purpose, so dependencies of the package do not affect dependencies on the server and viceversa. 
See https://github.com/mlco2/codecarbon/blob/master/pyproject.toml and https://github.com/mlco2/codecarbon/blob/master/carbonserver/pyproject.toml

inimaz:
Nice thanks @benoit-cty for the comment with what is included!

benoit-cty (approved):
Sorry for the delay

PR #1029, opened 2026-01-18
chore: run pip-audit on ci

inimaz:
## Description
Test to see if `pip-audit` can flag issues on the dependencies of the `uv.lock`.

PR #1283, opened 2026-07-31
ci: add non-blocking PySentry dependency audit

nyudenkov:
## Description
Adds `.github/workflows/pysentry.yml` — a non-blocking dependency vulnerability audit using [PySentry](https://github.com/nyudenkov/pysentry). It parses `uv.lock` natively (no resolver, no dev-dependency added to `pyproject.toml`), uploads results as SARIF to the Security tab, and cross-checks three advisory databases (PyPA, PyPI, OSV). The job is `continue-on-error: true`, so it's advisory and never blocks a PR. The action and `checkout` are pinned by SHA to match your pinned-dependencies setup.

## Related Issue
Follow-up to #1029 (`pip-audit` doesn't flag `uv.lock` dependencies).

## Motivation and Context
#1029 found that `pip-audit` alone doesn't read `uv.lock`, so lockfile dependencies went unaudited. PySentry reads `uv.lock` directly, which covers that gap without changing project dependencies. Added as a separate, non-blocking job so it surfaces findings in the Security tab without gating CI.

## How Has This Been Tested?
YAML validated locally. The workflow uses PySentry's official pinned GitHub Action; SARIF upload targets Code Scanning, which is already enabled here via CodeQL. Full behavior is visible on the first run of this PR (Actions run summary + Security tab).

N/A

## Types of changes

## AI Usage Disclosure

## Checklist:


nyudenkov:
@inimaz I also can add scan to the pre-commit if u want it :)

inimaz:
Thanks a lot for this @nyudenkov ! Can you add the pre-commit indeed?
Does this action upload to the security tab? I was not able to see it. Thanks again!

nyudenkov:
@inimaz hi!

done, and made it a bit nicer while at it:
- bumped the action to v0.5.0
- added a shared .pysentry.toml - it's the single source of truth for both this ci workflow and the pre-commit hook, so local and ci scans stay identical. it sets fail_on = "high" for production deps and loosens dev/doc groups to critical-only so tooling doesn't create noise
- added the pre-commit hook (pysentry/pysentry-pre-commit@v0.5.0, on pre-push since it resolves the whole dependency tree)

on the security tab: that's a fork-pr limitation, not a bug. v0.5.0 skips the upload on pull_request by default and prints findings in the job summary instead. the security-tab upload happens on the `push: master` run once merged

PR #1286, opened 2026-08-02
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `Medium` | **File**: `carbonserver/carbonserver/api/infra/database/database_manager.py:L15`

The Database class initializes the SQLAlchemy engine with `echo=True`. This causes the engine to log all SQL statements and parameters to the console/logs. In a production environment, this can lead to sensitive data exposure (e.g., user data, tokens, or PII being included in query parameters) and significantly increases log volume, potentially degrading system performance.

## Solution

Remove `echo=True` or make it configurable via an environment variable so it defaults to `False` in production environments.

## Changes

- `carbonserver/carbonserver/api/infra/database/database_manager.py` (modified)
- `carbonserver/carbonserver/api/routers/users.py` (modified)

### DIFF
--- a/carbonserver/carbonserver/api/infra/database/database_manager.py
+++ b/carbonserver/carbonserver/api/infra/database/database_manager.py
@@ -33,7 +33,7 @@ class Database:
         try:
             yield session
             session.commit()
-            logger.debug("Database session completed successfully")
+            logger.info("Database session committed successfully")
         except exc.IntegrityError as e:
             session.rollback()
             logger.error(
--- a/carbonserver/carbonserver/api/infra/database/database_manager.py
+++ b/carbonserver/carbonserver/api/infra/database/database_manager.py
@@ -12,7 +12,7 @@ Base = declarative_base()
 
 class Database:
     def __init__(self, db_url: str) -> None:
-        logger.info("Initializing database connection")
+        logger.info("Initializing database connection", extra={"db_url": db_url})
         self._engine = create_engine(db_url, echo=False, pool_pre_ping=True)
         self._session_factory = orm.scoped_session(
             orm.sessionmaker(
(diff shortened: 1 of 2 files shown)


SaboniAmine (approved):
Thanks for your contribution!
The /users/{user_id} endpoint being not used anymore, we'll simply remove it in a follow up PR, thanks for identifying the issue!

carbonserver/carbonserver/api/infra/database/database_manager.py
```diff
@@ -12,7 +12,7 @@
 
 class Database:
     def __init__(self, db_url: str) -> None:
-        self._engine = create_engine(db_url, echo=True, pool_pre_ping=True)
+        self._engine = create_engine(db_url, echo=False, pool_pre_ping=True)
```
SaboniAmine, line 16: @inimaz this change will make the debug easier on the prod instance but we should think of adding more functional logs to cover main data paths

carbonserver/carbonserver/api/routers/users.py
```diff
@@ -1,6 +1,7 @@
 from dependency_injector.wiring import Provide, inject
-from fastapi import APIRouter, Depends, status
+from fastapi import APIRouter, Depends, HTTPException, status
 
+from carbonserver.api.dependencies import MandatoryUserWithAuthDependency
```
inimaz: This ` carbonserver.api.dependencies` does not exist
```suggestion
from carbonserver.api.services.auth_service import MandatoryUserWithAuthDependency
```

carbonserver/carbonserver/api/infra/database/database_manager.py
inimaz: Sounds good!
If needed also we can create an env var (e.g. `VERBOSE_DB_LOGS`) in the future to activate/deactivate those logs without committing code.

PR #1288, opened 2026-08-02
fix(api): raise on unsuccessful API responses

### LINKED ISSUE #820: API client to throw error instead of None

PSR94:
## Summary
Make the CodeCarbon API client raise `requests.exceptions.HTTPError` when the API returns an unexpected HTTP status, while preserving the existing tracker-facing output handling that catches upload failures.

## Why this helps
Callers can now distinguish API failures through standard `requests` exceptions instead of receiving `None` or `False` from failed HTTP responses. This supports clearer error handling while keeping emissions uploads from breaking tracker execution.

## Changes made
- Added a shared checked request helper in `ApiClient` that logs API response details and calls `raise_for_status()`.
- Updated API client methods to use the shared helper for expected HTTP status validation.
- Let unsuccessful emission uploads and run creation HTTP responses raise `HTTPError`.
- Updated API client tests to assert the new error contract.

## Testing
- `uv run pytest tests/test_api_call.py tests/output_methods/test_http.py -q`
- `uv run ruff check codecarbon/core/api_client.py tests/test_api_call.py tests/output_methods/test_http.py`
- `uv run black --check --diff codecarbon/core/api_client.py tests/test_api_call.py tests/output_methods/test_http.py`
- `CODECARBON_ALLOW_MULTIPLE_RUNS=True uv run pytest --ignore=tests/test_viz_data.py -q -m 'not integ_test' tests/`

Closes #820

### DIFF
--- a/codecarbon/core/api_client.py
+++ b/codecarbon/core/api_client.py
@@ -70,6 +70,14 @@ class ApiClient:  # (AsyncClient)
             headers["Authorization"] = f"Bearer {self.access_token}"
         return headers
 
+    def _request(self, method, url, payload=None, expected_status=200):
+        headers = self._get_headers()
+        response = method(url=url, json=payload, timeout=2, headers=headers)
+        if response.status_code != expected_status:
+            self._log_error(url, payload or {}, response)
+            response.raise_for_status()
+        return response
+
     def set_access_token(self, token: str):
         """This method sets the access token to be used for the API.
         Args:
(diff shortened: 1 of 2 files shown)


codecarbon/core/api_client.py
```diff
@@ -70,6 +70,14 @@ def _get_headers(self):
             headers["Authorization"] = f"Bearer {self.access_token}"
         return headers
 
+    def _request(self, method, url, payload=None, expected_status=200):
```
inimaz, line 73: I like this helper!

inimaz (commented):
Thanks for the PR @PSR94 . Unfortunately as you might have been in the issue, it is not enough with throwing the error, one has to go into the places where this client has been called (cli and emissionsTracker) and do a try-catch around them.

There is another PR related to this here --> https://github.com/mlco2/codecarbon/pull/1277.
I see 2 options either you go and try-catch them or you wait until that PR does it and once merged resolve the conflicts and use your `_request` function. What do you think?

codecarbon/core/api_client.py
benoit-cty: Indeed it elegant. I will do this in #1277 and will credit you @PSR94 in the release note.

I prefer to merge #1277 as it better address the problem as a whole.

PR #1314, opened 2026-08-12
fix: keep None columns in CSV append

### LINKED ISSUE #1304: CSV append drops all-None columns, shifting every subsequent row
Appending a run to an existing `emissions.csv` silently drops any column whose value is `None`, and because the append is headerless the remaining values shift left. Every row after the first is corrupt for users whose runs have `None` fields — which is the common case: `longitude`/`latitude` are `None` for offline runs, `region` is `None` without an explicit region, and `gpu_count`/`gpu_model` are `None` on CPU-only machines.

## Reproduction

```python
from codecarbon.output_methods.emissions_data import EmissionsData
from codecarbon.output_methods.file import FileOutput
import pandas as pd, tempfile

d = tempfile.mkdtemp()
e = EmissionsData(..., gpu_count=None, gpu_model=None, longitude=None, latitude=None,
                  ram_total_size=16, tracking_mode="machine")
f = FileOutput("t.csv", d, on_csv_write="append")
f.out(e, None); f.out(e, None)

print([len(l.split(",")) for l in open(f.save_file_path)])   # [38, 38, 33]
pd.read_csv(f.save_file_path)[["ram_total_size", "tracking_mode"]]
#   row 0: 16.0, "machine"
#   row 1:  0.0, "N"        <- shifted
```

Equivalently, two successive `OfflineEmissionsTracker` runs into the same file.

## Root cause

`codecarbon/output_methods/file.py:104`

```python
elif self.on_csv_write == "append":
    new_df = new_df.dropna(axis=1, how="all")
    new_df.to_csv(self.save_file_path, mode="a", header=False, index=False)
```

`new_df` is always a single row, so `dropna(axis=1, how="all")` drops every column that happens to be `None` for this run. The workaround is legitimate in `task_out` (`file.py:143`), where it suppresses a pandas `FutureWarning` on a `concat`; there is no `concat` in the append branch, and column identity there is positional.

The corruption is never detected: `has_valid_headers` (`file.py:64-70`) only inspects the first line, so no backup is triggered, and `pd.read_csv` pads the short row instead of raising.

## Expected vs actual

Expected: every appended row has the same field count as the header, with empty fields for `None`.
Actual: rows with `None` values are shorter, and all values to the right of a dropped column are misread as the wrong field.

`tests/output_methods/test_file.py:200` writes four such rows today and passes — it asserts only the row count and header, neither of which the shift disturbs.

davidberenstein1957:
Closes #1304

## What changed

Removed the `dropna(axis=1, how="all")` from the `"append"` branch of `FileOutput.out()` (`codecarbon/output_methods/file.py:104`).

## Why

`new_df` in that branch is always a single row, so `dropna(axis=1, how="all")` really means "drop every column that is None for this run". The append is headerless, so column identity is positional — dropping a column shifts every value after it one place to the left, and `pd.read_csv` pads the short row instead of raising. `has_valid_headers` only reads the first line, so nothing detects it.

This hits ordinary runs: `longitude`/`latitude` are `None` for offline tracking, `region` is `None` without an explicit region, and `gpu_count`/`gpu_model` are `None` on CPU-only machines.

The `dropna` is legitimate in `task_out`, where it suppresses a pandas `FutureWarning` on a `concat`; it was not applicable here. Left `task_out` untouched.

## How it was verified

New test `test_file_output_out_append_keeps_none_columns_aligned` in `tests/output_methods/test_file.py`. It appends two rows built from an `EmissionsData` with `gpu_count`, `gpu_model`, `longitude`, `latitude` and `region` set to `None`, then asserts every line of the raw file has the same field count and that `ram_total_size` / `tracking_mode` read back correctly.

- Without the fix: fails with `{38, 33}` field counts.
- With the fix: `uv run pytest tests/output_methods/test_file.py -q` → 23 passed.

The existing `test_file_output_out_append_no_gpu_consistent_columns` passed before this change because it only asserts row count and header, neither of which the shift disturbs.

Rows now carry empty fields where the column previously vanished, matching the header-carrying first row. No API or config change.

### DIFF
--- a/codecarbon/output_methods/file.py
+++ b/codecarbon/output_methods/file.py
@@ -102,7 +102,8 @@ class FileOutput(BaseOutput):
         if not file_exists:
             new_df.to_csv(self.save_file_path, index=False)
         elif self.on_csv_write == "append":
-            new_df = new_df.dropna(axis=1, how="all")
+            # Never drop all-NA columns here: the append is headerless, so column
+            # identity is positional and a missing column shifts every value after it.
             new_df.to_csv(self.save_file_path, mode="a", header=False, index=False)
         else:
             df = pd.read_csv(self.save_file_path)
--- a/tests/output_methods/test_file.py
+++ b/tests/output_methods/test_file.py
@@ -266,6 +266,41 @@ class TestFileOutput(unittest.TestCase):
         self.assertIn("gpu_count", df.columns)
         self.assertIn("gpu_model", df.columns)
 
+    def test_file_output_out_append_keeps_none_columns_aligned(self):
+        """Regression test: appended rows must have the same number of fields as the
+        header, even when some values are None.
+
+        The bug: dropna(axis=1, how="all") on the single-row new_df dropped every
+        None column.  The append is headerless, so the surviving values shifted left
+        and were read back under the wrong column names.
+        """
+        none_data = EmissionsData(
+            **{
+                **self.emissions_data.values,
+                "gpu_count": None,
+                "gpu_model": None,
+                "longitude": None,
+                "latitude": None,
+                "region": None,
+            }
+        )
+
+        file_output = FileOutput("test.csv", self.temp_dir, on_csv_write="append")
+        file_output.out(none_data, None)
+        file_output.out(none_data, None)
+
+        with open(file_output.save_file_path) as csv_file:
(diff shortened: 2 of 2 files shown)


inimaz (approved):
Thanks @davidberenstein1957 ! LGTM

PR #1319, opened 2026-08-12
fix: integrate GPU power when energy counter is absent

### LINKED ISSUE #1316: GPUs without an NVML energy counter report 0 kWh
## Problem

On GPUs where NVML does not implement `nvmlDeviceGetTotalEnergyConsumption` (pre-Volta cards such as the GTX 1080, Tesla P100 or K80, and many virtualised/vGPU setups), CodeCarbon reports `gpu_energy = 0.0` and `gpu_power = 0.0` for the entire run. The GPU is still detected and listed in the metadata, so a GPU-bound training run silently reports only its CPU + RAM consumption. The same applies to AMD devices whose amdsmi payload carries no energy accumulator.

## Reproduction

Run any tracker on such a GPU, or simulate it by making `pynvml.nvmlDeviceGetTotalEnergyConsumption` raise `NVMLError`: `AllGPUDevices().get_delta(Time.from_seconds(3600))` returns `power_usage = 0 W` and `delta_energy_consumption = 0 kWh` on every device, even though `nvmlDeviceGetPowerUsage` reports a non-zero draw.

## Root cause

`NvidiaGPUDevice._get_total_energy_consumption` returns `None` when NVML raises (`codecarbon/core/gpu_nvidia.py:55`). `GPUDevice._get_energy_kwh` then returns `self.last_energy` unchanged (`codecarbon/core/gpu_device.py:44`), so in `GPUDevice.delta` (`codecarbon/core/gpu_device.py:51`) `energy == last_energy`, giving `energy_delta = 0` and `Power.from_energies_and_delay(...) = 0 W`. In other words "no new reading" is treated as "no energy consumed". Unlike `CPU.measure_power_and_energy`, the GPU path has no power-integration fallback, even though `_get_power_usage()` works on every card.

## Expected vs actual

Expected: when the cumulative counter is unavailable, integrate the instantaneous power draw over the measurement interval (the approximation the CPU path already uses) and warn once per device.

Actual: zero energy and zero power for the whole run, with only a per-sample `logger.warning("Failed to retrieve gpu total energy consumption")` that never says the resulting number is zero.

davidberenstein1957:
## What

`GPUDevice.delta` now branches on the backend returning `None` for the cumulative energy counter. When it does, it computes `Power.from_watts(self._get_power_usage())` and `Energy.from_power_and_time(...)` instead of comparing two identical counter readings, and logs a single warning per device so users know which method produced their numbers.

The per-sample `logger.warning` in `NvidiaGPUDevice._get_total_energy_consumption` is demoted to `debug`, since the once-per-device warning now carries the signal.

## Why

On GPUs where NVML does not implement `nvmlDeviceGetTotalEnergyConsumption` (pre-Volta cards, many virtualised GPUs) the previous code returned `last_energy` unchanged, yielding `energy_delta = 0` and `power = 0` for the entire run — a GPU-bound job silently reported only CPU + RAM. The fallback lives in the shared `GPUDevice`, so the AMD backend (which returns `None` when no energy accumulator is present) is fixed by the same change.

Power integration is less accurate than the counter, but it is the same approximation the CPU path already uses and is dramatically better than zero.

## How verified

`tests/test_gpu_nvidia.py::TestGpu::test_gpu_without_energy_counter_falls_back_to_power` patches the NVML energy query to raise and asserts 26 W / 29 W and 0.026 / 0.029 kWh over an hour. It fails on master (`assert 0.0 == 26`) and passes here. `tests/test_gpu.py`, `tests/test_gpu_nvidia.py` and `tests/test_gpu_amd.py` all pass (51 tests).

Closes #1316

### DIFF
--- a/codecarbon/core/gpu_device.py
+++ b/codecarbon/core/gpu_device.py
@@ -1,7 +1,8 @@
 from dataclasses import dataclass, field
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 from codecarbon.core.units import Energy, Power, Time
+from codecarbon.external.logger import logger
 
 
 @dataclass
@@ -28,8 +29,9 @@ class GPUDevice:
     power: Power = field(default_factory=lambda: Power(0))
     # Energy consumed in kWh
     energy_delta: Energy = field(default_factory=lambda: Energy(0))
-    # Last energy reading in kWh
-    last_energy: Energy = field(default_factory=lambda: Energy(0))
+    # Last energy reading in kWh, None if the device has no energy counter
+    last_energy: Optional[Energy] = field(default_factory=lambda: Energy(0))
+    _warned_no_counter: bool = field(default=False, init=False)
 
     def start(self) -> None:
         self.last_energy = self._get_energy_kwh()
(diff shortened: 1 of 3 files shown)


benoit-cty:

**The bug is real.** On master, `_get_energy_kwh()` returns `last_energy` when NVML/AMD exposes no energy counter, so every delta is 0. The fallback is correct:
- `_get_power_usage()` already returns W on both backends (NVML mW/1000; AMD `average_socket_power`).
- It works per device, so a mix of GPUs with and without counters is fine.
- Recovering from `None` to a real counter re-baselines without double counting, and it's tested.
- A failing power read now reports 0 W instead of raising.

54 GPU tests pass on this PR merged with master.

**Nits / follow-ups:**
1. This is a rectangle rule over a single instantaneous sample taken at the end of the interval. Pre-Volta NVML power is instantaneous, so a bursty workload over the default 15 s interval can be noticeably off. A follow-up could average the samples the monitor already takes each second over the interval.
2. On AMD dual-GCD devices, `__post_init__` reads the counter before `_known_zero_energy_counter` is set, so the secondary GCD's first interval goes through the fallback. It's harmless because that GCD reports 0 W, but setting the flag first would be cleaner.
3. The description mentions 1 test; the diff adds 4. Please update it.
4. The branch is behind `master`; please update it.

davidberenstein1957:
Made the changes in 19613cef: merged master, set static details before the first energy read (AMD dual-GCD flag), and updated the description to list all 4 tests.
Not done: averaging the monitor's per-second power samples over the interval, left as the suggested follow-up.

PR #1321, opened 2026-08-12
fix: skip uncorrectable RAPL counter wraps

### LINKED ISSUE #1309: RAPL counter wrap yields negative energy when max_energy_range_uj is unreadable
When a RAPL counter wraps around and `max_energy_range_uj` could not be read, CodeCarbon records a large negative energy delta instead of skipping the sample. That value is summed straight into the CPU and total energy, so long-running sessions can report totals that go down over time.

## Reproduction

Point a `RAPLFile` at an energy file whose `max_path` does not exist (this is what happens when only `energy_uj` is readable, e.g. some container/VM RAPL passthroughs, MMIO domains, or a `chmod` scoped to `energy_uj`):

```python
write(energy_uj, "4000000000"); f = RAPLFile(...); f.start()
write(energy_uj, "10000");      f.delta(Time.from_seconds(10))
assert f.energy_delta.kWh < 0   # passes today, ~ -1.11e-3 kWh
```

## Root cause

`codecarbon/core/rapl.py:45` sets `max_energy_reading = 0` on any failure to read `max_energy_range_uj`, with a comment claiming wrap detection is then "effectively disabled". It is not — only the *correction* is disabled. `codecarbon/core/rapl.py:75-83` still takes the wrap branch, adds `0`, and computes `energy_delta = energy - last_energy`, a large negative number.

`codecarbon/core/cpu.py:597` (`_build_readable_domains`) gates domain acceptance on `energy_uj` only, so a domain with an unreadable max file is happily monitored. The negative delta then flows through `codecarbon/core/cpu.py:847` and `codecarbon/external/hardware.py:381-393` into the tracker totals with no sign check.

The same negative delta is produced by the transient-read fallback at `codecarbon/core/rapl.py:65`, which returns `0` on a read error and therefore trips the wrap branch on the next successful read.

The reported power still looks plausible, because `Power.from_energies_and_delay` takes `abs()` (`codecarbon/core/units.py:139`), which is what makes this hard to spot.

## Expected vs actual

- Expected: an uncorrectable backwards counter is dropped (zero delta) with a warning, as `codecarbon/core/windows_emi.py:567-574` already does for EMI.
- Actual: up to -1.19e-3 kWh subtracted from the totals per wrap (every ~hour at 1 W for a 2^32 uJ package domain, minutes under load). `_update_emissions` skips negative deltas (`codecarbon/emissions_tracker.py:957-976`), so `_last_energy_covered` also permanently desynchronises from `_total_energy`.

davidberenstein1957:
Closes #1309

## What changed

`RAPLFile.delta()` now re-checks the wrap correction: if the corrected reading is still below the previous one, the sample is dropped (zero energy delta, zero power) with a warning, instead of emitting a negative energy delta.

## Why

`max_energy_range_uj` is set to `0` whenever it cannot be read (`codecarbon/core/rapl.py:45`). The comment claims wrap detection is then disabled, but only the *correction* was disabled — the wrap branch still ran and produced `energy_delta = energy - last_energy`, a large negative value (up to -1.19e-3 kWh for a 2^32 uJ package domain). That flows unchecked into `_total_energy` / `_total_cpu_energy`, and the `abs()` in `Power.from_energies_and_delay` masks it in the power column.

The same guard also covers the other sources of a backwards counter: driver resets, suspend/resume, and the `_get_value()` transient-error fallback that returns `0`. It mirrors what `codecarbon/core/windows_emi.py:566-574` already does for EMI.

Dropping a sample loses at most one interval of genuine energy — strictly better than subtracting an hour's worth. Systems where `max_energy_range_uj` is readable are unaffected.

## Verification

`tests/test_rapl_permissions.py`:
- `test_rapl_wraparound_without_max_skips_sample` — fails on master (`energy_delta.kWh` is about -1.11e-3), passes here.
- `test_rapl_wraparound_with_max_is_corrected` — pins the existing wrap correction against regression.

`uv run pytest tests/test_rapl_permissions.py -q` → 2 passed, 2 skipped (the pre-existing Linux-only cases). `black --check` clean; `ruff` reports only pre-existing findings in the touched files.

## Merge order

This should land **before** the `rapl_include_dram` change: DRAM domains have smaller max ranges and wrap far more often, which makes this defect fire much more frequently.

### DIFF
--- a/codecarbon/core/rapl.py
+++ b/codecarbon/core/rapl.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass, field
+from typing import Optional
 
 from codecarbon.core.units import Energy, Power, Time
 from codecarbon.external.logger import logger
@@ -16,10 +17,11 @@ class RAPLFile:
     energy_delta: Energy = field(default_factory=lambda: Energy(0))
     # Power based on reading
     power: Power = field(default_factory=lambda: Power(0))
-    # Last energy reading in kWh
-    last_energy: Energy = field(default_factory=lambda: Energy(0))
+    # Last energy reading in kWh, None if it could not be read
+    last_energy: Optional[Energy] = field(default_factory=lambda: Energy(0))
     # Max value energy can hold before it wraps
     max_energy_reading: Energy = field(default_factory=lambda: Energy(0))
+    _warned_uncorrectable_wrap: bool = field(default=False, init=False)
 
     def __post_init__(self):
         self.last_energy = self._get_value()
@@ -44,9 +46,9 @@ class RAPLFile:
                 )
             self.max_energy_reading = Energy.from_ujoules(0)
 
-    def _get_value(self) -> Energy:
+    def _get_value(self) -> Optional[Energy]:
         """
-        Reads the value in the file at the path
+        Reads the value in the file at the path, or None if it cannot be read.
         """
         try:
             with open(self.path, "r") as f:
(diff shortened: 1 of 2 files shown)


benoit-cty:

## Verdict: 🔧 Request changes (small)

The bug is real, and worse than the description says:
- When `max_energy_range_uj` can't be read, a wrap produces a negative delta.
- On master, `_get_value()` returns 0 on a transient read error. That injects a spurious `+max_energy_reading` (up to about 0.07 kWh), and the next sample then counts the whole since-boot counter as a single delta.

Returning `None` from `_get_value()` and skipping values that still go backwards after correction fixes both. This PR merges cleanly with #1344, and I'd suggest landing it first.

**Must fix:**
1. **Crash in `IntelRAPL.start()` once merged with master.**
   - Master's mirror-counter detection does `float(rapl_file.last_energy)` (`codecarbon/core/cpu.py` ~L893).
   - With this PR, that value is `None` when a counter can't be read at `start()`, giving `TypeError: float() argument must be ... not 'NoneType'` from `CPU.start()`. I reproduced this on the PR merged with master.
   - Fix: skip `None` values in that list comprehension, and add a test.
   - Please rebase on master and rerun the RAPL tests: CI was green on the old base, before the mirror dedup landed.

**Should fix:**
2. **The description overstates the scope.** The guard covers a driver reset or suspend/resume only when the max range is unknown. With a readable max, a counter reset still becomes `energy + max - last`, a large positive delta. Either reword, or add a plausibility bound (e.g. a delta above what TDP × interval allows gets skipped). A follow-up is fine.
3. **Mention the `_get_value()` → `None` change in the description.** It's the most impactful part of the PR.

**Nit:**
- The wrap formula omits the +1 µJ. It's negligible.

davidberenstein1957:
Made the changes in b5aaca1e: merged master, skipped unreadable (None) counters in the start() mirror detection with a test, and reworded the description (scope, plus the `_get_value()` -> `None` change).
Not done: the TDP-based plausibility bound for resets with a readable max (fine as a follow-up, as suggested) and the +1 µJ wrap nit (negligible).

PR #1324, opened 2026-08-12
fix: stop PeriodicScheduler from re-arming itself after stop()

### LINKED ISSUE #1308: PeriodicScheduler re-arms itself after stop(), so measurements can outlive tracker.stop()
## Problem

`PeriodicScheduler` can keep running (and can run its payload) after `stop()` has returned. In the tracker this means `_measure_power_and_energy` may execute after `EmissionsTracker.stop()` has already written the final row and exited the output handlers — producing an extra measurement, a post-run `live_out` push (Prometheus/API handlers), or an exception logged after "Done!".

## Reproduction

```python
import threading, time
from codecarbon.external.scheduler import PeriodicScheduler

calls = []
s = PeriodicScheduler(0.05, lambda: calls.append(1))
gate, orig = threading.Event(), s.start
# widen the (real, but narrow) window between the timer firing and _run reaching start()
s.start = lambda from_run=False: (from_run and gate.wait(2), orig(from_run=from_run))[1]
orig(from_run=False)
time.sleep(0.2)   # timer fired, _run is blocked before start()
s.stop()
gate.set()
time.sleep(0.4)
print(s._stopped, len(calls))   # -> False, and the count keeps climbing forever
```

## Root cause

`codecarbon/external/scheduler.py:26-42`. `_run()` calls `start(from_run=True)`, and the `from_run` flag deliberately bypasses the `_stopped` check, so a fired timer *always* arms its successor:

```python
if from_run or self._stopped:
    self._stopped = False
    self._timer = Timer(self.interval, self._run)
```

If `stop()` lands between the timer firing and `_run` acquiring the lock, `stop()` cancels an already-fired (uncancellable) timer and flips `_stopped = True`; `_run` then sets it back to `False` and arms a timer nobody holds a reference to — the scheduler becomes unstoppable. Separately, `stop()` reads `_stopped` outside the lock (`scheduler.py:48`), and `_run` never re-checks the flag before invoking the payload, so the payload runs to completion after `stop()` returned. `EmissionsTracker.stop()` sets `self._scheduler = None` immediately after (`emissions_tracker.py:909-914`), so that late payload mutates tracker state whose handlers have already been `exit()`ed (`emissions_tracker.py:937-938`).

## Expected vs actual

Expected: after `stop()` returns, no further payload invocations and no re-arming.
Actual: the reschedule path is not serialized against the stop path over `_stopped`, so both a stray timer and a late payload invocation can survive `stop()`.

davidberenstein1957:
## What changed

`PeriodicScheduler.start()` now checks `_stopped` on the reschedule path too (inside the lock), `_run()` re-checks the flag before invoking the payload, and `stop()` reads the flag under the lock.

## Why

`_run()` called `start(from_run=True)`, which deliberately bypassed the `_stopped` check, so a fired timer always armed a successor. A `stop()` landing between the timer firing and `_run()` reaching `start()` cancelled an already-fired (uncancellable) timer and was then silently undone — the scheduler kept firing forever. The payload also ran to completion after `stop()` returned, so `_measure_power_and_energy` could mutate tracker state and call `live_out` after `EmissionsTracker.stop()` had written the final row and exited the handlers.

The `elif not self._stopped: return` branch preserves today's "a second `start()` on a running scheduler is a no-op" semantics. The payload deliberately stays outside the lock, so `stop()` never blocks for a whole measurement (and must not be called from the payload itself).

## Verification

New `tests/test_scheduler.py`. `test_stop_during_timer_firing_does_not_rearm` fails on master (scheduler keeps firing, `_stopped` flips back to `False`) and passes here; the other three cover stop-mid-payload, double `start()` arming one timer, and `stop()` before `start()` / twice. `tests/test_emissions_tracker.py` passes unchanged.

Closes #1308


davidberenstein1957:
Closing in favour of #1339, which fixes this same re-arm bug plus the re-entrancy problem in the same file. Two open PRs rewriting scheduler.py is review burden for no extra benefit, so it's better to have one.

PR #1325, opened 2026-08-12
fix: send measurement time to the API

### LINKED ISSUE #1312: API emissions are timestamped at send time, not measurement time
When emissions are uploaded to the API, the `timestamp` stored server-side is the moment the HTTP payload was built, not the moment the measurement window ended. The measured timestamp is present in the payload and simply never read.

## Impact

A user running a tracker with `output_methods=["file", "api"]` gets two different timestamps for the same measurement: the CSV row is stamped when the measurement was finalised, the API row when the POST was assembled. The two datasets do not line up, so a CSV row cannot be matched to its API row by time, and a dashboard chart and a local CSV chart of the same run do not agree.

The same defect applies to the run record itself, so a run's start time is also the time its creation request was assembled rather than when tracking started.

## Reproduction

Run a tracker with both `output_methods=["file", "api"]` and compare the CSV `timestamp` column against the API rows for the same run — they do not agree.

## Root cause

`EmissionsData.timestamp` is set when the measurement is finalised, in `codecarbon/emissions_tracker.py:1047`:

```python
timestamp=datetime.now().strftime("%Y-%m-%dT%H:%M:%S"),
```

It survives into `HTTPOutput._emit`, which passes the whole dataclass down (`codecarbon/output_methods/http.py:65-68`). But `ApiClient.add_emission` discards `carbon_emission["timestamp"]` and generates a fresh one (`codecarbon/core/api_client.py:192-193`):

```python
emission = EmissionCreate(
    timestamp=get_datetime_with_timezone(),
```

There is a second site with the same problem: `ApiClient._create_run` stamps the run the same way at `codecarbon/core/api_client.py:234-235`.

`get_datetime_with_timezone()` (`codecarbon/core/api_client.py:25-28`) is `arrow.now()` — i.e. "now, at payload-build time".

## Expected vs actual

- **Expected:** the stored `timestamp` locates the measurement window in time, and matches the CSV row for the same measurement.
- **Actual:** it records when the POST was assembled.

Fixing this is also a prerequisite for any client-side batching or retry/spill-buffer work, which would otherwise misdate every replayed measurement by the full outage duration.

davidberenstein1957:
## What changed

`ApiClient.add_emission` now sends the timestamp carried by the emission payload instead of generating a new one at payload-build time. A small `get_measurement_timestamp` helper localises the naive `EmissionsData.timestamp` to an offset-aware ISO string, falls back to `get_datetime_with_timezone()` when no timestamp is present (hand-built dicts), and logs a warning rather than failing on an unparseable value.

## Why

`EmissionsData.timestamp` records when the measurement window ended (`codecarbon/emissions_tracker.py:1063`), survives all the way into `HTTPOutput._emit` (`codecarbon/output_methods/http.py:65-68`), and was then thrown away by `add_emission`, which stamped `get_datetime_with_timezone()` instead (`codecarbon/core/api_client.py:191`). Every other field in the payload describes the measurement window; the one field that dates it described the HTTP call. Users running both `file` and `api` output methods see the CSV and the API disagree.

The stored value is left offset-aware, so no existing rows shift and the CSV output format is unchanged.

This is also a prerequisite for any future client-side batching or retry/spill-buffer work: without it, every replayed or batched emission would be stored at its send time, turning a sub-second skew into an outage-length one.

Closes #1312

## How it was verified

`tests/test_api_call.py::TestApi::test_add_emission_keeps_measurement_timestamp` asserts the POSTed `timestamp` starts with the supplied measurement time and is offset-aware, and that the no-timestamp payload still yields a parseable aware ISO string. It fails on master and passes here. `uv run pytest tests/test_api_call.py -q` → 27 passed.

Note: `uv run task format` reformats a large number of unrelated files on current master, so only the two files touched here are included in this PR.

### DIFF
--- a/codecarbon/core/api_client.py
+++ b/codecarbon/core/api_client.py
@@ -8,7 +8,7 @@ TODO : use async call to API
 # from httpx import AsyncClient
 import dataclasses
 import json
-from datetime import timedelta, tzinfo
+from datetime import datetime, timedelta, tzinfo
 
 import requests
 
@@ -26,6 +26,24 @@ def get_datetime_with_timezone():
     import arrow
 
     return str(arrow.now().isoformat())
+
+
+def get_measurement_timestamp(timestamp=None):
+    """
+    Return the time the measurement was taken, as an offset-aware ISO string.
+
+    EmissionsData.timestamp is naive local time, so it is localised here.
+    Falls back to now for hand-built payloads without a timestamp.
+    """
+    if timestamp:
+        try:
+            return datetime.fromisoformat(timestamp).astimezone().isoformat()
+        except (TypeError, ValueError):
+            logger.warning(
+                f"ApiClient : could not parse emission timestamp {timestamp!r}, "
+                + "using current time."
+            )
+    return get_datetime_with_timezone()
 
 
 class ApiClient:  # (AsyncClient)
@@ -190,7 +208,7 @@ class ApiClient:  # (AsyncClient)
             )
             return False
         emission = EmissionCreate(
-            timestamp=get_datetime_with_timezone(),
+            timestamp=get_measurement_timestamp(carbon_emission.get("timestamp")),
             run_id=self.run_id,
             duration=int(carbon_emission["duration"]),
             emissions_sum=carbon_emission["emissions"],
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -1044,6 +1044,9 @@ class BaseEmissionsTracker(ABC):
         )
 
         total_emissions = EmissionsData(
+            # Naive local time. Consumers that need an offset-aware value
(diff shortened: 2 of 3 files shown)


davidberenstein1957:
Closing as a duplicate of #1341, which fixes the same bug — the API client stamping emissions with the send time instead of the measurement time.

#1341 is the more complete version: it has broader fallback coverage (missing, `None`, and unparseable timestamps), no per-call log noise, and a comment at `_create_run` explaining why stamping "now" is correct *there*, so nobody "fixes" it later.

The one thing this PR has that #1341 lacks is the producer-side comment noting that `.astimezone()` assumes the local zone because the value comes from `datetime.now()`. That is worth keeping, so it will be ported across.

Nothing wrong with the work here — the two were developed in parallel against the same bug report.

PR #1326, opened 2026-08-12
fix: resolve RAM default pid at init time

### LINKED ISSUE #1318: RAM's default pid is evaluated at import time, so forked workers track the parent process
## Problem

`RAM.__init__` uses `pid: int = psutil.Process().pid` as a default argument. Python evaluates default arguments once, when the module is first imported, so the default pid is frozen to the importing process for the life of the interpreter — including across `fork()`. A child process that constructs `RAM()` therefore records the *parent's* pid.

## Reproduction

```python
import os, multiprocessing as mp
from codecarbon.external.ram import RAM   # imported in the parent

def worker():
    print(os.getpid(), RAM(tracking_mode="process")._pid)  # these differ

mp.set_start_method("fork")
p = mp.Process(target=worker); p.start(); p.join()
```

Linux/fork only — with `spawn` the child re-imports and the default is recomputed.

## Root cause

- `codecarbon/external/ram.py:37` — `pid: int = psutil.Process().pid` (evaluated at import).
- `codecarbon/external/ram.py:58` — `self._pid = pid`.
- Neither real construction site passes a pid: `codecarbon/core/resource_tracker.py:37` and `codecarbon/core/hardware_cache.py:164`, so the stale default is always the one in use.

By comparison `CPU` resolves it correctly inside `__init__` (`codecarbon/external/hardware.py:216`).

## Expected vs actual

Expected: with `tracking_mode="process"`, RAM usage is measured for the process that created the tracker.

Actual, in a forked child:
- If the parent is alive, `process_memory_GB` (`ram.py:285-296`) measures the parent's RSS plus *all* its descendants — i.e. every sibling worker — so N workers each report roughly the whole pool.
- If the parent has exited, `psutil.Process(pid)` raises `NoSuchProcess`; `total_power` swallows it (`ram.py:337-339`) and RAM power silently reads 0 W for the rest of the run.

Machine tracking mode is unaffected (it never reads `_pid`).

## Fix

Default `pid` to `None` and resolve `os.getpid()` in `__init__`.

davidberenstein1957:
## What

`RAM.__init__` took `pid: int = psutil.Process().pid`. Python evaluates default arguments once at import, so the value was frozen to whichever process first imported `codecarbon.external.ram` — and survived `fork()`. The default is now `None`, resolved to `os.getpid()` inside `__init__` (matching how `CPU` already does it in `external/hardware.py`).

## Why

Neither construction site passes a pid (`core/resource_tracker.py:37`, `core/hardware_cache.py:164`), so the stale default was always in use. In a forked child with `tracking_mode="process"` this meant measuring the parent's RSS plus every sibling worker, or — if the parent had exited — `NoSuchProcess`, which `total_power` swallows into a silent 0 W for the rest of the run.

## Verification

New test `tests/test_ram.py::TestRAM::test_default_pid_is_resolved_in_forked_child` forks a child, constructs `RAM(tracking_mode="process")` there and asserts `_pid` is the child's own pid. It fails on master and passes with this change; `uv run pytest tests/test_ram.py` is green (21 passed).

Closes #1318

### DIFF
--- a/codecarbon/external/ram.py
+++ b/codecarbon/external/ram.py
@@ -1,4 +1,5 @@
 import math
+import os
 import re
 import subprocess
 from dataclasses import dataclass
@@ -34,7 +35,7 @@ class RAM(BaseHardware):
 
     def __init__(
         self,
-        pid: int = psutil.Process().pid,
+        pid: Optional[int] = None,
         children: bool = True,
         tracking_mode: str = "machine",
         force_ram_power: Optional[int] = None,
@@ -46,7 +47,7 @@ class RAM(BaseHardware):
 
         Args:
             pid (int, optional): Process id (with respect to which we'll look for
-                                 children). Defaults to psutil.Process().pid.
+                                 children). Defaults to the current process id.
             children (int, optional): Look for children of the process when computing
                                       total RAM used. Defaults to True.
             tracking_mode (str, optional): Whether to track "machine" or "process" RAM.
@@ -55,7 +56,7 @@ class RAM(BaseHardware):
                                            this value is used instead of estimating RAM power.
                                            Defaults to None.
         """
-        self._pid = pid
+        self._pid = os.getpid() if pid is None else pid
         self._children = children
         self._tracking_mode = tracking_mode
         self._force_ram_power = force_ram_power
--- a/tests/test_ram.py
+++ b/tests/test_ram.py
@@ -1,9 +1,11 @@
+import os
 import subprocess
 import unittest
 from textwrap import dedent
 from unittest import mock
 
 import numpy as np
+import pytest
 
 from codecarbon.external.ram import RAM, RAM_SLOT_POWER_X86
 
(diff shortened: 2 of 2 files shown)


benoit-cty:

**The bug is real.** `pid: int = psutil.Process().pid` runs once, when the module is imported. So a forked child in process mode measures its parent, and reports 0 W silently on `NoSuchProcess`. Resolving `os.getpid()` in `__init__` when `pid is None` is correct. The hardware cache rebuilds RAM from specs, so it can't hold on to a stale instance. 21 tests pass.

**Nits:**
1. The test forks inside multi-threaded pytest, which gives a `DeprecationWarning` on Python 3.12+ and can be flaky. Patching `os.getpid` (e.g. `mock.patch("os.getpid", return_value=12345)`) and asserting `RAM()._pid == 12345` would be simpler and deterministic.
2. If `RAM()` raises in the child, the parent currently fails with a confusing `int('')`. If you keep the fork, write an explicit sentinel or the exit status instead.
3. The branch is behind `master`; please update it before merge.

davidberenstein1957:
Made the changes in bb22b62e: swapped the fork test for a deterministic os.getpid mock (which also removes the int('') issue), and merged master in.

PR #1327, opened 2026-08-12
fix: propagate the API run id to the tracker

### LINKED ISSUE #1303: tracker.run_id stays None when OutputMethod.API is enabled, CSV rows record run_id="None"
## Problem

When `OutputMethod.API` is enabled, `EmissionsTracker.run_id` stays `None` for the whole run. Every emission record stringifies it, so all CSV rows are written with the literal string `"None"` instead of a real run id, and the data can no longer be joined with what the API stored under a real UUID.

## Reproduction

```python
from codecarbon.output import OutputMethod

tracker = EmissionsTracker(output_methods=[OutputMethod.CSV, OutputMethod.API], api_key="...")
print(tracker.run_id)  # None
tracker.start(); ...; tracker.stop()
# emissions.csv -> run_id column contains "None"
```

## Root cause

`emissions_tracker.py:648` takes the id eagerly from the API output handler:

```python
self.run_id = cc_api__out.run_id
```

but `CodeCarbonAPIOutput.__init__` builds its `ApiClient` with `create_run_automatically=False` (`output_methods/http.py:51-58`), so `ApiClient.run_id` is still the class attribute `None` (`core/api_client.py:36`). The run is created lazily later in `CodeCarbonAPIOutput._ensure_api_run` (`output_methods/http.py:60-63`), which updates the *handler's* id but never writes back to the tracker. `emissions_tracker.py:1049` then records `run_id=str(self.run_id)` → `"None"`.

## Expected vs actual

Expected: `tracker.run_id` is never `None`, and matches the API run id once the run exists.

Actual: it is `None`, with knock-on effects:

- `on_csv_write="update"` matches rows on `df.run_id == total.run_id` (`output_methods/file.py:109`); with every run writing `"None"`, a second run **overwrites the first run's row**, and a third logs "CSV contains more than 1 rows with current run ID".
- `FileOutput.task_out` builds per-task filenames as `emissions_<experiment>_<run_id>.csv` (`output_methods/file.py:134-137`), so every run's task data collides in `emissions_<experiment>_None.csv`.
- The BOAMPS output names its report `boamps_report_None.json` (`output_methods/boamps/output.py:155`) for the same reason.

davidberenstein1957:
## What

`EmissionsTracker.run_id` is now a property backed by a locally generated uuid that returns the API run id as soon as one exists, and the API run is created in `start()` rather than on the first emission.

## Why

With `OutputMethod.API` enabled, the tracker read `run_id` from `CodeCarbonAPIOutput` at construction time, but that handler builds its `ApiClient` with `create_run_automatically=False` and only creates the run lazily in `_ensure_api_run`. Nothing ever wrote the resulting id back, so `tracker.run_id` stayed `None` and `_prepare_emissions_data` recorded the literal string `"None"` in every row.

Consequences for users: CSV data could not be joined with the API data, `on_csv_write="update"` matched on `"None"` and overwrote the previous run's row, and per-task / BOAMPS filenames all collided on `..._None.csv`.

Creating the run in `start()` (instead of only fixing the property) makes sure a single run has a single id everywhere, including rows persisted before the first API call. Failures there are logged and fall back to the local uuid, so an unreachable API no longer prevents tracking.

## Verified

`tests/test_emissions_tracker.py::TestCarbonTracker::test_run_id_with_api_output_is_never_none` asserts `tracker.run_id` is not `None` and that the run id written to `emissions.csv` matches the API run id. It fails on `master` and passes with this change.

`uv run pytest tests/test_emissions_tracker.py tests/output_methods tests/test_tracking_inference.py tests/test_api_call.py -q` passes locally.

Closes #1303

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -267,6 +267,8 @@ class BaseEmissionsTracker(ABC):
             self._output_methods.append(OutputMethod.LOGFIRE)
 
     def _initialize_runtime_state(self) -> None:
+        self._api_output = None
+        self.run_id = uuid.uuid4()
         self._start_time: Optional[float] = None
         self._last_measured_time: float = time.perf_counter()
         self._total_energy: Energy = Energy.from_energy(kWh=0)
@@ -612,7 +614,6 @@ class BaseEmissionsTracker(ABC):
         methods = set(self._output_methods) if self._output_methods else set()
 
         if not methods and not self._emissions_endpoint:
-            self.run_id = uuid.uuid4()
             return
 
         from codecarbon.output_methods.boamps import BoAmpsOutput
@@ -645,10 +646,8 @@ class BaseEmissionsTracker(ABC):
                 api_key=api_key,
                 conf=self._conf,
             )
-            self.run_id = cc_api__out.run_id
+            self._api_output = cc_api__out
             self._output_handlers.append(cc_api__out)
-        else:
-            self.run_id = uuid.uuid4()
 
         if OutputMethod.PROMETHEUS in methods:
             self._output_handlers.append(
@@ -710,6 +709,15 @@ class BaseEmissionsTracker(ABC):
             return
 
         self._ensure_hardware_ready()
+
+        if self._api_output is not None:
+            # Create the run now, so every record carries the API run id.
+            try:
+                self._api_output._ensure_api_run()
+                self.run_id = self._api_output.run_id or self.run_id
+            except Exception as e:
+                logger.error(e, exc_info=True)
+
         self._last_measured_time = self._start_time = time.perf_counter()
 
         # Clear utilization history for fresh measurements
--- a/tests/test_emissions_tracker.py
+++ b/tests/test_emissions_tracker.py
@@ -356,6 +356,92 @@ class TestCarbonTracker(unittest.TestCase):
             )
         )
 
+    def test_run_id_with_api_output_is_never_none(
+        self,
+        mock_cli_setup,
+        mock_log_values,
+        mocked_get_gpu_details,
+        mocked_env_cloud_details,
+        mocked_get_gpu_utilization_list,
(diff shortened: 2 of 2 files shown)


benoit-cty:

## Verdict: ✅ Approve with one fix suggested

**The bug is real.** With API output, `tracker.run_id` stays `None`, so every CSV row gets the string `"None"`. The new tests fail on master. The API run is now created in `start()`, with a fallback to a locally generated id if that fails. Each tracker has its own API output, so multiple trackers are fine.

**Issue to fix: CSV and API ids can still differ** (medium)
- If the API is down at `start()`, `run_id` keeps the local id (`emissions_tracker.py` ~L712-719).
- Later, `_emit()` calls `_ensure_api_run()` (`output_methods/http.py` L60-63), which creates the API run once the API is back, with a **different** id from every CSV row. This happens silently, and defeats the goal of "a single id everywhere".
- Suggested fix: when the API run is eventually created, sync `self.run_id` from `_api_output.run_id`, e.g. in `_prepare_emissions_data`, and log once that earlier CSV rows used the local id. Or, if the API accepts a client-supplied run id, create the run with the local id.
- Please add a test: API fails at `start()`, recovers at the first `_emit()`, then assert the ids agree afterwards.

**Nits:**
1. `start()` now makes a blocking HTTP POST (2 s timeout) on the caller's thread. That's acceptable, but worth a line in the docs.
2. A `ConnectionError` is logged twice, once in `_create_run` and once in `start()`.
3. `run_id` read before `start()` differs from the one after. Document this or keep it consistent.
4. The description mentions a `run_id` property that isn't in the code; please update it.
5. The branch is behind `master`; please update it.

davidberenstein1957:
Made the changes in d53e2de9: run_id now syncs to the API run once it's created late (with a test for API down at start then recovering), the start() fallback logs once as a warning, docs mention the blocking call and the run_id switch, PR description fixed, and master merged in.

PR #1329, opened 2026-08-12
fix: restore signal handlers on lock release

### LINKED ISSUE #1310: Lock permanently replaces the host application's SIGINT/SIGTERM handlers
When `allow_multiple_runs=False`, `EmissionsTracker` constructs a `Lock`, and `Lock.__init__` installs its own `SIGINT`/`SIGTERM` handlers with `signal.signal(...)`. This only happens when the tracker is built on the main thread — `codecarbon/lock.py:32` guards it with `if threading.current_thread() is threading.main_thread():` — but on that path the return value (the previous handler) is discarded and nothing ever restores it.

Two user-visible consequences:

1. `KeyboardInterrupt` stops being raised. Python's default SIGINT handler is what turns Ctrl-C into `KeyboardInterrupt`; after a `Lock` exists, Ctrl-C raises `SystemExit(1)` from `Lock._handle_exit` instead, so `except KeyboardInterrupt:` blocks in the embedding application no longer run. CodeCarbon's own at `codecarbon/cli/monitor.py:95` is affected, and that is what terminates the tracked subprocess.
2. An application handler installed before the tracker is silently destroyed. A web server, trainer or job runner that registered a graceful-shutdown handler no longer receives SIGTERM.

## Reproduction

```python
import signal

signal.signal(signal.SIGTERM, my_graceful_shutdown)

tracker = EmissionsTracker(allow_multiple_runs=False)
# signal.getsignal(signal.SIGTERM) is now Lock._handle_exit;
# my_graceful_shutdown will never run.
```

## Root cause

`codecarbon/lock.py:34-35` overwrite both handlers without saving them:

```python
signal.signal(signal.SIGINT, self._handle_exit)
signal.signal(signal.SIGTERM, self._handle_exit)
```

`_handle_exit` (`lock.py:37-41`) releases the lock and unconditionally raises `SystemExit(1)` rather than delegating to whatever was there before, and `Lock.release()` (`lock.py:57-68`) only removes the lock file — it never restores signal state. The `Lock` is created at `codecarbon/emissions_tracker.py:170`.

## Expected vs actual

Expected: CodeCarbon releases its lock file on a signal and then lets the application's original handler (or the Python default, so `KeyboardInterrupt` is raised) take over; after `release()` the handlers are back to what they were.

Actual: the original handlers are lost for the lifetime of the process and every signal ends in `SystemExit(1)`.

A secondary, smaller issue in the same constructor: `atexit.register(self.release)` (`lock.py:26-28`) is never unregistered, so every `Lock` ever built stays reachable for the process lifetime.

davidberenstein1957:
Closes #1310

## What changed

`codecarbon/lock.py`:

- `Lock.__init__` now keeps the handlers returned by `signal.signal()` in `self._previous_handlers` instead of discarding them.
- `_handle_exit` releases the lock and then delegates to the handler it replaced: it calls a previous Python handler, or reproduces the default disposition with `os.kill(os.getpid(), signum)` when it was `SIG_DFL`, or does nothing for `SIG_IGN`. It no longer raises `SystemExit(1)` unconditionally.
- `release()` restores the saved handlers (only when the currently installed handler is still ours, so an application that registered its own afterwards is not clobbered) and unregisters the `atexit` hook.
- The `atexit` callback is stored once as `self._atexit_hook`, because `atexit.unregister()` will not match a freshly-created bound method.

## Why

`Lock` is constructed whenever `allow_multiple_runs=False`. Because it overwrote the process signal disposition permanently, restoring the default SIGINT handler never happened and `KeyboardInterrupt` stopped being raised at all — every `except KeyboardInterrupt:` in the embedding application became dead code, including CodeCarbons own in `codecarbon/cli/monitor.py:95`. Applications that had registered a graceful-shutdown SIGTERM handler lost it silently.

## Verification

Two new tests in `tests/test_lock.py` (`TestLockSignalHandlers`):

- `test_release_restores_previous_handlers` — a sentinel SIGTERM handler is installed, a `Lock` is built, and the sentinel (plus the original SIGINT handler) must be back after `release()`.
- `test_signal_is_forwarded_to_previous_handler` — raises a real SIGTERM and asserts the applications handler ran.

Both fail on `master` and pass with this change; `uv run pytest tests/test_lock.py -q` is green (7 passed). `uv run task format`/`lint` reformat the whole repository with the current tool versions, so only the two touched files were formatted and checked (`ruff`/`black` clean apart from pre-existing findings in `tests/test_lock.py`).

## Ordering note

The `stop()` idempotency fix should land **first**. Now that `_handle_exit` chains to the previous handler, the CLIs own SIGINT handler (`codecarbon/cli/main.py`) runs after the lock releases and calls `tracker.stop()`; on a session that already stopped the tracker that becomes a second `stop()`, which would turn a hijacked signal into a duplicate emissions row.

## Behaviour change

Previously any SIGINT/SIGTERM ended in `SystemExit(1)`. Now the process does whatever the application asked for. That is the intent of the fix, but anyone relying on CodeCarbon force-exiting on Ctrl-C will notice.

### DIFF
--- a/codecarbon/lock.py
+++ b/codecarbon/lock.py
@@ -27,18 +27,20 @@ class Lock:
             self.release
         )  # Ensure release() is called on unexpected exit of the user's python code
         # If there is more than one thread add a lock
-        self._thread_lock = threading.Lock()
-        # If the current thread is the main thread, register signal handlers
-        if threading.current_thread() is threading.main_thread():
-            # Register signal handlers to ensure lock release on interruption
-            signal.signal(signal.SIGINT, self._handle_exit)  # Ctrl+C
-            signal.signal(signal.SIGTERM, self._handle_exit)  # Termination signal
+        # Reentrant: _handle_exit -> release() can fire on a thread already holding it.
+        self._thread_lock = threading.RLock()
+        # Previous signal handlers, restored on release().
+        self._previous_handlers = {}
 
     def _handle_exit(self, signum, frame):
-        """Ensures the lock file is removed when the script is interrupted."""
-        logger.debug(f"Signal {signum} received. Releasing lock and exiting.")
-        self.release()
-        raise SystemExit(1)  # Exit gracefully to prevent further execution
+        """Releases the lock, then delegates to the handler we replaced."""
+        logger.debug(f"Signal {signum} received. Releasing lock.")
+        previous = self._previous_handlers.get(signum, signal.SIG_DFL)
+        self.release()  # also restores the previous handlers
+        if callable(previous):
+            return previous(signum, frame)
+        if previous == signal.SIG_DFL:
+            os.kill(os.getpid(), signum)
 
     def acquire(self):
         """Creates a lock file and ensures it's the only instance running."""
@@ -48,6 +50,14 @@ class Lock:
                 with open(LOCKFILE, "x") as _:
                     logger.debug(f"Lock file created. Path: {LOCKFILE}")
                     self._has_created_lock = True
+                # Only now that we own the lock file: a failed acquire() must not
+                # leave the host application's handlers hijacked for good, since
+                # release() is never reached on that path.
+                if threading.current_thread() is threading.main_thread():
+                    for sig in (signal.SIGINT, signal.SIGTERM):
+                        self._previous_handlers[sig] = signal.signal(
+                            sig, self._handle_exit
+                        )
             except FileExistsError:
                 logger.debug(
                     f"Lock file {LOCKFILE} already exists. This usually means another instance of codecarbon is running. You can safely delete it if you want or use allow_multiple_runs parameter to always bypass it."
@@ -55,12 +65,19 @@ class Lock:
                 raise
 
     def release(self):
-        """Removes the lock file on exit."""
+        """Removes the lock file and restores the signal handlers we replaced."""
         with self._thread_lock:
             logger.debug("Removing the lock")
+            while self._previous_handlers:
+                sig, handler = self._previous_handlers.popitem()
+                # Only restore if nobody installed another handler after us.
+                if signal.getsignal(sig) == self._handle_exit:
+                    signal.signal(sig, handler)
+            atexit.unregister(self.release)
             try:
                 # Remove the lock file only if it was created by this instance
                 if self._has_created_lock:
+                    self._has_created_lock = False
                     os.remove(LOCKFILE)
             except OSError as e:
                 logger.debug(f"Error: {e}")
--- a/tests/test_lock.py
+++ b/tests/test_lock.py
@@ -1,3 +1,6 @@
+import atexit
+import os
+import signal
 import threading
 import unittest
 from unittest.mock import mock_open, patch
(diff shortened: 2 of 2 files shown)


benoit-cty:

## Verdict: 🔧 Request changes

**The bug is real.** `Lock` permanently took over SIGINT/SIGTERM and always raised `SystemExit(1)`. The fix saves the previous handlers, restores them on release, forwards the signal to them, and uses an `RLock`.
- Nested trackers are fine: a second lock never acquires, so it installs no handlers.
- Handlers the user sets after `acquire()` are kept (the `getsignal` check).

**Must fix:**
1. **Calling `release()` off the main thread loses the final emissions and leaves the lock file behind** (`codecarbon/lock.py` L71-75).
   - `signal.signal()` only works on the main thread, and `release()` now calls it with no thread check, so it raises `ValueError`.
   - `stop()` calls `_lock.release()` before saving anything. `@suppress` swallows the exception, so the rest of `stop()` never runs.
   - I verified this: with `OfflineEmissionsTracker(allow_multiple_runs=False)` and `stop()` called from a worker thread, master writes the CSV and removes the lock. With this PR, `stop()` returns `None`, **no CSV is written, and the lock file stays** until the process exits.
   - Stopping from a callback or worker thread is a common pattern.
   - Fix: only restore handlers when `threading.current_thread() is threading.main_thread()`. Remove the lock file first, and wrap the restore in `try/except`. Please add a test that stops from a worker thread.
2. **SIGTERM now kills the process with no final write** (`lock.py` L42-43).
   - When the previous SIGTERM handler was `SIG_DFL`, the PR restores it and re-sends the signal with `os.kill`. The process then dies immediately: no `finally`, no `__exit__`, no `tracker.stop()`.
   - On master, `SystemExit(1)` let `with` / `@track_emissions` write the final emissions. I verified this: master exits with code 1 and `finally` runs; the PR exits with 143 and `finally` does not run.
   - On SLURM/Kubernetes shutdowns this loses the final row that users get today.
   - Suggested fix: forward only to *callable* previous handlers. When the previous handler is `SIG_DFL`, raise `SystemExit(128 + signum)` (or `KeyboardInterrupt` for SIGINT), so cleanup runs and the exit code stays conventional.

**Nits:**
- If the previous handler was installed from C code, `signal.getsignal()` returns `None`. `release()` then calls `signal.signal(sig, None)`, which raises `TypeError`, and `_handle_exit` swallows the signal. Treat `None` as `SIG_DFL`.
- The description refers to an `_atexit_hook` that isn't in the code, and the ordering note about stop idempotency is already covered by #1408. Please update it.

davidberenstein1957:
Made the changes in f1321785: worker-thread release and SIGTERM cleanup fixed.

- `release()` removes the lock file first and only restores handlers on the main thread (new worker-thread test).
- A default previous handler now raises `SystemExit(128 + signum)` / `KeyboardInterrupt` instead of `os.kill`, so `finally` and the final CSV row still happen (checked: exit 143, row written).
- A `None` previous handler is restored as `SIG_DFL`. Updated the description too.

PR #1330, opened 2026-08-12
fix: keep null coordinates when creating a run

### LINKED ISSUE #1317: API output silently fails for OfflineEmissionsTracker: round(None, 1) raises TypeError
When the API output method is used with `OfflineEmissionsTracker` (or any tracker whose geolocation could not be resolved to coordinates), no run is ever created and no emissions are ever sent. The log only shows `TypeError: type NoneType doesn't define __round__ method`, which points at rounding rather than at the missing coordinates.

## Reproduction

```python
tracker = OfflineEmissionsTracker(
    country_iso_code="FRA",
    output_methods=[OutputMethod.API],
    api_key="...",
)
tracker.start()
...
tracker.stop()
```

## Root cause

`codecarbon/core/api_client.py:245-246` builds the run payload with:

```python
longitude=round(self.conf.get("longitude", 0), 1),
latitude=round(self.conf.get("latitude", 0), 1),
```

`dict.get(key, default)` only applies the default when the key is *absent*. `emissions_tracker.py:384-386` writes both keys unconditionally once geo is resolved on private infra, and the offline tracker builds its `GeoMetadata` without coordinates (`emissions_tracker.py:1400-1409`, `external/geography.py:75-76`), so the keys exist with the value `None`. `round(None, 1)` then raises.

The initial run creation is skipped entirely today (`output_methods/http.py:56` passes `create_run_automatically=False`), so the only path is the lazy retry in `HTTPOutput._ensure_api_run` (`output_methods/http.py:60-63`), which runs after geo has been resolved. `_emit` swallows the exception via `logger.error`, so the failure is silent apart from the traceback.

## Expected vs actual

Expected: coordinates are omitted (both `RunBase.longitude/latitude` are `Optional[float]` in `core/schemas.py:50-51` and the DB column is nullable) and the run is created normally.

Actual: `TypeError` on every `out()`/`live_out()`, run never created, all emissions dropped.

Fix is to round only when the value is not `None`, keeping the null rather than coercing to `(0, 0)` (which would pin every offline run to the Gulf of Guinea on a map).

davidberenstein1957:
Closes #1317

## What changed

`ApiClient._create_run` used `round(self.conf.get("longitude", 0), 1)`. The `0` default only applies when the key is *absent*, but the tracker writes `longitude`/`latitude` into the conf unconditionally (`emissions_tracker.py:384-386`), and the offline tracker resolves them to `None`. `round(None, 1)` then raised `TypeError`, the run was never created, and `HTTPOutput._emit` swallowed the exception — so `OfflineEmissionsTracker` with `OutputMethod.API` sent no data at all.

Coordinates are now rounded only when present, and left as `None` otherwise. Both `RunBase.longitude/latitude` are `Optional[float]` and the DB column is nullable, so a null is preferred over coercing to `(0, 0)`, which would place every offline run in the Gulf of Guinea on a map.

## Verification

Two tests in `tests/test_api_call.py`:

- `test_create_run_keeps_unknown_coordinates_null` — fails on master with `TypeError: type NoneType doesn't define __round__ method`, passes with the fix.
- `test_create_run_rounds_coordinates` — regression guard on the privacy rounding (`-7.6174` -> `-7.6`).

`uv run pytest tests/test_api_call.py -q` -> 28 passed.

Note: `uv run task format` reformats ~120 unrelated files on the current ruff version, so it was reverted outside the two files touched here.


inimaz (approved):
Looks good thanks @davidberenstein1957 !

PR #1331, opened 2026-08-12
fix(cli): report actual emissions file path

### LINKED ISSUE #1322: codecarbon monitor prints a "Saved to" path that ignores output_dir
The post-run report printed by `codecarbon monitor -- <cmd>` resolves the emissions file path from the file *name* only, so whenever `output_dir` is not the current working directory the CLI tells the user to look in the wrong place.

## Reproduction

```bash
cd /home/user
CODECARBON_OUTPUT_DIR=/var/log/codecarbon codecarbon monitor -- ./train.sh
```

Prints:

```
   Saved to: /home/user/emissions.csv
```

while the CSV was actually written to `/var/log/codecarbon/emissions.csv`. Same happens with `output_dir` set in `.codecarbon.config`.

## Root cause

`codecarbon/cli/monitor.py:119-125` uses `tracker._conf["output_file"]` and calls `os.path.abspath()` on it. But `output_file` is only the basename (default `"emissions.csv"`, `codecarbon/emissions_tracker.py:567`); the directory lives in a separate key `output_dir` (`emissions_tracker.py:566`). The two are joined only inside `FileOutput` (`codecarbon/output_methods/file.py:49`). `os.path.abspath` therefore resolves the bare name against the CWD, which is correct only for the default `output_dir="."`.

`FileOutput` does log the correct absolute path at construction (`output_methods/file.py:50-52`), but the CLI defaults to `log_level="error"` (`cli/main.py`, forwarded at `cli/monitor.py:54`), so that INFO line is suppressed and the wrong path is the only thing the user sees.

## Expected vs actual

- Expected: the report shows the path the CSV was actually written to (and shows nothing when CSV output is disabled).
- Actual: `<cwd>/emissions.csv`, unconditionally, even when no file output handler exists.

Display-only defect; measurement and file writing are unaffected.

davidberenstein1957:
The `codecarbon monitor` post-run report resolved the emissions file path from `tracker._conf["output_file"]`, which is only the basename. `os.path.abspath` then resolved it against the current working directory, so any `output_dir` other than the default `"."` produced a path where the file is not. The CLI defaults to `log_level="error"`, which suppresses the correct INFO line `FileOutput` logs, so the wrong path was the only one the user saw.

## What changed

`codecarbon/cli/monitor.py` now asks the tracker's output handlers for their `save_file_path` instead of reconstructing the path from config keys. This keeps the CLI correct if the path logic changes, prints nothing when CSV output is disabled (previously a confidently wrong path), and handles multiple file handlers.

## Verification

Three new tests in `tests/cli/test_monitor.py` (`test_run_and_monitor_reports_output_dir`, `test_run_and_monitor_reports_relative_path_from_cwd`, `test_run_and_monitor_reports_no_path_without_file_output`). The first and third fail on master and pass with this change. `uv run pytest tests/cli/test_monitor.py -q` -> 10 passed.

Display-only change; no measurement or file-writing behaviour is affected.

Closes #1322

### DIFF
--- a/codecarbon/cli/monitor.py
+++ b/codecarbon/cli/monitor.py
@@ -116,13 +116,12 @@ def run_and_monitor(
         else:
             print("   Emissions: N/A")
 
-        # Show where the data was saved
-        if hasattr(tracker, "_conf") and "output_file" in tracker._conf:
-            output_path = tracker._conf["output_file"]
-            # Make it absolute if it's relative
-            if not os.path.isabs(output_path):
-                output_path = os.path.abspath(output_path)
-            print(f"   Saved to: {output_path}")
+        # Show where the data was saved, asking the output handlers themselves
+        # so the path stays right whatever `output_dir` / `output_file` are.
+        for handler in getattr(tracker, "_output_handlers", []):
+            save_file_path = getattr(handler, "save_file_path", None)
+            if save_file_path:
+                print(f"   Saved to: {os.path.abspath(save_file_path)}")
 
         print("   ⚠️  Note: Tracked the command process and its children")
         print("=" * 60)
--- a/tests/cli/test_monitor.py
+++ b/tests/cli/test_monitor.py
@@ -1,3 +1,4 @@
+import os
 from types import SimpleNamespace
 
 import pytest
@@ -11,6 +12,7 @@ class FakeTracker:
         self.kwargs = kwargs
         self.stopped = 0
         self._conf = {"output_file": "emissions.csv"}
+        self._output_handlers = [SimpleNamespace(save_file_path="emissions.csv")]
 
     def start(self):
         return None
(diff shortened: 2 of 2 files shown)


benoit-cty:

**The bug is real.** The CLI report used `abspath(_conf["output_file"])`, which ignores `output_dir`. The new tests fail on master. Reading `save_file_path` from each output handler, and printing no path when CSV output is off, is the right fix.

**Nits:**
1. BoAmps output also writes files but has no `save_file_path`, so its files are not reported. Consider giving it a `save_file_path`, or reporting every handler that has one.
2. The description lists `test_run_and_monitor_reports_relative_path_from_cwd`, which is not in the diff. Please update the description, or add the test.
3. The branch is behind `master`; please update it.

davidberenstein1957:
Made the changes in 869852b2: merged master, BoAmps reports now listed, added the relative-path test.

`BoAmpsOutput` sets `save_file_path` after writing its report, so the CLI summary picks it up with no special case.

PR #1332, opened 2026-08-12
fix: apply PUE to reported power too

### LINKED ISSUE #1320: PUE is applied to energy but not to reported power
When `pue` is set to anything other than the default `1.0`, the PUE factor is applied to the accumulated energy but not to the power values reported in the same row. The result is a CSV (and API record) where `energy_consumed` cannot be derived from `cpu_power + gpu_power + ram_power` times the duration, with nothing in the row explaining the gap.

## Reproduction

```python
t = EmissionsTracker(pue=1.5, measure_power_secs=1)
t.start(); time.sleep(10); t.stop()
```

In `emissions.csv` with `cpu_power=45.0`, `ram_power=10.0`, `duration=10.0`, the reported `energy_consumed` is `0.000229 kWh`, while `(45 + 10) W x 10 s = 0.000153 kWh`. The difference is exactly the 1.5 PUE factor.

## Root cause

In `codecarbon/emissions_tracker.py:1178` the PUE is applied to energy only:

```python
energy *= self._pue
```

The `power` object returned by `measure_power_and_energy` is then stored unscaled for every hardware branch (`emissions_tracker.py:1184`, `:1196`, `:1206`, `:1216`, `:1225`), and `_prepare_emissions_data` reports those raw averages next to the scaled energy. Water consumption is already derived from the PUE-scaled energy (`emissions_tracker.py:1179`), so power is the only quantity left at device scope.

## Expected vs actual

Expected: `power` and `energy` in the same record describe the same scope, so `energy_consumed` is reconstructible from the reported power and duration.

Actual: energy and water are facility-level (PUE applied), power is device-level (PUE not applied), and the two disagree by a factor of `pue`.

Users on the default `pue=1.0` see no change either way; only `pue != 1` deployments are affected.

davidberenstein1957:
## What changed

`_do_measurements` now scales the measured `power` by `self._pue` at the same point where it already scales `energy` (`codecarbon/emissions_tracker.py:1177`). `Power.__mul__` already exists, so every hardware branch below picks up the scaled value automatically.

## Why

PUE was applied to energy (and, through it, to water) but not to the power values reported in the same record. With `pue != 1` the CSV row was internally inconsistent: `energy_consumed` did not match `(cpu_power + gpu_power + ram_power) x duration`, and nothing in the row explained the gap. PUE is a facility property, so it applies equally to instantaneous power and integrated energy.

## Verification

Added `tests/test_emissions_tracker_constant.py::TestCarbonTrackerConstant::test_carbon_tracker_offline_constant_pue`, which runs the offline tracker with a forced CPU power and `pue=2.0`, asserts the reported `cpu_power` includes the PUE, and asserts `energy_consumed` is reconstructible from the reported power columns and duration. The test fails on master and passes with this change. Full file: `uv run pytest tests/test_emissions_tracker_constant.py -q` -> 7 passed.

## Note on visible numbers

**This changes reported power values for users running with `pue != 1`** — `cpu_power`, `gpu_power` and `ram_power` (and the per-hardware `logger.info` lines) are now facility-level. `energy_consumed` and emissions are unchanged. The default `pue=1.0` means the overwhelming majority of runs see no difference, but the change is not silent and warrants a changelog entry. Anyone manually multiplying power by PUE downstream would now double-apply it.

Follow-up, out of scope here: `EmissionCreate` (`codecarbon/core/api_client.py`) carries `wue` but no `pue`, so the server cannot tell whether an incoming `cpu_power` is device- or facility-level.

Closes #1320

## Changelog / release note

`CHANGELOG.md` is not on `master` yet (it arrives with `docs/traction-batch`), so the note lives in the docs instead: `docs/reference/output.md` now carries a "PUE and the power columns" admonition. When the changelog lands, copy this entry:

> **Changed** — the `pue` multiplier is now applied to the reported power columns (`cpu_power`, `gpu_power`, `ram_power`) as well as to the energy columns, so `energy_consumed` stays reconstructible from the reported power. Runs with `pue != 1` will see larger power values; energy and emissions are unchanged.

Consequence worth spelling out: `pue` also scales forced values, so `force_cpu_power=100` with `pue=1.5` now reports a 150 W CPU. That is consistent with treating the columns as facility-level, but it is surprising for a parameter named "force" — documented in `docs/reference/output.md` and in both `pue` docstrings.

Because this changes user-visible numbers, it should ride a **minor** release (3.4.0), not a patch.

### DIFF
--- a/docs/reference/output.md
+++ b/docs/reference/output.md
@@ -66,7 +66,7 @@ The package has an in-built logger that logs data into a CSV file named `emissio
 | ram_used_gb | Average RAM used during tracking period (GB) |
 
 !!! note "PUE and the power columns"
-    Since v3.4.0, the `pue` multiplier is applied to the reported power columns
+    Since v3.3.1, the `pue` multiplier is applied to the reported power columns
     (`cpu_power`, `gpu_power`, `ram_power`) as well as to the energy columns, so
     that `energy_consumed` stays reconstructible from the reported power. Before
     that, only the energy was scaled. With `pue=1.5`, a machine measured at 100 W
--- a/docs/reference/output.md
+++ b/docs/reference/output.md
@@ -65,6 +65,15 @@ The package has an in-built logger that logs data into a CSV file named `emissio
 | ram_utilization_percent | Average RAM utilization during tracking period (%) |
 | ram_used_gb | Average RAM used during tracking period (GB) |
 
+!!! note "PUE and the power columns"
+    Since v3.4.0, the `pue` multiplier is applied to the reported power columns
+    (`cpu_power`, `gpu_power`, `ram_power`) as well as to the energy columns, so
+    that `energy_consumed` stays reconstructible from the reported power. Before
+    that, only the energy was scaled. With `pue=1.5`, a machine measured at 100 W
+    is reported as 150 W, and that also applies to forced values: `force_cpu_power=100`
+    with `pue=1.5` reports a 150 W CPU. This is intended: the columns describe the
+    power drawn at the facility level, not at the socket.
+
 !!! note
     Developers can enhance the Output interface by implementing a custom class that extends `BaseOutput` at `codecarbon/output.py`. For example, to log into a database.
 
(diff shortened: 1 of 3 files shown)


inimaz (approved):
Thanks @davidberenstein1957 . This will keep aligned power and energy if PUE is given.

docs/reference/output.md
```diff
@@ -65,6 +65,15 @@ The package has an in-built logger that logs data into a CSV file named `emissio
 | ram_utilization_percent | Average RAM utilization during tracking period (%) |
 | ram_used_gb | Average RAM used during tracking period (GB) |
 
+!!! note "PUE and the power columns"
+    Since v3.4.0, the `pue` multiplier is applied to the reported power columns
```
inimaz:
```suggestion
    Since v3.3.1, the `pue` multiplier is applied to the reported power columns
```

PR #1333, opened 2026-08-12
fix: reject powermetrics on non-Apple-Silicon Macs

### LINKED ISSUE #1313: ApplePowermetrics._setup_cli silently succeeds on Intel Macs
## Problem

`ApplePowermetrics._setup_cli()` is meant to validate that powermetrics can be used, but on an Intel Mac it falls through all of its branches and returns without either assigning `self._cli` or raising. As a result `ApplePowermetrics()` succeeds on non-Apple-Silicon hardware, `is_powermetrics_available()` returns `True` whenever a passwordless sudo rule exists, and `ResourceTracker` ends up registering an `AppleSiliconChip` for a machine that has no Apple Silicon chip. The measurements it then reports come from a code path that was never intended for that hardware.

The same function also crashes with an unexpected `AttributeError` when `detect_cpu_model()` returns `None`, which the broad `except Exception` in `is_powermetrics_available()` turns into a misleading "Not using PowerMetrics" debug line, hiding a genuine cpuinfo failure.

## Reproduction

```python
from unittest import mock
from codecarbon.core.powermetrics import ApplePowermetrics

with mock.patch("codecarbon.core.powermetrics.sys.platform", "darwin"), \
     mock.patch("codecarbon.core.powermetrics.detect_cpu_model",
                return_value="Intel(R) Core(TM) i7-9750H"):
    pm = ApplePowermetrics()   # succeeds, should not
    pm._cli                    # AttributeError
```

## Root cause

- `codecarbon/core/powermetrics.py:117` — `_setup_cli()` has no terminal `else` for the "darwin but not Apple" case, so it returns silently.
- `codecarbon/core/powermetrics.py:122` — `cpu_model.startswith(...)` on the `Optional[str]` returned by `codecarbon/core/util.py:77`.
- `codecarbon/core/powermetrics.py:141` — `_log_values()` hardcodes `"powermetrics"` in the argv instead of using `self._cli`, which is why the missing assignment never surfaced as an error; the validation is effectively dead for its only caller.

Two smaller defects in the same argv:

- `codecarbon/core/powermetrics.py:146` — a stray `""` is passed to `powermetrics` as a positional argument. Harmless on current macOS, but it is unintentional and a stricter parser would reject it.
- `codecarbon/core/powermetrics.py:156` — `subprocess.call` has no timeout, so a hung `powermetrics` blocks the measurement thread indefinitely.

## Expected vs actual

Expected: on a Mac that is not Apple Silicon, `_setup_cli()` raises `SystemError`, `is_powermetrics_available()` returns `False`, and `ResourceTracker` falls through to Intel Power Gadget or the CPU-load fallback as `_try_platform_cpu_backend` (`codecarbon/core/resource_tracker.py:225`) intends.

Actual: `ApplePowermetrics()` is constructed successfully, powermetrics is selected, and an `AppleSiliconChip` is registered on Intel hardware. With `detect_cpu_model()` returning `None`, an `AttributeError` is swallowed and reported as powermetrics simply being unavailable.

Note for the changelog: machines that were accidentally using powermetrics on Intel Macs will change tracking method after this fix.

davidberenstein1957:
## What changed

`ApplePowermetrics._setup_cli()` now raises instead of falling through:

- a Mac whose CPU is not Apple Silicon raises `SystemError` (previously the method returned without assigning `self._cli` and without raising, so `ApplePowermetrics()` succeeded on Intel Macs);
- `detect_cpu_model()` returning `None` raises the same `SystemError` rather than an unexpected `AttributeError`;
- `self._cli` is assigned on the one path that validated the executable.

`_log_values()` now uses `self._cli` instead of a hardcoded `"powermetrics"`, so the validated path is the one actually executed. Two smaller defects in the same argv are fixed alongside it: a stray `""` positional argument passed verbatim to `powermetrics`, and the missing timeout on `subprocess.call`, which let a hung `powermetrics` block the measurement thread indefinitely.

## Why

Because `_setup_cli()` never raised on Intel Macs, `is_powermetrics_available()` returned `True` whenever a permissive sudo rule existed, and `_try_platform_cpu_backend` (`codecarbon/core/resource_tracker.py:225`) routed those machines into `_setup_powermetrics`, registering an `AppleSiliconChip` on hardware that has none. With `detect_cpu_model()` returning `None`, the resulting `AttributeError` was swallowed by the broad `except Exception` in `is_powermetrics_available()` and reported as powermetrics simply being unavailable, hiding genuine cpuinfo failures.

Note for the changelog: machines that were accidentally using powermetrics on Intel Macs now fall through to Intel Power Gadget or the CPU-load fallback, which is the intended routing but does change their tracking method.

## How it was verified

Four new tests in `tests/test_powermetrics.py` cover the Intel Mac rejection, the `None` CPU model, the clean argv with a timeout, and the timeout warning path. All four fail on `master` and pass with this change; the full `tests/test_powermetrics.py` and `tests/test_resource_tracker.py` suites pass (51 tests). `ruff format --check` is clean on both touched files.

Closes #1313

### DIFF
--- a/codecarbon/core/powermetrics.py
+++ b/codecarbon/core/powermetrics.py
@@ -118,17 +118,19 @@ class ApplePowermetrics:
         """
         Setup cli command to run Powermetrics
         """
-        if self._system.startswith("darwin"):
-            cpu_model = detect_cpu_model()
-            if cpu_model.startswith("Apple"):
-                if shutil.which(self._osx_silicon_exec):
-                    self._cli = self._osx_silicon_exec
-                else:
-                    raise FileNotFoundError(
-                        f"Powermetrics executable not found on {self._system}"
-                    )
-        else:
+        if not self._system.startswith("darwin"):
             raise SystemError("Platform not supported by Powermetrics")
+        cpu_model = detect_cpu_model() or ""
+        if not cpu_model.startswith("Apple"):
+            raise SystemError(
+                "Powermetrics is only supported on Apple Silicon, "
+                f"detected CPU: {cpu_model!r}"
+            )
+        if not shutil.which(self._osx_silicon_exec):
+            raise FileNotFoundError(
+                f"Powermetrics executable not found on {self._system}"
+            )
+        self._cli = self._osx_silicon_exec
 
     def _log_values(self) -> None:
         """
@@ -140,10 +142,9 @@ class ApplePowermetrics:
             # Run the powermetrics command with sudo and capture its output
             cmd = [
                 "sudo",
-                "powermetrics",
+                self._cli,
                 "-n",
                 str(self._n_points),
-                "",
                 "--samplers",
                 "cpu_power",
                 "--format",
--- a/tests/test_powermetrics.py
+++ b/tests/test_powermetrics.py
@@ -209,6 +209,40 @@ class TestApplePowerMetrics:
             with pytest.raises(FileNotFoundError):
                 ApplePowermetrics()
 
+    def test_setup_cli_raises_on_intel_mac(self):
+        with (
+            mock.patch("codecarbon.core.powermetrics.sys.platform", "darwin"),
+            mock.patch(
+                "codecarbon.core.powermetrics.detect_cpu_model",
+                return_value="Intel(R) Core(TM) i7-9750H",
+            ),
+        ):
+            with pytest.raises(SystemError):
+                ApplePowermetrics()
+
+    def test_setup_cli_raises_when_cpu_model_unknown(self):
+        with (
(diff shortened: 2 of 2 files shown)


davidberenstein1957:
Closing as folded into #1345. Both are about ApplePowermetrics being dishonest in two ways: it constructs successfully on hardware it cannot measure, and it returns NaN when it measures nothing. Same file, same class, adjacent methods, and they collided textually while both extending tests/test_powermetrics.py, so they are now one 'harden powermetrics' review. The subprocess timeout that was bundled in here has been split out separately as #1397, since it is an unrelated change with its own failure mode.

PR #1334, opened 2026-08-12
fix: raise on offline tracker configuration errors

### LINKED ISSUE #1311: OfflineEmissionsTracker silently returns a half-built object when construction fails
When `OfflineEmissionsTracker` is constructed with an invalid configuration, no exception is raised. The constructor returns a partially initialised object, every subsequent call fails silently, and the run produces no emissions data at all. The online `EmissionsTracker` raises correctly for the same input, so the two constructors have opposite error semantics.

## Reproduction

```python

tracker = OfflineEmissionsTracker(country_iso_code="FRA", output_dir="/does/not/exist")
tracker.start()
# ... user workload ...
print(tracker.stop())   # None, and emissions.csv was never written
```

`EmissionsTracker(output_dir="/does/not/exist")` raises `OSError: Folder ... does not exist !` as expected.

## Root cause

`OfflineEmissionsTracker.__init__` is decorated with `@suppress(Exception)` at `codecarbon/emissions_tracker.py:1312`. The validation in `BaseEmissionsTracker._set_from_conf` (`codecarbon/emissions_tracker.py:138-140`) raises `OSError`, but the decorator swallows it, so `super().__init__` never reaches `_initialize_runtime_state()` / `_initialize_scheduler_state()` (`codecarbon/emissions_tracker.py:603-604`). The object then has no `_start_time`, `_hardware`, `_scheduler` or `_output_handlers`.

`start()` and `stop()` are also `@suppress(Exception)` (lines 692 and 886), so the resulting `AttributeError: OfflineEmissionsTracker object has no attribute _start_time` is downgraded to warnings as well. Under the CLI default `log_level="error"` (`codecarbon/cli/main.py`) nothing is printed at all.

`suppress` itself lives at `codecarbon/core/util.py:21-30`.

## Expected vs actual

- Expected: a misconfigured offline tracker raises at construction time, like the online tracker does.
- Actual: construction succeeds, `start()`/`stop()` no-op with warnings, `stop()` returns `None`, and no emissions file is produced. Users reasonably conclude CodeCarbon does not work on their machine.

Suppressing exceptions in `start`/`flush`/`stop` is deliberate and should stay: CodeCarbon must never crash a training job at runtime. The decorator on the constructor is a different case, since failure there means the object invariants were never established.

davidberenstein1957:
Closes #1311

## What changed

Removed `@suppress(Exception)` from `OfflineEmissionsTracker.__init__` (`codecarbon/emissions_tracker.py`).

## Why

The decorator swallowed configuration errors raised during construction, most visibly the `OSError` from `_set_from_conf` when `output_dir` does not exist. The constructor then returned an object that had never reached `_initialize_runtime_state()` / `_initialize_scheduler_state()`, so it had no `_start_time`, `_hardware`, `_scheduler` or `_output_handlers`. The subsequent `start()` and `stop()` calls are themselves suppressed, so they degraded into `AttributeError` warnings and the run finished with no emissions file and no exception. Under the CLI default `log_level="error"` nothing was printed at all.

The online `EmissionsTracker` already raises for the same input, so this also removes an asymmetry between the two constructors.

Suppression on `start`, `flush` and `stop` is deliberately left in place: runtime measurement errors must never crash a user job. Only construction, where the object invariants were never established, now fails loudly.

## Verification

Added `test_offline_tracker_raises_on_invalid_output_dir` in `tests/test_offline_emissions_tracker.py`, asserting both the offline and the online tracker raise `OSError` for a non-existent `output_dir`. It fails on master and passes with this change.

```
uv run pytest tests/test_offline_emissions_tracker.py tests/test_emissions_tracker.py -q
35 passed
```

## Note for reviewers

This is a behaviour change at a public boundary: code that today constructs a misconfigured offline tracker and carries on will now raise. That is the intended correction but it deserves a changelog entry, and ideally a minor release.

## Changelog / release note (added)

`CHANGELOG.md` is not on `master` yet (it lands with `docs/traction-batch`), so the entry lives in the docs: `docs/reference/api.md` now carries a "Constructor errors (changed in v3.4.0)" warning. When the changelog lands, copy:

> **Changed** — `OfflineEmissionsTracker` no longer swallows exceptions raised while constructing the tracker. Configuration errors (for example a non-existent `output_dir`) now propagate to the caller, matching `EmissionsTracker`. Suppression on `start`/`flush`/`stop` is unchanged.

This must ride a **minor** release (3.4.0), not a patch.

A second constructor-raise case is now covered: `test_offline_tracker_raises_on_invalid_region` exercises the offline-only region check, which runs before `super().__init__` and so was suppressed by a different code path than the `output_dir` `OSError`. Both new tests fail when `@suppress(Exception)` is put back on the constructor.

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -1309,7 +1309,6 @@ class OfflineEmissionsTracker(BaseEmissionsTracker):
     _country_iso_code = None
     _country_name, _region, country_2letter_iso_code = None, None, None
 
-    @suppress(Exception)
     def __init__(
         self,
         *args,
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -8,6 +8,13 @@ Parameters can be set via `EmissionsTracker()`, `OfflineEmissionsTracker()`, the
     PUE is a multiplication factor provided by the user. Old datacenters have PUE
     up to 2.2, new greener ones as low as 1.1.
 
+!!! warning "Constructor errors (changed in v3.4.0)"
+    `OfflineEmissionsTracker(...)` used to swallow every exception raised while
+    building the tracker, returning a half-initialised object that silently
+    recorded nothing. It now propagates those errors, like `EmissionsTracker`
+    always did — for example an `output_dir` that does not exist raises instead of
+    logging. Code that relied on the constructor never raising needs a `try`/`except`.
+
 !!! note "GPU selection"
     If you use `CUDA_VISIBLE_DEVICES` or `ROCR_VISIBLE_DEVICES` to set GPUs, CodeCarbon will automatically
     populate `gpu_ids`. Manual `gpu_ids` overrides this.
--- a/tests/test_offline_emissions_tracker.py
+++ b/tests/test_offline_emissions_tracker.py
@@ -6,7 +6,7 @@ from unittest import mock
 
 import pandas as pd
 
-from codecarbon.emissions_tracker import OfflineEmissionsTracker
+from codecarbon.emissions_tracker import EmissionsTracker, OfflineEmissionsTracker
 from tests.testutils import get_custom_mock_open, get_test_data_source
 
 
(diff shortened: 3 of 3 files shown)


benoit-cty:

## Verdict: 💬 Needs a maintainer decision (leaning approve, for a minor release)

**The bug is real.** A bad `output_dir` returns a half-built `OfflineEmissionsTracker` that silently records nothing, and the new tests fail on master. Removing `@suppress(Exception)` from `OfflineEmissionsTracker.__init__` (`emissions_tracker.py:1312`) matches the online `EmissionsTracker`, which already raises. `start`, `stop` and `flush` stay suppressed.

**Behaviour change to be aware of:**
- CodeCarbon's contract has been "never crash user code". With this PR:
  - `@track_emissions(offline=True, output_dir="missing")` builds the tracker inside `wrapped_fn`, so the decorated function **never runs** and the exception reaches the user's training job. Before, the job ran untracked.
  - `with OfflineEmissionsTracker(...)` now raises too.
- Suggestion for keeping both behaviours: keep direct construction strict (this PR), but in `track_emissions` catch construction errors, log them at ERROR, and run the wrapped function untracked.
- The docs note says "changed in v3.4.0" (latest tag is v3.3.1), so this must go into a minor release, not a patch. Please also add a changelog entry.

**Nit:**
- `test_offline_tracker_raises_on_invalid_region` expects an `AssertionError` from an `assert` (`emissions_tracker.py` ~L1344). Under `python -O`, asserts are stripped and nothing is raised. Please raise `ValueError` explicitly.

davidberenstein1957:
Made the changes in e77657f4: `region` / `country_2letter_iso_code` type checks now raise `ValueError` instead of `assert`.

Not done, pending a maintainer decision: whether `track_emissions` should catch construction errors and run the function untracked, and which release this lands in. No changelog file exists in the repo, so the entry is left for the release notes.

davidberenstein1957:
Made the changes in 1e238ba5: `@track_emissions` now logs construction errors and runs the function untracked, same as `start_task`. Direct construction still raises. Targeting a minor release (v3.4.0).

PR #1336, opened 2026-08-12
fix: make tracker.stop() idempotent

### LINKED ISSUE #1307: tracker.stop() is not idempotent: a second call writes a duplicate emissions row
Calling `stop()` twice on a tracker produces two complete rows in `emissions.csv` (and two POSTs with the API handler, two `handler.exit()` calls, etc.) for a single run. Summing the output double counts the run.

This is easy to hit because `__exit__` calls `stop()` unconditionally, so the common shape of a `with` block plus an explicit `stop()` to grab the return value double-stops.

## Reproduction

```python
import time

tracker.start()
time.sleep(2)
tracker.stop()
tracker.stop()
# emissions.csv now has 2 rows, same run_id, slightly different durations
```

## Root cause

`BaseEmissionsTracker.stop()` (`codecarbon/emissions_tracker.py:887`) has no terminal-state flag. `_start_time` is never cleared, so the guard at `emissions_tracker.py:905` does not trip on the second call. Both schedulers are already `None`, so teardown is skipped, but the code then carries straight on through `_measure_power_and_energy_if_stale()`, `_prepare_emissions_data()`, `_persist_data()` and `handler.exit()`.

The `else: logger.warning("Tracker already stopped !")` at `emissions_tracker.py:914` is bound to the `_scheduler_monitor_power` truthiness check rather than to an explicit stopped state, and it is purely advisory — nothing returns early.

Additionally, `self._lock.release()` (`emissions_tracker.py:901`) runs before the guard, so the second call retries `os.remove` on an already-removed lock file (the `ENOENT` is swallowed in `codecarbon/lock.py`).

## Expected vs actual

- Expected: the second `stop()` is a no-op that returns the already-computed emissions value.
- Actual: a second full measurement, a duplicate row through every output handler, and a second `handler.exit()`.

davidberenstein1957:
Calling `stop()` twice wrote a second complete emissions row (two CSV rows, two API POSTs, two `handler.exit()` calls) for a single run, because nothing recorded that the tracker had stopped — `_start_time` stays set and both schedulers being `None` was only used to emit an advisory warning.

## What changed

- `_initialize_runtime_state()` now initializes `_is_stopped`, `final_emissions` and `final_emissions_data`.
- `stop()` returns early on `_is_stopped`, returning the cached `final_emissions`.
- The lock release moved after the guard, so a repeat `stop()` no longer retries `os.remove` on an already-removed lock file.
- The misplaced `else: logger.warning("Tracker already stopped !")` (bound to the `_scheduler_monitor_power` check) is replaced by the real terminal-state guard.

## Why

`__exit__` calls `stop()` unconditionally, so a `with` block plus an explicit `stop()` — the usual way to get the return value — double counts the whole run.

## Verification

`tests/test_emissions_tracker.py::TestCarbonTracker::test_offline_tracker_stop_is_idempotent` asserts one CSV row and an identical return value after a double stop. It fails on master (2 rows) and passes with this change. Full file: 32 passed.

Closes #1307

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -268,6 +268,10 @@ class BaseEmissionsTracker(ABC):
 
     def _initialize_runtime_state(self) -> None:
         self._start_time: Optional[float] = None
+        # None while the tracker is running.
+        self._stopped_at: Optional[float] = None
+        self.final_emissions: Optional[float] = None
+        self.final_emissions_data: Optional[EmissionsData] = None
         self._last_measured_time: float = time.perf_counter()
         self._total_energy: Energy = Energy.from_energy(kWh=0)
         self._total_emissions: float = 0.0
@@ -706,6 +710,14 @@ class BaseEmissionsTracker(ABC):
             )
             return
         if self._start_time is not None:
+            if self._stopped_at is not None:
+                # `start()` is wrapped in @suppress(Exception), so raising here
+                # would be swallowed: log instead of pretending it worked.
+                logger.error(
+                    "This tracker was already stopped and cannot be restarted. "
+                    "Create a new tracker instead."
+                )
+                return
             logger.warning("Already started tracking")
             return
 
@@ -899,12 +911,17 @@ class BaseEmissionsTracker(ABC):
                 "Another instance of codecarbon is already running. Exiting."
             )
             return
-        if not self._allow_multiple_runs:
-            # Release the lock
-            self._lock.release()
         if self._start_time is None:
             logger.error("You first need to start the tracker.")
             return None
+        if self._stopped_at is not None:
+            logger.warning("Tracker already stopped !")
+            return self.final_emissions
+        self._stopped_at = time.perf_counter()
+
+        if not self._allow_multiple_runs:
+            # Release the lock
+            self._lock.release()
 
         if self._scheduler:
             self._scheduler.stop()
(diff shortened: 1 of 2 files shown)


benoit-cty:

## Verdict: 🔧 Request changes (rebase down, or close)

**Mostly superseded by #1408** (8171d136, now on master). `stop()` already sets `_start_time = None` and returns the cached `final_emissions`, and issue #1307 is closed. The only new behaviour left here is "refuse to restart after stop".

**Must fix:**
1. **Conflicts with master**, and the remaining logic won't work after a rebase.
   - The `_stopped_at` guard duplicates #1408, and conflicts with it in `stop()` (`emissions_tracker.py` ~L911-925).
   - The restart refusal sits inside `if self._start_time is not None:` in `start()`. After #1408 that value is `None` once stopped, so **the refusal would never fire**.
   - If you keep this PR, rebase it down to the restart refusal only, and use a dedicated flag (e.g. `self._stopped = True` set in `stop()`).
2. **Behaviour choice to confirm with maintainers.** On master today, start → stop → start → stop "works", noisily:
   - The second `start()` logs a suppressed `AttributeError: 'NoneType' object has no attribute 'start'`, because the scheduler is `None`.
   - The second `stop()` writes a second CSV row whose energy includes the first run's.

   With this PR, the second run writes nothing and only logs an error. Refusing is defensible, but the error message should say so clearly and suggest creating a new tracker. The alternative is to support restarting properly: re-create the scheduler and reset the counters.

**Nit:**
- `test_offline_tracker_stop_is_idempotent` overlaps with #1408's tests. Drop it when rebasing.

davidberenstein1957:
Made the changes in ee7ff846: merged master, reduced to the restart refusal only.

- Dropped the `_stopped_at` guard (#1408 covers it) and `test_offline_tracker_stop_is_idempotent`.
- `start()` after `stop()` now checks a dedicated `_stopped` flag and logs a clear error suggesting a new tracker.
- Kept the lock-released-once test, since #1408 added no tests.

Refusing vs. properly supporting restart is still your call.

PR #1337, opened 2026-08-12
fix: allow restarting a tracker after stop()

### LINKED ISSUE #1328: Tracker cannot be restarted: start() after stop() is a no-op and the next row's duration is wrong
## Problem

A tracker cannot be restarted. After `stop()`, a subsequent `start()` is a no-op ("Already started tracking"), and the next `stop()` still writes a row — but that row's `duration` covers everything since the *original* `start()`, including the idle gap, and the phase was never sampled. This is the natural pattern for benchmarking loops ("measure phase A, stop, measure phase B") and for notebooks where a cell is re-run.

## Reproduction

```python
tracker.start(); time.sleep(3); tracker.stop()   # row 1: duration ~3
time.sleep(3)
tracker.start()                                  # logs "Already started tracking", does nothing
time.sleep(3); tracker.stop()                    # row 2: duration ~9, unsampled
assert tracker._scheduler is None                # no periodic measurement at all in phase B
```

## Root cause

`stop()` tears down both schedulers but never clears the start time (`codecarbon/emissions_tracker.py:909-914`):

```python
if self._scheduler:
    self._scheduler.stop()
    self._scheduler = None
if self._scheduler_monitor_power:
    self._scheduler_monitor_power.stop()
    self._scheduler_monitor_power = None
```

`start()` then bails out because `_start_time` is still set (`emissions_tracker.py:708-710`), and even if that guard is bypassed, `self._scheduler.start()` at `emissions_tracker.py:725` raises `AttributeError` on `None`, which the `@suppress(Exception)` decorator on `start()` turns into a warning. `_initialize_scheduler_state()` (`emissions_tracker.py:344`) is only ever called from `__init__` (`emissions_tracker.py:604`), so the schedulers are never rebuilt.

Lifecycle state is split across `_start_time`, `_scheduler` and `_scheduler_monitor_power` with no single reset path.

## Expected vs actual

Expected: `start()` after `stop()` resumes tracking; the periodic measurement runs again and the next row's `duration` reflects active tracking time.

Actual: the second `start()` is silently a no-op, phase B gets a single catch-all measurement from `_measure_power_and_energy_if_stale()` at `stop()`, and the row's `duration` includes the paused interval — so `emissions_rate` for that row is wrong too.

Measured on current master (`measure_power_secs=1`, 3 s per phase): row 1 `duration=3.088958`, row 2 `duration=6.104781`.

davidberenstein1957:
## What

`stop()` tore down both schedulers but never cleared `_start_time`, so a later `start()` returned early with `"Already started tracking"`. The tracker looked started but was never sampled, and the next `stop()` wrote a row whose `duration` covered everything since the *original* `start()` — including the idle gap — making that row's `duration` and `emissions_rate` wrong.

## Changes

- `stop()` records `_paused_at` after the final persist, and returns early (with the existing `"Tracker already stopped !"` warning) if the tracker is already stopped, instead of writing a duplicate row.
- `start()` rebuilds the schedulers when they were discarded, and on a restart shifts `_start_time` by the paused interval rather than re-stamping it. The energy accumulators are deliberately not reset on restart, so shifting keeps `duration` and the accumulators on the same clock: both cover active time only.

Note: a restarted run stays cumulative, and row 2's `duration` is now the sum of the active phases rather than wall-clock since the first `start()`.

## Verification

New test `tests/test_emissions_tracker.py::TestCarbonTracker::test_tracker_can_be_restarted_after_stop` — start/2s/stop, 2s pause, start/2s/stop — asserts the scheduler exists again after the second `start()` and that the second row's `duration` is ~4 s, not ~6 s. It fails on master and passes with this change. Full suite: 627 passed, 21 skipped.

Closes #1328

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -706,6 +706,16 @@ class BaseEmissionsTracker(ABC):
             )
             return
         if self._start_time is not None:
+            if self._scheduler is None:
+                # stop() drops the schedulers, releases the lock and exits the
+                # output handlers, so a stopped tracker cannot be restarted.
+                # `start()` is wrapped in @suppress(Exception), so raising here
+                # would be swallowed: log instead of pretending it worked.
+                logger.error(
+                    "This tracker was already stopped and cannot be restarted. "
+                    "Create a new tracker instead."
+                )
+                return
             logger.warning("Already started tracking")
             return
 
--- a/tests/test_emissions_tracker.py
+++ b/tests/test_emissions_tracker.py
@@ -1108,3 +1108,20 @@ class TestCarbonTracker(unittest.TestCase):
 
         # Verification: If it wasn't cumulative, it would be 3.0 kWh * 300 g/kWh = 0.9 kg
         self.assertLess(data3.emissions, 0.8)
+
+
+class TestRestartAfterStop(unittest.TestCase):
+    def test_start_after_stop_is_refused(self):
+        tracker = OfflineEmissionsTracker(
+            country_iso_code="FRA",
+            save_to_file=False,
+            allow_multiple_runs=True,
+            measure_power_secs=10,
+        )
+        tracker.start()
+        tracker.stop()
+        with self.assertLogs("codecarbon", level="ERROR") as logs:
+            tracker.start()
+        self.assertIn("cannot be restarted", "".join(logs.output))
+        # Refused, not half-restarted: nothing was rebuilt.
+        self.assertIsNone(tracker._scheduler)


davidberenstein1957:
Closing as folded into #1336. Both PRs were about start() and stop() disagreeing on whether the tracker has run, and they had already converged: #1336 introduces _stopped_at as the canonical lifecycle flag, while this PR was inferring the same state from _scheduler is None. Two flags for one fact. The refusal now reads 'if self._stopped_at is not None' on #1336, so it is one flag and three guards in a single review. The honest-error change and its test came across unchanged.

Issue #1338, opened 2026-08-12
Scaling the client: scheduler re-entrancy, API delivery, install weight, CI coverage


davidberenstein1957:
Umbrella issue for a set of changes making the `codecarbon` client cheaper to run and cheaper to install. Each item is a separate, independently reviewable PR; this issue only tracks them.

## Checklist

- [ ] #1339 — `refactor(scheduler): run ticks on one daemon thread instead of chained Timers`. Stops the scheduler re-entering its own callback when a measurement outlasts the interval, and stops it re-arming after `stop()`.
- [ ] #1340 — `feat(api): reuse connections and retry transient API failures`. Adds a `Session`, a retry policy and a run-creation herd guard, instead of one connection per request with failures logged away.
- [ ] #1341 — `fix(api): stamp emissions with measurement time, not send time`. See #1312.
- [ ] #1342 — `ci: test every Python version the package promises`. The matrix skips 3.10 and 3.11, which the classifiers advertise; also adds a bare-install wheel smoke job.
- [ ] #1343 — `perf(deps): drop pandas and numpy from the default install`.

## Overlap to resolve before reviewing

**#1324 (`fix: stop PeriodicScheduler from re-arming itself after stop()`) and #1339 fix the same scheduler defect by different means.** #1324 is a narrow fix to the existing chained-`Timer` design; #1339 replaces that design with a single daemon thread. Please pick one before reviewing — merging both, or reviewing them independently, means duplicated work and a conflicting merge.

## Out of scope, decided against

- **Client-side batching of emissions.** The `/emissions/batch` endpoint does not exist — `carbonserver` has only the single-row `POST /emissions` — so the batch path would be dead code that always 404s and falls back to the loop it replaced. Should land together with the server-side endpoint, or not at all.
- **A disk spill buffer in `ApiClient`.** Retries cover transient failures, which is the overwhelming majority of them. A spill buffer adds write-failure surface to the measurement path and, being off by default, is code nobody exercises until it is needed and broken. Worth building when someone actually reports losing data across a restart.
- **Moving the CLI and auth dependencies into extras.** Worth about 23 MB with no runtime benefit, and it breaks `pip install codecarbon && codecarbon monitor`, a documented headline feature.

Before/after measurements for each item live in the corresponding PR description rather than here, so they stay next to the code they describe.

davidberenstein1957:
[...]

PR #1339, opened 2026-08-12
refactor(scheduler): run ticks on one daemon thread instead of chained Timers

davidberenstein1957:
Part of #1338.

`PeriodicScheduler` chained a fresh `threading.Timer` per tick and armed the successor *before* running the payload. Replaced with a single daemon thread waiting on an `Event` against an absolute `time.monotonic()` deadline, with a catch-up guard, in-loop exception logging, and a `stop()` that sets the event and joins the worker.

## Measured, master vs. this branch

| | master | branch |
|---|---|---|
| overlapping entries (payload 2.4x interval, ~18 calls) | 17 | 0 |
| extra calls after `stop()` (gated race repro) | 8, and never stops | 0 |
| drift per tick @ 1s | +4.3 ms (0.4%) | +0.1 ms |
| distinct worker thread idents over 30 ticks | 2 | 1 |
| live threads during a run | 3 | 2 |
| `stop()` latency @ 10s interval | 0.0 ms | 0.1 ms |

The case for merging is the first two rows. Re-entrancy needs no race window — just a measurement slower than the interval, which is routine for `_scheduler_monitor_power` (1s, hard-coded) when `powermetrics` or RAPL is slow, and two concurrent mutations of `_total_energy` / `_last_measured_time` corrupt the numbers users report. The `stop()` race is unlikely per tick but unbounded in consequence: measurements and API pushes continue after the tracker has written its final row.

Two arguments **against** overselling this, both from the same measurements: thread count never grows (each Timer dies as its successor starts), and drift is 0.4%, invisible because energy is computed from measured elapsed time. Neither would justify the change alone.

## ⚠️ Behaviour changes (no CHANGELOG file exists in this repo, so recording them here)

1. **`tracker.stop()` can now block for up to `min(measure_power_secs, 5.0)` seconds.** It previously returned immediately. Normally the wait is ~0.1 ms; the bound is only reached if a measurement is genuinely wedged. That wait *is* the fix for "stop() returns while the callback is still running", but it is the change most likely to be noticed in the wild. The 5s cap is a judgement call and carries a `ponytail:` comment.
2. **`tracker.start_task()` inherits the same bound.** `start_task` calls `self._scheduler.stop()` (`emissions_tracker.py:754-755`), so it too can now block for up to `min(measure_power_secs, 5.0)` where it previously returned at once. Same normal-case cost (~0.1 ms); worth knowing for callers that start many short tasks in a loop.
3. **A slow callback now yields missing measurements instead of overlapping ones**, so the existing "Background scheduler didn't run for a long period" warning will fire where master silently produced corrupted concurrent measurements.
4. **A wedged scheduler now refuses to restart rather than double-starting.** See below.

## Wedged-thread handling

If the bounded join in `stop()` times out, `stop()` keeps its reference to the thread instead of setting `self._thread = None`. Nulling it was unsafe: `_stopped` is `self._thread is None or not self._thread.is_alive()`, so a forgotten-but-alive thread made the next `start()` clear `_stop_event` and arm a second thread — and the original, unblocked a moment later, resumed ticking from the cleared event. Two live schedulers mutating `_total_energy` and `_last_measured_time`, which is exactly what this PR set out to prevent.

Holding the reference makes `_stopped` report the truth, so `start()` stays a no-op until the wedged thread actually exits (it exits on its next `wait()`, since `_stop_event` is still set). `stop()` logs a warning when the join gives up, so the no-op is not silent.

## Other notes

- `from_run` is gone rather than kept as an ignored parameter — nothing outside the removed `_run` ever passed it, and `PeriodicScheduler` is not re-exported from `codecarbon/__init__.py`. `_stopped` is kept as a property so its two existing readers need no change.
- The new tests use real sleeps at a 0.05s interval. Stable here, but they are wall-clock tests on shared runners; if they flake, the right fix is injecting a clock, not loosening tolerances.
- Dropped from the original proposal: a drift-assertion test (it flaked under full-suite load while passing standalone, and 0.4% drift isn't worth guarding) and a frozen-clock test for the three-line catch-up guard.

## Tests

New `tests/test_scheduler.py`, **7 tests, 7 passed**. Against master's scheduler, 4 of them fail. Three are fail-before/pass-after for the threading rewrite: `test_ticks_run_on_a_single_thread`, `test_slow_function_is_never_re_entered`, `test_start_is_idempotent_and_restartable`.

`test_start_after_a_timed_out_stop_does_not_run_two_loops` guards the wedged-thread case: it blocks the callback, lets `stop()`'s join time out, calls `start()`, then asserts only one thread ident ever ran the callback. Reverting the one-line `stop()` change fails it with `two scheduler loops ran: {...}` (verified).

Full suite `633 passed, 21 skipped` (excluding `tests/test_viz_data.py`, which fails to import on master too — `dash` not installed). `uv run pre-commit run --all-files` passes.

Not addressed here: nothing restarts `_scheduler` after `start_task` stops it (`emissions_tracker.py:754-755`) — a real pre-existing bug, but a separate one.

### DIFF
--- a/codecarbon/external/scheduler.py
+++ b/codecarbon/external/scheduler.py
@@ -1,52 +1,74 @@
-from threading import Lock, Timer
+import time
+from threading import Event, Thread, current_thread
+
+from codecarbon.external.logger import logger
 
 
 class PeriodicScheduler:
     """
-    A periodic task running in threading.Timers
-    From https://stackoverflow.com/a/18906292/14541668
+    Run ``function`` every ``interval`` seconds on a single daemon thread.
+
+    The deadline is absolute, so the cadence does not drift with the time the
+    function itself takes. A tick that overruns its slot is skipped rather than
+    queued, so the function is never re-entered.
     """
 
     def __init__(self, interval, function, *args, **kwargs):
         """
-        Init the scheduler. You have to call start() after initialization.
-        ::interval:: interval in seconds to run the function.
-        ::function:: function to run.
+        ::interval:: in seconds, the delay between two calls to function.
+        ::function:: the function to call.
         ::args:: args to pass to the function.
         ::kwargs:: kwargs to pass to the function.
         """
-        self._lock = Lock()
-        self._timer = None
-        self.function = function
         self.interval = interval
+        self.function = function
         self.args = args
         self.kwargs = kwargs
-        self._stopped = True
+        self._thread = None
+        self._stop_event = None
 
-    def start(self, from_run=False):
-        """
-        Start the scheduler.
-        ::from_run:: For internal purposes to allow re-scheduling
-        Please do not use from_run=True until you know what you do !
-        """
-        self._lock.acquire()
-        if from_run or self._stopped:
-            self._stopped = False
-            self._timer = Timer(self.interval, self._run)
-            self._timer.daemon = True
-            self._timer.start()
-        self._lock.release()
+    @property
+    def _stopped(self):
+        return self._thread is None
 
-    def _run(self):
-        self.start(from_run=True)
-        self.function(*self.args, **self.kwargs)
+    def start(self):
+        """
+        Start the scheduler. Calling it on a running scheduler is a no-op.
+        """
+        if self._thread is not None:
+            return
+        # One event per run: a thread left behind by a timed-out stop() keeps
+        # its own, still-set event and exits as soon as its callback returns.
+        self._stop_event = Event()
+        self._thread = Thread(
+            target=self._loop,
+            args=(self._stop_event,),
+            daemon=True,
+            name=f"codecarbon-{getattr(self.function, '__name__', 'scheduler')}",
+        )
+        self._thread.start()
 
-    def stop(self):
+    def _loop(self, stop_event):
+        next_call = time.monotonic() + self.interval
+        while not stop_event.wait(max(0.0, next_call - time.monotonic())):
+            try:
+                self.function(*self.args, **self.kwargs)
+            except Exception:  # noqa: BLE001 - must not kill the only thread
+                logger.error("Scheduled measurement failed", exc_info=True)
+            # Absolute deadline, but never a burst of catch-up ticks.
+            next_call = max(next_call + self.interval, time.monotonic())
+
+    def stop(self, timeout=None):
         """
-        Stop the scheduler.
+        Stop the scheduler and wait for the in-flight call to return.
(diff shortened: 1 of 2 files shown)


davidberenstein1957:
Note for whoever reviews this: #1324 fixes the same scheduler defect by a different route. It is a narrow fix to the existing chained-Timer design, while this PR replaces the design with a single daemon thread. Worth picking one before reviewing either, otherwise they conflict on merge.

benoit-cty:

Replacing the chained `Timer`s with one daemon thread on a `monotonic` deadline is a real improvement:
- no drift
- overrunning ticks are skipped rather than overlapping
- exceptions are logged inside the loop
- the final measure and flush always still run

The 7 new scheduler tests pass, and the PR merges cleanly with master.

**Issues:**
1. **The "wedged-thread handling" in the description isn't implemented** (medium).
   - The description says a timed-out `stop()` keeps its thread reference, logs a warning, and `start()` then refuses to run a second loop.
   - The code (`scheduler.py:69`) always sets `_thread = None`, logs nothing, and uses one Event per run.
   - So after a `stop()` that timed out (e.g. via `start_task()`, `emissions_tracker.py:754`), `start()` launches a new loop while the old callback is still running. That gives two concurrent `_measure_power_and_energy` calls, which is exactly the overlap this PR wants to remove.
   - `test_start_after_a_timed_out_stop_leaves_only_the_new_loop` explicitly accepts that overlap.
   - Fix: if `join()` times out, keep the reference and log a warning. In `start()`, if the previous thread is still alive, wait for it (or refuse) before starting a new loop. Update the test.
2. **The final flush can still overlap an in-flight tick, silently** (medium).
   - `stop()` waits at most `min(interval, 5 s)`, but a tick can make network calls: Electricity Maps (30 s timeout), and the API (up to about 26 s with #1340).
   - If one is in flight, `tracker.stop()` (`emissions_tracker.py:909-930`) runs the final measurement and `_persist_data` concurrently with it.
   - At minimum, log a warning when the join times out. Better, make the final measurement wait for the in-flight tick, e.g. with a lock around the tick body.

**Nits:**
- Pre-existing: `start_task` never restarts `_scheduler`.
- The tests rely on real sleeps and may flake on shared CI runners. Consider an injectable clock/event.

davidberenstein1957:
Made the changes in 61bc10d7: timed-out stop() now warns and blocks a second loop.

Not done: making the final flush wait on the in-flight tick (needs a lock shared between scheduler and tracker, bigger change); the pre-existing start_task restart and the real-sleep tests are left as is.

PR #1340, opened 2026-08-12
feat(api): reuse connections, add timeouts, and stamp emissions with measurement time

davidberenstein1957:
Part of #1338.

`ApiClient._request` called bare `requests.get/post/patch` with a hardcoded `timeout=2`, no `Session` and no retry policy, and `http.py:_emit` logged failures away. This adds a `Session` with a jittered `Retry` adapter, configurable timeouts, and a backoff guard around run creation.

**Stacked on #1341** — this branch is rebased on `scaling/01-emission-timestamps` and its base is set to that branch, so #1341 merges first and this PR's diff stays limited to its own two commits. Both touched `api_client.py`; the conflict (the imports, and one function added next to the other) was resolved in the rebase by keeping both sides. Retarget to `master` once #1341 lands.

## Measured against a loopback stub server (no real network)

| Scenario | before | after |
|---|---|---|
| 50 sequential emission POSTs | 50 TCP connections, 0.085s | **1 connection**, 0.023s |
| row hits two transient 503s | 1 request, **0/1 rows delivered** | 3 requests, **1/1 delivered** |
| endpoint refusing connections | 0.00s | 1.73s |
| endpoint hung, emission POST | 2.00s | **5.0s** (one read timeout, not retried) |
| endpoint unreachable, emission POST | 2.00s | 10.6s (3 connect timeouts + backoff) |
| endpoint hung, CLI GET path | 2.00s | 31.1s |

The second row is the point: a transient 503 used to lose a measurement permanently. Two caveats stated plainly — the wall-clock win is over plain-HTTP loopback with no TLS handshake to save, so the connection count is the real measurement and the HTTPS-over-WAN latency benefit is inferred from it; and failure paths genuinely got *slower*, which is the deliberate trade for delivering the row.

## POST is not retried once the request has landed

`carbonserver` has no idempotency key on `POST /emissions`, and the dashboard **sums** emission rows. So replaying a POST whose response was lost *after* the server committed the insert silently inflates a user's reported emissions — and unlike a dropped row, nothing shows it happened. For a carbon-accounting product that is the worse failure of the two.

The retry policy is therefore split:

| | GET / PATCH | POST |
|---|---|---|
| connection error, connect timeout | retry | retry |
| 429, 502, 503 | retry | retry |
| read timeout, truncated response | retry | **no** |
| 500, 504 | retry | **no** |

The POST column is exactly the set of failures where the request plausibly never reached the application. Widening it needs a server-side idempotency key first.

## Timing budget (the two numbers that were inconsistent)

The earlier "16s worst case" and the docstring's `(3.05 + 5) * 3` were describing different things and neither said which. Now stated separately, for the emission path's `(3.05, 5)` / `retries=2` budget:

- **hung endpoint** (connects, never answers): **~5s**. Only the read times out, and POSTs no longer retry that. Was 16.5s.
- **unreachable endpoint** (every connect times out): **~11s**, i.e. `3.05 * 3` plus jittered backoff.
- **full chain worst case** (connect timeouts until the last attempt connects and then hangs): **~16s**.

All well under the tracker's own 45s stale-measurement warning at `emissions_tracker.py`. The CLI path keeps `(3.05, 10)` and the broad GET policy, so its hung-endpoint case is ~31s — also under 45s, and it is not on the measurement loop.

## Interaction with #1339 (scheduler)

This send runs inline on the scheduler thread, so a 5–16s blocking call is 5–16s the next tick is not taken. Today's chained-`Timer` scheduler would let a slow send overlap the next measurement; once #1339 lands and ticks run on one thread, the same blocking call produces a **skipped** measurement instead of an **overlapping** one. Skipping is the better failure — it is why the budget is kept under the 45s warning rather than merely bounded — but the two PRs should be read together.

## Choices that differ from the original proposal

- **Tighter defaults than proposed.** The proposal wanted `timeout=10, retries=3` on the emission path; measured, that blocks a scheduler thread for 44s against a 15s tick. This uses `(3.05, 5)/retries=2` for emissions and `(3.05, 10)/retries=2` for the CLI.
- **The run-creation herd guard is mandatory, not optional.** With retries, a down API means three attempts per tick per client without it. The guard lives inside `ApiClient._create_run`, so both callers route through one fix rather than two.
- **Spill buffer skipped**, agreeing with the proposal's own recommendation: retries cover transient failures, and off-by-default disk-write code on the measurement path is code nobody exercises until it's needed and broken.

## Configuration

`api_timeout` (default 5, read timeout in seconds) and `api_retries` (default 2) are documented in `docs/how-to/configuration.md`, including which failures are retried on POST and why.

## Tests

`tests/test_api_client_retry.py`, 14 tests: retry-then-succeed on 503, give-up after N, no-retry on 4xx, retry on connection error, connection reuse, timeout pass-through, both run-creation backoff directions, and four pinning the POST policy — read timeout on POST is **not** retried (1 request) while a 503 on POST still is (3 requests), 504 on POST is not retried, and a read timeout on GET still is. Plus `test_api_timeout_and_retries_reach_the_api_client` in `tests/test_emissions_tracker.py`, asserting the two config knobs survive the constructor → config → `ApiClient` path. Reverting either change fails these.

They use a stdlib loopback server rather than `requests_mock` **on purpose**: `requests_mock` replaces the transport adapter, so it never executes the `HTTPAdapter`/`Retry` layer under test — a `requests_mock` retry test would pass while testing nothing.

One thing to flag honestly: `tests/test_emissions_tracker_flush.py` and `tests/test_logging_output.py` flake on this machine (flush produces 0 or 2 rows instead of the expected 1 or 3). Verified pre-existing: it reproduces on `master` and on this branch's parent commit at the same or higher rate, and it is timing-driven, so #1339 is the PR that touches it.

## Not included

Aligning the direct `requests.get` calls in `electricitymaps_api`/`geography.py` with the same policy, and a delivery-failure counter in the final log line. Both separate concerns.

### DIFF
--- a/codecarbon/core/api_client.py
+++ b/codecarbon/core/api_client.py
@@ -8,7 +8,7 @@ TODO : use async call to API
 # from httpx import AsyncClient
 import dataclasses
 import json
-from datetime import timedelta, tzinfo
+from datetime import datetime, timedelta, tzinfo
 
 import requests
 
@@ -31,6 +31,26 @@ def get_datetime_with_timezone():
     import arrow
 
     return str(arrow.now().isoformat())
+
+
+# (connect, read) seconds, replacing a flat 2s that timed out on a loaded API.
+_TIMEOUT = (3.05, 10)
+
+
+def _measurement_timestamp(carbon_emission: dict) -> str:
+    """
+    Offset-aware ISO timestamp of *when the measurement was taken*, taken from
+    EmissionsData.timestamp. Falls back to now for hand-built payloads that
+    carry no usable timestamp.
+    """
+    try:
+        return (
+            datetime.fromisoformat(carbon_emission["timestamp"])
+            .astimezone()
+            .isoformat()
+        )
+    except (KeyError, TypeError, ValueError):
+        return get_datetime_with_timezone()
 
 
 class ApiClient:  # (AsyncClient)
@@ -58,6 +78,8 @@ class ApiClient:  # (AsyncClient)
         :create_run_automatically: If False, do not create a run. To use API in read only mode.
         """
         # super().__init__(base_url=endpoint_url) # (AsyncClient)
+        # A Session so the socket and TLS handshake are reused across calls.
+        self._session = requests.Session()
         self.url = endpoint_url
         self.experiment_id = experiment_id
         self.api_key = api_key
@@ -80,15 +102,19 @@ class ApiClient:  # (AsyncClient)
         Call the API and return the response, raising on anything that is not
         the status code the API answers on success.
 
-        :method: the requests function to call, for example requests.get
+        :method: the session function to call, for example self._session.get
         :payload: the JSON body to send, if any
         :expected_status: the http code the API returns when the call succeeds
         """
         headers = self._get_headers()
-        response = method(url=url, json=payload, timeout=2, headers=headers)
+        response = method(url=url, json=payload, timeout=_TIMEOUT, headers=headers)
         if response.status_code != expected_status:
             self._raise_api_error(url, payload or {}, response)
         return response
+
+    def close(self):
+        """Release the pooled sockets. Safe to call more than once."""
+        self._session.close()
 
     def set_access_token(self, token: str):
         """This method sets the access token to be used for the API.
@@ -102,14 +128,14 @@ class ApiClient:  # (AsyncClient)
         Check API access to user account
         """
         url = self.url + "/auth/check"
-        return self._request(requests.get, url).json()
+        return self._request(self._session.get, url).json()
 
     def get_list_organizations(self):
         """
         List all organizations
         """
         url = self.url + "/organizations"
-        return self._request(requests.get, url).json()
+        return self._request(self._session.get, url).json()
 
     def check_organization_exists(self, organization_name: str):
         """
@@ -134,7 +160,7 @@ class ApiClient:  # (AsyncClient)
             return organization
         else:
             return self._request(
-                requests.post, url, payload=payload, expected_status=201
+                self._session.post, url, payload=payload, expected_status=201
             ).json()
 
     def get_organization(self, organization_id):
@@ -142,7 +168,7 @@ class ApiClient:  # (AsyncClient)
         Get an organization
         """
         url = self.url + "/organizations/" + organization_id
-        return self._request(requests.get, url).json()
+        return self._request(self._session.get, url).json()
 
     def update_organization(self, organization: OrganizationCreate):
         """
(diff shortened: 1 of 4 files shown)


benoit-cty:

## Verdict: ✅ Approve with nits (please merge #1339 first)

Session reuse (with `close()` wired into `CodeCarbonAPIOutput.exit()`), the `(3.05, 10)` connect/read timeouts, and stamping emissions with `EmissionsData.timestamp` all look good. 80 tests pass, and the loopback test that counts connections is nice. There are no retries at all, so there is no risk of a replayed POST creating duplicate emissions.

**Issues:**
1. **The description doesn't match the code.** It describes:
   - a `Retry` adapter with a split POST vs GET/PATCH policy
   - `api_timeout` / `api_retries` config, and docs for them
   - a run-creation backoff guard
   - `tests/test_api_client_retry.py`
   - "stacked on #1341" (now closed)

   None of these are in the diff. Please update the description, or add the missing parts.
2. **A hung API now blocks the measurement thread much longer** (`api_client.py:37`, `output_methods/http.py:62-68`).
   - When `run_id` is None, `_emit` calls `_ensure_api_run()` and then `add_emission()`, and each can call `_create_run`.
   - Against a hung API, one `live_out` can block the scheduler thread for about 2 × 13 s ≈ **26 s**, versus about 4 s on master. That is longer than the default 15 s tick.
   - Suggested fix: at most one `_create_run` attempt per `_emit`, plus a cooldown after a failed run creation (the "backoff guard" from the description).
   - Together with #1339 (5 s join cap in `stop()`), this also makes it more likely that the final flush overlaps an in-flight tick.

**Nit:**
- `_measurement_timestamp` (`api_client.py:40`) takes a naive, second-precision local time and calls `.astimezone()` on it. In the repeated hour at the DST fall-back, the offset can be wrong by an hour. Creating a timezone-aware timestamp at the source (`emissions_tracker.py:1076`, e.g. `datetime.now().astimezone()`) avoids that.

davidberenstein1957:
Made the changes in 72c53b77: 60 s cooldown after a failed run creation.

On the 2 x 13 s concern: `_create_run` re-raises every failure, so `_emit` stops after the first attempt; with the cooldown a down API now costs one ~13 s call per minute, not per tick. Stays compatible with #1339 (no shared files).

Not done: rewriting the description (my edit was blocked; the retry/config/backoff sections are stale, the diff is Session + timeouts + measurement timestamp + cooldown). Skipped the DST nit: making the tracker timestamp offset-aware changes the CSV timestamp format, which needs a maintainer call.

davidberenstein1957:
Deferring the DST timestamp nit: making the timestamp timezone-aware at the source changes the CSV timestamp format, so it will ship in the next minor release together with #1334 rather than in this PR.

PR #1343, opened 2026-08-12
perf(powermetrics): drop numpy for statistics.fmean/math.fsum

davidberenstein1957:
Part of #1338.

> **Base branch**: stacked on `fix/powermetrics-nan-totals` (#1345), not `master`, because it touches the same `ApplePowermetrics.get_details()` block. **Retarget to `master` once #1345 merges.**

## This PR was split

It originally removed **both** numpy and pandas from the default install. The pandas half reimplements `read_csv`/`dropna` semantics by hand across 14 files — the `Unnamed: <i>` column reconstruction in `core/cpu.py`, ragged-row filtering, and a `csv.DictWriter` rewrite of the `emissions.csv` writer. That is the whole regression risk of the change and it should not ride along with a three-line one.

- **This PR** now contains only the **numpy removal**: `core/powermetrics.py`.
- **The pandas removal** moved to #1391 (`scaling/03b-drop-pandas`), stacked on this branch so its diff shows only the pandas work.

## What is left here

`import numpy as np` at the top of `codecarbon/core/powermetrics.py` was the only numpy use in the measurement core, and **numpy was never a declared dependency** — it arrived transitively through pandas. It cost ~20 ms of the ~78 ms cold `import codecarbon`, on every platform, for two calls on a short list of floats:

```python
details[f"{chip_part} Power"] = statistics.fmean(watts)
details[f"{chip_part} Energy Delta"] = math.fsum(
    (self._interval / 1000) * watt for watt in watts
)
```

`fmean`/`fsum` are exactly rounded where numpy uses pairwise summation, so results differ in the last ulp only. `test_get_details` was asserting exact equality against numbers produced by numpy's summation order; it is now `pytest.approx`, which is what it always meant.

`codecarbon/viz/components.py` still uses numpy; it is behind the `carbonboard`/`viz-legacy` extras, which install pandas (and therefore numpy). No dependency metadata changes here — that is #1391's job.

`#1345`'s warn-once guard on the missing-samples warning is preserved (an earlier revision of this branch had reverted it).

## Tests

`uv run pytest tests/ -q --ignore=tests/test_viz_data.py` → **629 passed, 21 skipped**. (`test_viz_data.py` fails to collect on `master` identically — `dash` is not installed.)

### DIFF
--- a/codecarbon/core/api_client.py
+++ b/codecarbon/core/api_client.py
@@ -22,6 +22,11 @@ from codecarbon.core.schemas import (
 from codecarbon.external.logger import logger
 
 
+def _round_or_none(value: float | None) -> float | None:
+    """Round a coordinate, keeping None when it is unknown."""
+    return None if value is None else round(value, 1)
+
+
 def get_datetime_with_timezone():
     import arrow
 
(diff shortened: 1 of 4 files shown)


davidberenstein1957:
Closing as redundant. The numpy to statistics.fmean/math.fsum swap is three lines touching the same file as #1345, so it now lives there as an additional commit rather than as a separate PR that has to be reviewed and merged in order. #1345 carries both the NaN guard and the numpy drop.

PR #1344, opened 2026-08-12
fix: count DRAM in RAPL energy total

### LINKED ISSUE #1305: rapl_include_dram=True reads DRAM domains but never adds them to the energy total
On Linux, `rapl_include_dram=True` selects the DRAM RAPL domains and reads them every cycle, but their energy never reaches the reported total. The measurement is silently identical to `rapl_include_dram=False`, while the log claims DRAM is included.

## Reproduction

On a machine exposing a top-level `dram` domain alongside `package-0`:

```python
a = EmissionsTracker(rapl_include_dram=False)
b = EmissionsTracker(rapl_include_dram=True)
```

Both report the same CPU energy, even though `b` logs `RAPL - Including 1 DRAM domain(s) for complete hardware power measurement (CPU+DRAM)`.

## Root cause

Two naming conventions that were never tied together.

- `codecarbon/core/cpu.py:708` — domain *selection* correctly appends DRAM domains when `rapl_include_dram` is set.
- `codecarbon/core/cpu.py:772` — `_create_rapl_files` only assigns the Power-Gadget-compatible display name `Processor Energy Delta_N(kWh)` to domains whose name contains `package` or `psys`. A DRAM domain falls into the `else` and keeps its raw sysfs name `dram`.
- `codecarbon/external/hardware.py:390` — `_get_energy_from_cpus` sums only keys matching `^Processor Energy Delta_\d`, so the `dram` entry is dropped. Same for power: `codecarbon/core/cpu.py:849` only derives a `Processor Power` key from names containing `Energy`.

Windows EMI does not have this bug: `codecarbon/core/windows_emi.py:581` names every selected channel `Processor Energy Delta_N(kWh)`, DRAM included. So `emi_include_dram=True` works and `rapl_include_dram=True` does not.

## Expected vs actual

Expected: with `rapl_include_dram=True`, reported CPU energy is package + DRAM (DRAM is typically 10-20% of package power on memory-heavy workloads).

Actual: DRAM energy is read, converted, logged, and discarded. Reported energy equals the package-only value.

Note this is a measurement change for anyone already setting the flag — their reported figures will (correctly) increase.

davidberenstein1957:
On Linux, `rapl_include_dram=True` selected DRAM domains and read them every cycle, but their energy never reached the reported total — the result was identical to `rapl_include_dram=False` while the log claimed DRAM was included.

## What changed

`codecarbon/core/cpu.py` — `_create_rapl_files` now assigns the Power-Gadget-compatible display name `Processor Energy Delta_N(kWh)` to DRAM domains as well as package/psys. Previously DRAM fell into the `else` branch and kept its raw sysfs name `dram`, which the aggregator in `codecarbon/external/hardware.py` (`^Processor Energy Delta_\d`) filters out.

DRAM domains only reach `_create_rapl_files` when `rapl_include_dram` is true — they are filtered out earlier otherwise — so the default path is unchanged. This also makes Linux match Windows EMI, which already names every selected channel this way (`codecarbon/core/windows_emi.py:581`) and so has never had the bug.

## How it was verified

Added `test_rapl_include_dram_energy_is_aggregated` in `tests/test_rapl_parameters.py`, parametrized over both flag values. It builds a fake sysfs tree with `package-0` and `dram`, advances both counters, calls `get_cpu_details()`, and asserts the summed `Processor Energy Delta_*` value. With `rapl_include_dram=True` it expects package + DRAM; it fails on master (package-only) and passes with this change. The `False` case pins the default path.

`test_rapl_include_dram_true_explicit` was updated: it previously asserted a RAPL file was still named `dram`, which is exactly the behaviour being fixed. It now asserts both domains carry aggregated names.

The suite is gated on Linux and CI will exercise it; locally (macOS) the tests were driven directly with `sys.platform` patched, and all tests in `test_rapl_parameters.py` and `test_rapl_mmio_scanning.py` pass.

## Note for the changelog

This is a measurement change: anyone already setting `rapl_include_dram=True` will see reported CPU energy increase (correctly). DRAM is typically 10-20% of package power on memory-heavy workloads.

Closes #1305

### DIFF
--- a/codecarbon/core/cpu.py
+++ b/codecarbon/core/cpu.py
@@ -770,8 +770,9 @@ class IntelRAPL:
             domain_name,
         ) in domain_map.values():
             try:
-                if domain_name and (
-                    "package" in domain_name.lower() or "psys" in domain_name.lower()
+                domain_lower = (domain_name or "").lower()
+                if any(
+                    keyword in domain_lower for keyword in ("package", "psys", "dram")
                 ):
                     display_name = f"Processor Energy Delta_{domain_index}(kWh)"
                     domain_index += 1
--- a/tests/test_rapl_parameters.py
+++ b/tests/test_rapl_parameters.py
@@ -3,11 +3,13 @@ Test to verify that rapl_include_dram and rapl_prefer_psys parameters work corre
 """
 
 import logging
+import re
 import sys
 
 import pytest
 
 from codecarbon.core.cpu import IntelRAPL
+from codecarbon.core.units import Time
 
 
 @pytest.mark.skipif(not sys.platform.lower().startswith("lin"), reason="requires Linux")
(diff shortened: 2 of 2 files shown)


benoit-cty:

## Verdict: 🔧 Request changes

**The bug is real.** DRAM domains keep the raw name `dram`, and `_get_energy_from_cpus` / `_get_power_from_cpus` only sum `^Processor Energy Delta_\d` / `^Processor Power`. So on master, `rapl_include_dram=True` does nothing. Renaming DRAM to `Processor Energy Delta_N` fixes that, but it collides with code that landed on master after this branch was created.

**Must fix:**
1. **It breaks master's mirror detection, and a master test fails once merged.**
   - `IntelRAPL.start()` (`codecarbon/core/cpu.py` ~L890-895) leaves DRAM out of mirror detection with `if "dram" not in rapl_file.name.lower()`. After the rename, DRAM files are called `Processor Energy Delta_N(kWh)`, so the guard never matches.
   - On the PR merged with master, `tests/test_rapl_mmio_scanning.py::test_rapl_start_keeps_dram_when_it_matches_a_package_counter` fails: DRAM gets flagged as a mirror of a package counter and dropped.
   - Fix: base the exclusion on the domain type, not the display name. For example, keep `domain_name` on `RAPLFile` or add an `is_dram` flag. Also update that test's `assert "dram" in details`.
2. **The fallback path ignores the flag.**
   - In the fallback branch of `_select_domains_to_use` ("No package or psys domains found, using all available domains"), `readable_domains` includes top-level DRAM regardless of `rapl_include_dram`.
   - With the rename, DRAM would then be summed as CPU energy **even with the flag off**. That contradicts the description.
   - Please filter DRAM by the flag there too. Add a test next to `test_rapl_non_power_domain_keeps_its_own_name`, which only covers `core`.

**Worth documenting:**
- With the flag on, memory power ends up in both `cpu_energy` (RAPL DRAM) and `ram_energy` (the RAM estimate). `docs/how-to/configuration.md` says this is by design, but it should go in the changelog.
- DRAM exposed as a package subzone (`intel-rapl:N:M`) is still skipped, so the flag stays a no-op there. That's pre-existing; maybe open an issue.
- No psys+DRAM double count: DRAM is only added in the package branch.

davidberenstein1957:
Made the changes in c2cc54b8: merged master, DRAM mirror exclusion now uses a `RAPLFile.is_dram` flag (master's DRAM test updated and passing), and the fallback path respects `rapl_include_dram` (new test). Added the CPU+RAM double-count and subzone notes to the description's changelog section. Didn't open the subzone issue; that's for a maintainer to decide.

PR #1345, opened 2026-08-12
fix(powermetrics): reject unsupported Macs and report 0 W when no samples

### LINKED ISSUE #1306: Empty powermetrics output yields NaN power, poisoning all downstream totals on Apple Silicon
When `powermetrics` produces no matching output, `ApplePowermetrics.get_details()` returns `nan` for CPU/GPU power. NaN is absorbing under addition, so a single such sample turns `_total_energy`, `emissions` and `emissions_rate` into `nan` for the remainder of the run. The CSV row and the API payload then contain `NaN`, which also makes the JSON invalid under the strict spec.

## Reproduction

```python
from codecarbon.core.powermetrics import ApplePowermetrics

pm = ApplePowermetrics(output_dir=tmpdir, log_file_name="empty.txt")
open(pm._log_file_path, "w").close()   # sudoers rule missing, or sampler emitted nothing
d = pm.get_details()
assert math.isnan(d["CPU Power"])      # passes today
```

## Root cause

`codecarbon/core/powermetrics.py:181` and `:193` call `np.mean([...])` on the result of `re.findall`. `np.mean([])` returns `nan` and only emits a `RuntimeWarning`; it does not raise, so the `except Exception` at `codecarbon/core/powermetrics.py:201` never fires and nothing is logged.

The value flows straight through `AppleSiliconChip._get_power` (`codecarbon/external/hardware.py:504-515`) into `Power.from_watts`, then through `BaseHardware.measure_power_and_energy` into the tracker accumulators.

The GPU branch is the likelier trigger: `_log_values` runs powermetrics with `--samplers cpu_power` only (`codecarbon/core/powermetrics.py:145-153`), and whether that sampler emits `GPU Power:` lines depends on the machine and macOS version. Where it does not, `gpu_power_list` is empty on every call.

## Expected vs actual

Expected: no samples means 0 W with a visible warning, and the run still reports finite emissions.
Actual: no samples means `nan`, silently, and every total for the rest of the run is `nan`.

davidberenstein1957:
## What

`ApplePowermetrics.get_details()` now guards against an empty match list before averaging: if `powermetrics` produced no `CPU Power:` / `GPU Power:` lines it logs a warning and reports `0.0` W instead of `nan`. The CPU and GPU branches were identical, so they are now one loop.

## Why

`np.mean([])` returns `nan` and only emits a `RuntimeWarning`, so the surrounding `except Exception` never fired and nothing was logged. The NaN flowed through `AppleSiliconChip._get_power` into the tracker accumulators, and since NaN is absorbing under addition, one bad sample made `energy_consumed`, `emissions` and `emissions_rate` NaN for the remainder of the run — written to `emissions.csv` and POSTed to the API as bare `NaN`.

This is reachable in the field whenever the powermetrics sudoers rule is missing (empty log), and on machines where the `cpu_power` sampler emits no `GPU Power:` lines, where it happened on every single call.

Reporting 0 W understates rather than invalidates, and the warning makes the condition visible.

## Verification

Two new cases in `tests/test_powermetrics.py` (`test_get_details_without_samples`, `test_get_details_without_gpu_samples`) both fail on master and pass with this change. `uv run pytest tests/test_powermetrics.py -q` -> 18 passed. The existing `test_get_details` is unchanged, confirming the non-empty path still behaves identically.

Note: `uv run task lint` reports pre-existing violations across the repo and `uv run task format` reformats ~120 unrelated files, so neither was applied wholesale; `black --check` is clean on the two touched files.

Closes #1306

### DIFF
--- a/codecarbon/core/powermetrics.py
+++ b/codecarbon/core/powermetrics.py
@@ -1,13 +1,12 @@
 import os
 import re
 import shutil
+import statistics
 import subprocess
 import sys
 import time
 from functools import lru_cache
 from typing import Dict
-
-import numpy as np
 
 from codecarbon.core.util import detect_cpu_model
 from codecarbon.external.logger import logger
@@ -118,17 +117,19 @@ class ApplePowermetrics:
         """
         Setup cli command to run Powermetrics
         """
-        if self._system.startswith("darwin"):
-            cpu_model = detect_cpu_model()
-            if cpu_model.startswith("Apple"):
-                if shutil.which(self._osx_silicon_exec):
-                    self._cli = self._osx_silicon_exec
-                else:
-                    raise FileNotFoundError(
-                        f"Powermetrics executable not found on {self._system}"
-                    )
-        else:
+        if not self._system.startswith("darwin"):
             raise SystemError("Platform not supported by Powermetrics")
+        cpu_model = detect_cpu_model() or ""
+        if not cpu_model.startswith("Apple"):
+            raise SystemError(
+                "Powermetrics is only supported on Apple Silicon, "
+                f"detected CPU: {cpu_model!r}"
+            )
+        if not shutil.which(self._osx_silicon_exec):
+            raise FileNotFoundError(
+                f"Powermetrics executable not found on {self._system}"
+            )
+        self._cli = self._osx_silicon_exec
 
     def _log_values(self) -> None:
         """
@@ -140,10 +141,9 @@ class ApplePowermetrics:
             # Run the powermetrics command with sudo and capture its output
             cmd = [
                 "sudo",
-                "powermetrics",
+                self._cli,
                 "-n",
                 str(self._n_points),
-                "",
                 "--samplers",
                 "cpu_power",
                 "--format",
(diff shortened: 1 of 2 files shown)


benoit-cty:

## Verdict: 🔧 Request changes (rebase needed)

Both fixes are real:
- `np.mean([])` gives NaN, which then poisons every total.
- On an Intel Mac without Power Gadget, master selects powermetrics. Intel output has no `CPU Power: X mW` line, so every sample is NaN.

Making `_setup_cli` raise `SystemError` for non-Apple CPUs correctly falls back to TDP/cpu_load. Removing the stray `""` argument and using `self._cli` is good too.

**Must fix:**
1. **Merge conflicts with master** in `codecarbon/core/powermetrics.py` and `tests/test_powermetrics.py`.
   - Master commit 1ea5f934 rewrote `_log_values` to use `Popen` with a timeout, and added `if not self._log_values(): return dict()` in `get_details`.
   - The rebase must keep master's timeout and early return.
   - `test_log_values_builds_clean_command` mocks `subprocess.call`, which master no longer uses, so it needs rewriting against `Popen`.
2. **No warning is logged, despite the description.**
   - `get_details` falls back to `0.0` silently when there are no CPU samples.
   - In powermetrics mode there is no further fallback, so a missing sudoers rule or an empty log reports **0 W CPU for the whole run** with nothing in the logs.
   - Please add a `logger.warning`, once, that points to the powermetrics/sudoers troubleshooting docs.

**Nits:**
- After this change, the Intel-Mac `elif powermetrics.is_powermetrics_available()` branch in `core/resource_tracker.py` (~L237) is dead code. Remove it or comment it.
- There are now two different Apple Silicon checks: `startswith("Apple")` here and `is_mac_arm` (`\bM\d`) in `resource_tracker`. Please use one helper.
- The title mentions rejecting unsupported Macs, but the description only covers the empty-sample fix. Please describe both.

davidberenstein1957:
Made the changes in d537043b: merged master (kept the Popen timeout and early return, rewrote the command test against Popen), added a one-time warning for empty CPU samples, switched to `is_mac_arm`, removed the dead Intel-Mac branch, and updated the description.

PR #1357, opened 2026-08-12
feat(cli): add `codecarbon badge` command

### LINKED ISSUE #1349: Feature proposal: `codecarbon badge` — generate a README badge from a local emissions.csv
This is a feature proposal, offered for discussion before implementation.

## The user problem

Once a run finishes, everything CodeCarbon measured lives in `emissions.csv` on the machine that produced it. Nobody else ever sees it, and the person who ran it rarely looks again. The cheapest way to make a measurement stick is to put it where open-source projects already curate their signals: the README, next to the CI and coverage badges. A visible number is also what makes maintainers re-measure after a change.

Today there is no CLI command that summarises or reports on an existing `emissions.csv` at all — the CLI only does config, auth, live monitoring and hardware detection.

## Proposed design

A single new command that reads the CSV and writes badge artefacts locally. No network calls, no hosted service, no account:

```
codecarbon badge [OPTIONS]

  --file PATH                  emissions.csv to read        [default: ./emissions.csv]
  --project TEXT               filter rows by project_name  [default: all rows]
  --select [last|mean|total]   which run(s) to report       [default: last]
  --output-dir PATH            where to write               [default: .]
  --label TEXT                 left-hand badge text         [default: "carbon"]
  --color TEXT                 badge colour                 [default: grey]
  --format [svg|json|markdown|all]                          [default: all]
```

It writes `codecarbon-badge.svg` (a flat shields-style badge rendered from a small string template) and `codecarbon-badge.json` (the documented shields.io endpoint schema, for users who want to publish it themselves, e.g. on GitHub Pages), then prints the markdown snippet to paste into a README.

Supporting module `codecarbon/badge.py` holds the whole feature: CSV load, aggregation, unit formatting, SVG template, JSON writer.

## Why it fits

- The CLI is a plain typer app (`codecarbon/cli/main.py`); this is one more `@codecarbon.command` alongside `detect`.
- The data source is the existing `FileOutput` CSV, whose schema is `EmissionsData` — no new output method, no tracker changes.
- No new dependencies: stdlib `csv`, `json` and string templating are enough. Text widths are estimated from character count, deliberately avoiding a font-metrics dependency.

## Scope boundary, and one honest caveat

A badge turns a private measurement into a public claim, and CodeCarbon numbers are often estimates (TDP-modelled CPU power, country-average carbon intensity). Two consequences for the design:

- **Colour-neutral by default.** No automatic green/orange/red thresholds. A green badge on a large model is greenwashing, and a fixed gram threshold means nothing across wildly different workloads. Colour only if the user explicitly asks for one.
- **The badge reports what the CSV says, nothing more.** No trend graphs, no history, no "improvement" claims.

Also deliberately out of scope: a badge output method that fires on every tracker stop (the CSV is the source of truth; regenerate on demand), any hosted endpoint or ingestion service, and a GitHub Action (that is a few lines of YAML in the docs, not code in this repo).

Feedback on the CLI surface and on the colour decision especially welcome before this is built out.

davidberenstein1957:
Implements #1349.

## What it adds

A `codecarbon badge` command that turns an existing `emissions.csv` into a README badge. Strictly local: it reads the CSV, writes one JSON file, and prints a snippet. No network call, no hosted service, no account, no new dependency (stdlib `csv` and `json` only).

## User-facing surface

```
codecarbon badge [--file PATH] [--project TEXT] [--select last|mean|total] [--output-dir PATH]
```

It writes `codecarbon-badge.json`, the [shields.io endpoint](https://shields.io/badges/endpoint-badge) schema, then prints the markdown to paste into a README once that file is published at a public URL. shields.io renders the image; CodeCarbon does not reimplement badge drawing.

The logic lives in `codecarbon/badge.py` (`load_runs`, `last_row_per_run`, `summarise`, `format_value`, `message_for`, `render`, `write`), usable directly from Python. The CLI command sits next to `detect` in `codecarbon/cli/main.py`. No tracker or output-method changes.

**One row per flush, not per run.** CodeCarbon appends a row on every flush and each row holds the *cumulative* totals of its run (`FileOutput.out()` discards the delta it is handed). The badge therefore groups rows by `run_id` and keeps the last row of each run, so `total` is the sum over runs, `mean` the average per run, and `runs` a run count.

Rows are ordered by `timestamp`, not by file order: `allow_multiple_runs` defaults to `True`, so two trackers can interleave their flushes into a single `emissions.csv` and the run that appears first is not necessarily the run that finished last. Rows with a blank or missing `run_id` (older CSVs) each count as their own run rather than collapsing into one.

**Colour-neutral, no knobs.** No `--label`, `--color` or `--metric`. CodeCarbon numbers are frequently estimates (TDP-modelled CPU power, country-average intensity) and no fixed gram threshold means anything across arbitrary workloads, so an automatic green badge would be greenwashing; the badge is grey and says grams. Anyone who wants different text or colour is already editing a JSON file they self-host. `--select` is an `enum.Enum`, so typer validates it and `--help` lists the choices.

## Verification

`tests/test_badge.py`, 18 tests, no network, `tmp_path` for outputs: project filtering, the three aggregations against hand-computed values, a multi-flush CSV proving cumulative rows are not double-counted, an interleaved-`run_id` CSV proving `last` follows the timestamp rather than file order, blank `run_id`s counting separately, signed formatting, unit formatting from mg to Mg, endpoint JSON schema, clean errors on a missing file and an empty project selection, and three `typer.testing.CliRunner` runs against the real command including the rejection of an invalid `--select`.

```
uv run pytest tests/test_badge.py -q   # 18 passed
```

`black` is clean on the touched files; the remaining `ruff` notes are pre-existing repo-wide style patterns (`Optional[...]`, `typer.Option` defaults) that the surrounding code already uses.

## Deliberately left out

- **Local SVG rendering.** An earlier revision shipped a hand-rolled shields-style SVG template plus a character-advance font-metrics guess for text width. That is a shields.io reimplementation the maintainers would own forever. Deleted; the endpoint JSON is the feature, and `json.dumps` escapes correctly for free.
- Colour thresholds and per-badge cosmetics (see above).
- The four `badge_*` config keys from the proposal. Flags only for now.
- A badge output method firing on every tracker stop. The CSV is the source of truth, regenerate on demand.
- Any hosted endpoint or ingestion service, badge history/trend graphs, and a GitHub Action.

Docs: a `codecarbon badge` section in `docs/reference/cli.md` and a cross-link from `docs/how-to/visualize.md`.

Closes #1349

### DIFF
--- a/codecarbon/badge.py
+++ b/codecarbon/badge.py
@@ -1,10 +1,9 @@
 """
 Generate a README badge from an existing ``emissions.csv``.
 
-Everything here is local: read the CSV written by ``FileOutput``, render a flat
-SVG badge and a `shields.io endpoint
-<https://shields.io/badges/endpoint-badge>`_ JSON file. No network call, no
-hosted service, no account.
+Everything here is local: read the CSV written by ``FileOutput`` and render a
+`shields.io endpoint <https://shields.io/badges/endpoint-badge>`_ JSON file.
+No network call, no hosted service, no account.
 """
 
 import csv
@@ -19,27 +18,6 @@ BADGE_STEM = "codecarbon-badge"
 # The badge is deliberately colour-neutral by default. CodeCarbon numbers are
 # often estimates, and a green badge on a large model would be greenwashing:
 # no gram threshold is meaningful across arbitrary workloads.
-
-_SVG_TEMPLATE = """<svg xmlns="http://www.w3.org/2000/svg" width="{width}" height="20" role="img" aria-label="{label}: {message}">
-  <title>{label}: {message}</title>
-  <linearGradient id="s" x2="0" y2="100%">
-    <stop offset="0" stop-color="#bbb" stop-opacity=".1"/>
-    <stop offset="1" stop-opacity=".1"/>
-  </linearGradient>
-  <clipPath id="r"><rect width="{width}" height="20" rx="3" fill="#fff"/></clipPath>
-  <g clip-path="url(#r)">
-    <rect width="{label_width}" height="20" fill="#555"/>
-    <rect x="{label_width}" width="{message_width}" height="20" fill="{color}"/>
-    <rect width="{width}" height="20" fill="url(#s)"/>
-  </g>
-  <g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" font-size="11">
-    <text x="{label_x}" y="15" fill="#010101" fill-opacity=".3">{label}</text>
-    <text x="{label_x}" y="14">{label}</text>
-    <text x="{message_x}" y="15" fill="#010101" fill-opacity=".3">{message}</text>
-    <text x="{message_x}" y="14">{message}</text>
-  </g>
-</svg>
-"""
 
 
 def load_runs(emissions_file, project: Optional[str] = None) -> List[Dict]:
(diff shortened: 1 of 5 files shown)


davidberenstein1957:
Closing. This writes a shields.io endpoint JSON that the user then has to host somewhere themselves, so it does not actually deliver a badge end to end. Vanity surface with permanent maintenance cost and no demand behind it.

Issue #1354, opened 2026-08-12
Pluggable live carbon-intensity providers (with caching and backoff)


davidberenstein1957:
*Feature proposal.*

## The problem

Carbon intensity is half of CodeCarbon's output — emissions are energy × intensity. We measure energy carefully, but intensity is a yearly national average from a bundled JSON file for everyone except users who hold an Electricity Maps token. In a country like France or Germany real grid intensity moves by a factor of three within a day, so a run at 03:00 and the same run at 19:00 report identical numbers. That is a systematic error in the headline figure.

For the users who *do* have a token, the current live path has concrete defects. `codecarbon/core/electricitymaps_api.py` is the whole of live-intensity support, and its single caller is `EmissionsCalculator.get_private_infra_emissions` in `codecarbon/core/emissions.py`:

- **No caching.** `get_emissions()` is called on every emissions computation. A long run with a short `measure_power_secs` issues thousands of HTTP requests for a value the grid publishes hourly.
- **No backoff.** A wrong or expired token produces one failed request and one `logger.error` line per measurement tick for the entire run. The call site does catch `Exception` and fall through to bundled data, so the run survives — but it degrades silently and noisily.
- **No visibility.** Nothing in the output says whether a number came from a live API or from a yearly average, so silent fallback is undetectable after the fact.
- **One hardcoded vendor.** The URL and auth scheme are baked into core. There is no way to use ENTSO-E (free with registration, all of Europe), WattTime, or a utility's own feed.

## Proposed design

A small provider seam plus a resolver, sitting exactly where the existing "give me an intensity in gCO2e/kWh" seam already is — `get_private_infra_emissions` already honours `self._force_carbon_intensity_g_co2e_kwh` before anything else, which is proof the abstraction belongs at that level.

```python
@dataclass(frozen=True)
class CarbonIntensity:
    g_co2e_per_kwh: float
    source: str            # "electricitymaps" | "entsoe" | "watttime" | "static"
    zone: str | None
    measured_at: datetime | None
    is_live: bool

class IntensityProvider(Protocol):
    name: str
    def available(self) -> bool: ...       # pure config check, no network
    def latest(self, geo: GeoMetadata) -> CarbonIntensity: ...

def resolve_intensity(geo, conf) -> CarbonIntensity:
    """Never raises. Falls back to bundled static data."""
```

[...]

PR #1358, opened 2026-08-12
feat: cache and back off Electricity Maps carbon intensity

davidberenstein1957:
**Slice 1 of 5** of the pluggable carbon-intensity provider work described in #1354. It stands on its own: it fixes the two live defects in the current Electricity Maps path without introducing any new abstraction, config key, or output field.

## What is in this slice

`codecarbon/core/electricitymaps_api.py` fetched the grid carbon intensity on *every* emissions computation. On a long run that is a lot of HTTP requests for a value the grid publishes hourly at best, and when the token is wrong or the network is down, every one of them is doomed and logs an error.

To be precise about the frequency, since the earlier version of this description overstated it: during a run, `_prepare_emissions_data()` → `_update_emissions()` is called once every `api_call_interval` measures (default 8), not on every measurement tick — see `codecarbon/emissions_tracker.py:1278-1284`. It *is* once per call for code that uses tasks, where `start_task()` / `stop_task()` each prepare emissions data (`codecarbon/emissions_tracker.py:778`, `:810`), and once more at `stop()`. So the wasted-request and repeated-error-line count scales with the run length divided by `api_call_interval`, or with the number of tasks — not with the tick count.

- Extract `get_carbon_intensity(geo, token) -> float` (gCO2e/kWh) out of `get_emissions()`. Asking for the intensity rather than for emissions-given-energy is the shape the provider layer needs, and it is useful on its own.
- Cache the intensity for 60 seconds, keyed by location (`lat/lon` or `countryCode`) and token, so a long run makes far fewer requests.
- After a failure, put that location+token in an exponential cooldown doubling from 30 s to a 1 hour ceiling, during which no request is issued. A successful call resets it.
- Skipped requests raise `ElectricityMapsAPICooldownError`, which `emissions.py` logs at debug level, so a bad token produces one error line instead of one per API call.
- `reset_cache()` for tests.

`get_emissions()` keeps its exact signature and return value and now delegates, so the three existing Electricity Maps test files are untouched. `codecarbon/core/emissions.py` gains one `except` clause for the cooldown error.

**Behaviour change, not a pure optimisation:** the 60 s TTL means measurements inside that window convert energy with the same intensity value rather than a freshly fetched one. The TTL is deliberately shorter than the default `api_call_interval × measure_power_secs` (~2 minutes), so the cadence the maintainers configured is what actually drives API calls — the cache only collapses the extra calls that tasks and `stop()` add on top. The cache is keyed by location *and* token, so two trackers in one process with different tokens neither share a value nor share a failure cooldown. Documented in `docs/how-to/configuration.md`.

## What is deferred

- **Slice 2** — `CarbonIntensity` / `IntensityProvider` in `codecarbon/core/intensity/`, with today's bundled-data branches lifted out of `emissions.py` unchanged as `StaticProvider`. No number moves.
- **Slice 3** — Electricity Maps behind the protocol, this module kept as a deprecation shim.
- **Slice 4** — `resolve_intensity()` with the fallback chain, wired into `get_private_infra_emissions()`, plus the `carbon_intensity_providers` config key and its backward-compatibility default.
- **Slice 5** — `carbon_intensity_g_co2e_kwh` and `carbon_intensity_source` on `EmissionsData`, so silent fallback to yearly averages becomes visible in the CSV and in Prometheus.
- Further out, and out of scope of #1354's first version: ENTSO-E and WattTime providers, and time-weighted intensity.

Also deliberately not in this slice: stale-serve (returning an expired cached value when the API errors). It trades a silent inaccuracy for continuity and should land together with the `is_live` / `carbon_intensity_source` reporting in slice 5, not before there is any way to see it happened.

## Tests

`tests/test_electricitymaps_cache.py`, all network mocked with `responses` following `tests/test_electricitymaps_api.py`: cache hit within TTL, a long run bounded to one request, refetch after expiry, per-location keying, no request while in cooldown, cooldown doubling to the ceiling, cooldown reset after success, and cooldown isolation between tokens.

```
uv run pytest tests/test_electricitymaps_cache.py tests/test_electricitymaps_api.py \
  tests/test_emissions.py -q
37 passed, 1 skipped
```

Draft pending review of the overall direction in #1354.

Refs #1354

### DIFF
--- a/codecarbon/core/electricitymaps_api.py
+++ b/codecarbon/core/electricitymaps_api.py
@@ -1,12 +1,105 @@
-from typing import Any, Dict
+import threading
+import time
+from typing import Any, Dict, Tuple
 
 import requests
 
 from codecarbon.core.units import EmissionsPerKWh, Energy
 from codecarbon.external.geography import GeoMetadata
+from codecarbon.external.logger import logger
+
+_Key = Tuple[Tuple[Tuple[str, Any], ...], str]
 
 URL: str = "https://api.electricitymaps.com/v3/carbon-intensity/latest"
 ELECTRICITYMAPS_API_TIMEOUT: int = 30
+
+# Grid carbon intensity is published hourly at best, while emissions are computed
+# on every measurement tick, so the value is cached instead of refetched.
+ELECTRICITYMAPS_CACHE_TTL: int = 60
+# After a failure (bad token, network down), wait before retrying instead of
+# issuing one doomed request per measurement tick.
+ELECTRICITYMAPS_COOLDOWN: int = 60
+
+# {(sorted query params, token): (monotonic fetch time, intensity in gCO2e/kWh)}
+_cache: Dict[_Key, Tuple[float, float]] = {}
+# {cache key: monotonic time until which requests are skipped}
+# Keyed like the cache: one tracker's bad token must not block another's good one.
+_cooldown: Dict[_Key, float] = {}
+# The state above is read-modify-written from the measurement thread.
+_lock = threading.Lock()
+
+
+def reset_cache() -> None:
+    """Drop the cached carbon intensities and any pending failure cooldown."""
+    with _lock:
+        _cache.clear()
+        _cooldown.clear()
+
+
+def _start_cooldown(key: _Key) -> None:
+    with _lock:
+        _cooldown[key] = time.monotonic() + ELECTRICITYMAPS_COOLDOWN
+
+
+def get_carbon_intensity(
+    geo: GeoMetadata, electricitymaps_api_token: str = ""
+) -> float:
+    """
+    Retrieve the carbon intensity of the grid, in gCO2e/kWh, from the Electricity
+    Maps API (formerly CO2 Signal) for the given geographic location.
+
+    Raises:
+        ElectricityMapsAPIError: the request failed, returned an error, or was
+            skipped because a previous one failed (``...CooldownError``).
+    """
+    if geo.latitude:
+        params: Dict[str, Any] = {"lat": geo.latitude, "lon": geo.longitude}
+    else:
+        params = {"countryCode": geo.country_2letter_iso_code}
+    key = (tuple(sorted(params.items())), electricitymaps_api_token)
+    with _lock:
+        cached = _cache.get(key)
+        cooldown_until = _cooldown.get(key, 0.0)
+    if cached and time.monotonic() - cached[0] <= ELECTRICITYMAPS_CACHE_TTL:
+        logger.debug(
+            f"electricitymaps_api: using cached carbon intensity {cached[1]} gCO2e/kWh"
+        )
+        return cached[1]
+
+    remaining = cooldown_until - time.monotonic()
+    if remaining > 0:
+        raise ElectricityMapsAPICooldownError(
+            "Electricity Maps API is in cooldown after a previous failure, "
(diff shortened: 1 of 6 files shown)


davidberenstein1957:
CI fix pushed.

The failing test (`test_cumulative_emissions_with_varying_intensity`) was written to protect a different contract than the one it appeared to break. Its job is to prove that emissions are accumulated as per-tick deltas (delta_energy x intensity at that tick) rather than recomputed as total_energy x latest_intensity; varying the intensity across three measurements was only the mechanism for telling those two behaviours apart. The 5-minute TTL cache is correct behaviour on its own terms - grid intensity is published hourly at best, and refetching on every measurement tick was the request storm this PR set out to remove - so I kept the caching design exactly as it is and did not change any implementation code.

What changed is the test only: it now patches `ELECTRICITYMAPS_CACHE_TTL` to a negative value for the duration of the test, so each measurement expires the entry and sees a fresh mocked intensity, and it resets the module-level cache before and after so the global state cannot leak in from or out to other tests. The cumulation assertions (0.1 / 0.3 / 0.6, and the < 0.8 guard against the non-cumulative 0.9) are untouched, and no network calls are made - the responses are mocked as before.

The caching behaviour itself remains covered by `tests/test_electricitymaps_cache.py`, including TTL expiry, per-location keying and the failure cooldown.

Verified locally: `tests/test_emissions_tracker.py`, `tests/test_electricitymaps_cache.py`, `tests/test_electricitymaps_api.py` and `tests/test_emissions.py` all pass, and pre-commit is clean on the changed file.

benoit-cty:

## Verdict: 🔧 Request changes (small)

Extracting `get_carbon_intensity()`, caching per (location, token), a cooldown after a failure, and debug-level skip logs are all good ideas. Thread safety under the lock is fine: parallel cache misses only cause a harmless duplicate fetch. 66 tests pass, and the PR merges cleanly with master.

**Must fix:**
1. **The documented backoff doesn't exist.**
   - `docs/how-to/configuration.md` and the description promise an "exponentially growing cooldown (30 s up to 1 hour)".
   - The code uses a fixed `ELECTRICITYMAPS_COOLDOWN = 60` (`electricitymaps_api.py:22`, `:40`), and no test covers doubling.
   - Please implement it: start at 30 s, double on each consecutive failure, cap at 3600 s, reset on success, with tests. Or fix the docs. Implementing it is preferable, see point 2.
2. **With default settings, the cache and cooldown almost never trigger.**
   - During a run, intensity is fetched about every `api_call_interval × measure_power_secs` = 8 × 15 = **120 s** (`emissions_tracker.py:1278-1284`). That is longer than both the 60 s TTL and the 60 s cooldown.
   - So the docs' claim that the cache "removes roughly three requests out of four" is wrong. Please correct it.
   - A bad token still logs one error per API call on a normal run. The exponential cooldown from point 1 fixes that.

**Low:**
3. A 30 s request timeout is still hit once per cooldown window, on the scheduler thread. This interacts with #1339's 5 s join cap in `stop()`. Consider a shorter timeout for intensity lookups.

**Nit:**
- Pre-existing: `if geo.latitude:` treats latitude 0.0 as missing. Use `is not None`.

davidberenstein1957:
Made the changes in d9b956f4: cooldown now 30 s doubling to 1 h (reset on success, tested), docs no longer claim 3-in-4 savings, intensity timeout cut to 10 s, and `latitude is not None`. Kept the 60 s cache TTL as is.

PR #1359, opened 2026-08-12
feat: add a cloud_detection switch

### LINKED ISSUE #1352: Add a cloud_detection switch and probe cloud metadata concurrently
**Feature proposal.**

## Problem

Every online `EmissionsTracker` probes the cloud instance metadata service (IMDS) at startup to find out whether it runs on AWS, Azure or GCP. `get_env_cloud_details()` in `codecarbon/core/cloud.py` walks `CLOUD_METADATA_MAPPING` **sequentially**, with a 1 s timeout per provider. All three entries live on the same link-local address `169.254.169.254`, so on a laptop, an on-prem server or any machine where that address is simply not routed, we pay up to three consecutive timeouts before concluding "not on a cloud". That is a guaranteed-failing, guaranteed-slow startup cost for every non-cloud user, and it grows linearly with each provider we might add later.

Some users cannot pay it at all: in air-gapped or tightly egress-filtered environments the probe is not just slow, it is noise in security monitoring, and there is currently no way to turn it off. The only escape today is to use `OfflineEmissionsTracker`, which gives up much more than cloud detection.

## Proposed design

Two small, independent changes to the same code path:

1. **`cloud_detection` switch.** A new tracker option, following the existing configuration conventions, so it works as a constructor argument, a `.codecarbon.config` key and an environment variable:

   ```ini
   [codecarbon]
   cloud_detection = false
   ```
   ```python
   EmissionsTracker(cloud_detection=False)
   ```
   ```
   CODECARBON_CLOUD_DETECTION=false
   ```

   Default stays `true`, so behaviour is unchanged for everyone who does not set it. When disabled, `EmissionsTracker._get_cloud_metadata()` returns an empty `CloudMetadata` without issuing any HTTP request, and the tracker takes the normal non-cloud geolocation path.

2. **Concurrent IMDS probing.** Replace the sequential loop in `get_env_cloud_details()` with a bounded `concurrent.futures.ThreadPoolExecutor` fan-out over the same mapping. Total detection wall time becomes roughly one timeout instead of one per provider, and stays constant if the mapping grows. Provider selection stays deterministic: results are resolved in `CLOUD_METADATA_MAPPING` order, not in completion order, so a machine that somehow answers on two entries always reports the same provider it does today.

Neither change alters the public return shape of `get_env_cloud_details()` or the `CloudMetadata` dataclass, and no new dependency is involved — `concurrent.futures` is stdlib.

`CLOUD_METADATA_MAPPING` is already the single source of truth for what gets probed and how; the concurrency change is confined to the loop that consumes it, and the per-entry `postprocess_function` hook (used by GCP to strip `attributes`, which carries Kubernetes config and secrets) is preserved unchanged. The configuration switch uses `_set_from_conf`, the same mechanism as every other tracker option, so it inherits config-file and environment-variable support for free.

## Scope boundary

This proposal deliberately covers only the two items above. Explicitly **out of scope**, and better argued separately:

- Support for additional providers (OVH, Scaleway, Hetzner, OCI). Adding entries is cheap, but each needs a payload parser and a verification predicate — several vendors answer on the same link-local address, and the generic OpenStack metadata path answers on any OpenStack private cloud, so a naive entry risks mislabelling a machine's region. A confidently wrong region is worse than no region.
- Region-level carbon intensity in `codecarbon/data/cloud/impact.csv`. The schema already supports it; the real cost is data curation and provenance (location-based vs market-based figures, sourcing, staleness), which deserves its own discussion.
- A `carbon_intensity_source` field on `EmissionsData` recording which fallback rung produced the intensity.

Happy to open follow-ups for those once the plumbing here is in place.

davidberenstein1957:
## What this adds

Two small changes to the cloud metadata detection path.

**Concurrent probing.** `get_env_cloud_details()` walked `CLOUD_METADATA_MAPPING` sequentially with a 1 s timeout per provider. All three entries live on `169.254.169.254`, so any machine where that address is not routed paid up to three consecutive timeouts at tracker startup before concluding it is not on a cloud. The loop now fans out over a `ThreadPoolExecutor` (stdlib, no new dependency) and detection costs about one timeout regardless of how many providers are in the mapping. Results are resolved in mapping order rather than completion order, so the provider reported for a machine that answers on more than one entry is the same as before.

**A `cloud_detection` switch.** New tracker option, wired through `_set_from_conf` like every other one, so it works three ways:

```python
EmissionsTracker(cloud_detection=False)
```
```ini
[codecarbon]
cloud_detection = false
```
```shell
export CODECARBON_CLOUD_DETECTION=false
```

Default is `true`, so nothing changes for existing users. When disabled, `EmissionsTracker._get_cloud_metadata()` returns an empty `CloudMetadata` without any HTTP call and the tracker takes the normal geolocation path. This is aimed at air-gapped and egress-filtered environments, where the probe today is a guaranteed-failing request with no way to opt out short of switching to `OfflineEmissionsTracker`.

The public return shape of `get_env_cloud_details()` and the `CloudMetadata` dataclass are unchanged, as is the GCP `postprocess_function` that strips `attributes` (Kubernetes config and secrets).

## Verification

`uv run pytest tests/test_cloud.py tests/test_geography.py tests/test_config.py tests/test_emissions_tracker.py` passes. Four new tests in `tests/test_cloud.py`, none of them touching the network:

- all providers are probed and detection stays deterministic when several answer (guards the concurrency change),
- no request is issued when `cloud_detection=False`,
- detection still runs by default.

The existing per-provider `responses`-based tests are untouched and still pass, which covers the refactor of the probe itself.

`black --check` is clean on the four touched files. `uv run task lint` was not run: it currently reports pre-existing findings across the whole repository unrelated to this change.

## Deliberately left out

The proposal this came from also covered broader vendor support and region-level intensity. Both are better argued separately and neither is needed for the wins above:

- **New providers** (OVH, Scaleway, Hetzner, OCI). Each needs a payload parser and a verification predicate — several vendors share the same link-local address and the generic OpenStack metadata path answers on any OpenStack private cloud, so a naive mapping entry risks mislabelling a machine. A confidently wrong region is worse than no region.
- **Region-level carbon intensity** in `codecarbon/data/cloud/impact.csv`. The schema already supports it; the cost is data curation and provenance, not code.
- **A `carbon_intensity_source` field** on `EmissionsData` recording which fallback rung produced the intensity.
- **`track_emissions` decorator**: the new option is not exposed as a decorator keyword, since the config file and environment variable already cover it.

Closes #1352

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -424,6 +424,7 @@ class BaseEmissionsTracker(ABC):
         allow_multiple_runs: Optional[bool] = _sentinel,
         rapl_include_dram: Optional[bool] = _sentinel,
         rapl_prefer_psys: Optional[bool] = _sentinel,
+        cloud_detection: Optional[bool] = _sentinel,
     ):
         """
         :param project_name: Project name for current experiment run, default name
@@ -519,6 +520,10 @@ class BaseEmissionsTracker(ABC):
                                  (CPU + chipset + PCIe). When False, uses package domains which
                                  are more reliable. Note: psys can report higher values than
                                  CPU TDP and may be unreliable on older systems.
+        :param cloud_detection: Query the cloud provider metadata service at startup to
+                                detect AWS/Azure/GCP, defaults to True. Set to False on
+                                machines that are not on a cloud, or in air-gapped and
+                                egress-filtered environments, to skip the network probe.
         """
 
         # logger.info("base tracker init")
@@ -588,6 +593,7 @@ class BaseEmissionsTracker(ABC):
         self._set_from_conf(force_mode_cpu_load, "force_mode_cpu_load", False, bool)
         self._set_from_conf(rapl_include_dram, "rapl_include_dram", False, bool)
         self._set_from_conf(rapl_prefer_psys, "rapl_prefer_psys", False, bool)
+        self._set_from_conf(cloud_detection, "cloud_detection", True, bool)
         self._set_from_conf(
             experiment_id, "experiment_id", "5b0fa12a-3dd7-45bb-9766-cc326314d9f1"
         )
@@ -1434,7 +1440,11 @@ class EmissionsTracker(BaseEmissionsTracker):
         from codecarbon.external.geography import CloudMetadata
 
         if self._cloud is None:
-            self._cloud = CloudMetadata.from_utils()
+            if self._cloud_detection:
+                self._cloud = CloudMetadata.from_utils()
+            else:
+                logger.debug("Cloud detection is disabled, skipping metadata probe.")
+                self._cloud = CloudMetadata(provider=None, region=None)
         return self._cloud
 
 
--- a/docs/how-to/configuration.md
+++ b/docs/how-to/configuration.md
@@ -178,6 +178,25 @@ Despite its name, this option applies to every counter-based CPU interface:
 It has no effect when CodeCarbon falls back to TDP/CPU-load estimation, since
 that mode models the CPU package only.
 
+## Cloud Detection
+
+At startup, `EmissionsTracker` queries the cloud instance metadata service on the
+link-local address `169.254.169.254` to find out whether it runs on AWS, Azure or
+GCP.
+
+On a machine that is not on a cloud the probe always fails, and in air-gapped or
+egress-filtered environments it is unwanted network noise. Set `cloud_detection`
+to `false` to skip it entirely:
+
+``` ini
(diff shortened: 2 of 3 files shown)


davidberenstein1957:
Closing. With the concurrency half split out, this is a config knob for a problem nobody has reported: cloud metadata probing is one call at tracker init, and if it is slow that is a performance question, while if it returns the wrong thing that is a bug rather than something to switch off. Happy to revisit if someone turns up who actually needs to disable it.

PR #1360, opened 2026-08-12
feat: report measurement quality in `codecarbon detect`

### LINKED ISSUE #1350: Feature proposal: `codecarbon doctor` to report measurement quality
**Feature proposal.**

## The problem

CodeCarbon always produces a number, but it never tells the user how that number was obtained. On a Linux box where `/sys/class/powercap/intel-rapl` is root-only, the CPU silently falls back to a TDP-times-load model. On an unrecognised CPU model the TDP itself is a generic constant. On macOS without `sudo` for `powermetrics` it falls back again. RAM is always modelled, never measured. In every one of those cases the tracker keeps going and writes a confident float to `emissions.csv`.

The information exists — `ResourceTracker` picks a backend and logs it, `CPU._mode` records the mode, `CPU._is_generic_tdp` records whether the TDP was a real match — but it is only visible at INFO/DEBUG log level, mixed into startup noise, and `codecarbon detect` (which is where users look) prints a correct hardware list and says nothing about quality. Users find out their measurement was mostly an estimate later, usually from a reviewer.

This is a disclosure problem, not a measurement problem: the fallbacks are reasonable, they are just invisible.

## Proposed design

A new CLI command:

```
codecarbon doctor [--json] [--strict]
```

It runs the existing hardware detection (`EmissionsTracker(save_to_file=False).get_detected_hardware()` path) and, for each power component, prints a status, the method, why a better method was not used, and the concrete fix:

```
CPU     12th Gen Intel(R) Core(TM) i7-1260P
        ESTIMATED - CPU load model over a 28 W TDP
        Why: /sys/class/powercap/intel-rapl exists but its energy counters are
             not readable by this user (permission denied)
        Fix: sudo chmod -R a+r /sys/class/powercap/intel-rapl
             https://docs.codecarbon.io/how-to/enable-rapl/
```

`--json` gives a machine-readable version for CI gates and bug reports, `--strict` exits non-zero when any component is estimated.

The public shape is one dataclass and one function in `codecarbon/diagnostics.py`:

```python
@dataclass
class ComponentDiagnostic:
    component: str   # "CPU" | "RAM" | "GPU"
    detail: str
    status: str      # "measured" | "estimated" | "unavailable"
    method: str
    reason: str | None
    fix: str | None

def diagnose(hardware) -> list[ComponentDiagnostic]: ...
```

## Why it fits existing extension points

No new measurement code and no new dependency. `ResourceTracker` already centralises the "which backend won" decision, `CPU._mode` and `CPU._is_generic_tdp` already encode the answer, and `cpu.is_rapl_available()` / `powermetrics.is_powermetrics_available()` / `windows_emi.is_emi_available()` already run the checks whose failure reason we want to surface. `diagnose()` reads the hardware objects the tracker builds; it does not duplicate detection. `typer` and `rich` are already required.

## Scope boundary

First PR: `diagnostics.py`, the `doctor` command, docs. Deliberately out of scope, because they change behaviour for every existing user and deserve their own review:

- a start-of-run warning from `EmissionsTracker.__init__` when a component is estimated;
- a `measurement_quality` provenance field on `EmissionsData` / the CSV;
- per-backend `why_unavailable()` methods on each of `rapl.py`, `powermetrics.py`, `windows_emi.py`, `gpu_*.py` — the first version derives the reason from the existing availability checks instead, which covers the common Linux/macOS cases without touching six modules.

Also explicitly not in scope: `doctor` running privileged fix commands itself. It prints them.

davidberenstein1957:
## What this adds

`codecarbon detect` now tells the user, per power component, whether CodeCarbon **measures** it or **estimates** it, why a better method was not used, and the concrete fix.

Today the fallbacks are invisible: on a Linux box where `/sys/class/powercap/intel-rapl` is root-only the CPU silently drops to a load-over-TDP model, an unrecognised CPU model silently uses a generic TDP constant, RAM is always modelled — and `codecarbon detect` prints a correct hardware list that suggests all is well. This is a disclosure problem, not a measurement problem, so the disclosure goes on the command that already claims to describe the machine.

No new dependency, no new measurement code: it reads the mode each hardware object landed in (`CPU._mode`, `CPU._is_generic_tdp`) and reuses the existing availability checks (`powermetrics.is_powermetrics_available`, `windows_emi.is_emi_available`, the GPU driver's own power query) to explain the failure.

## User-facing surface

```
codecarbon detect [--json]
```

- default: the existing hardware list, followed by a per-component report with `MEASURED` / `ESTIMATED` / `UNAVAILABLE`, the method, a `Why:` line and a `Fix:` line, plus a one-line summary;
- `--json`: the hardware list and the report as JSON, for bug reports.

No new command: the report reuses the tracker `detect` already builds. The hardware summary comes from the public `get_detected_hardware()`; the per-component diagnostics read the hardware objects themselves (`tracker._hardware`), which that dict does not expose.

Public API: `codecarbon.diagnostics.ComponentDiagnostic` and `diagnose(hardware) -> list[ComponentDiagnostic]`.

Sample:

```
CPU  12th Gen Intel(R) Core(TM) i7-1260P
    ESTIMATED - CPU load model over a 28 W TDP
    Why: /sys/class/powercap/intel-rapl exists but its energy counter is not
         readable by this user (permission denied)
    Fix: sudo chmod -R a+r /sys/class/powercap/intel-rapl
```

The reason distinguishes "no RAPL interface on this platform" from "RAPL is there but unreadable" — that distinction is the point of the report; without it the user is back on the RAPL how-to page guessing which stanza applies.

## Statuses are probed, not assumed

A status is only `MEASURED` when something can actually be read:

- an Apple Silicon chip is only measurable when `powermetrics` runs without a password, so it probes `is_powermetrics_available()`; a Mac missing the sudoers rule is `UNAVAILABLE` with the fix, not `MEASURED`;
- a driver that lists a GPU does not always report its power (consumer cards and virtualised GPUs often answer "not supported"), so the GPU diagnostic probes `get_gpu_details()` for a power reading and is `UNAVAILABLE` when there is none.

## How it was verified

- `tests/test_diagnostics.py`: measured/estimated CPU classification, the RAPL absent-vs-unreadable distinction against a fake powercap tree in `tmp_path`, generic-TDP flagging, RAM always estimated, Apple Silicon with and without powermetrics, a GPU with and without a power reading, an absent GPU, the JSON schema and the text output of `detect`.
- `uv run pytest tests/test_diagnostics.py tests/cli tests/test_package_integrity.py -q` → 97 passed.
- Ran the real command on an Apple Silicon machine without passwordless `sudo`; both text and `--json` output are correct.
- `black` and `ruff check` run scoped to the touched files (the repo-wide `task format` / `task lint` reformat ~120 unrelated files, so they were not run); the remaining ruff findings in `cli/main.py` are pre-existing.

## Deliberately left out

These change behaviour for every existing user and belong in their own PR with their own release note:

- the start-of-run warning from `EmissionsTracker.__init__` when a component is estimated;
- the `measurement_quality` provenance field on `EmissionsData` / the CSV — the vocabulary is settled here (`measured` / `estimated` / `unavailable`) so a follow-up can adopt it;
- per-backend `why_unavailable()` methods on `rapl.py`, `powermetrics.py`, `windows_emi.py`, `gpu_*.py`; the reason is derived from the existing availability checks instead, which covers the common Linux and macOS cases without touching six modules;
- a cloud/location component (needs an IMDS probe with a timeout budget, worth its own discussion);
- a `--strict` CI gate: a gate whose pass condition is "your kernel lets you read RAPL" is speculative, and `--json` covers the bug-report use case that motivated it;
- running privileged fix commands — they are printed.

Closes #1350

### DIFF
--- a/codecarbon/cli/main.py
+++ b/codecarbon/cli/main.py
@@ -3,11 +3,12 @@ import signal
 import sys
 import time
 from pathlib import Path
-from typing import Annotated
+from typing import Optional
 
 import typer
 from rich import print
 from rich.prompt import Confirm
+from typing_extensions import Annotated
 
 from codecarbon import __app_name__, __version__
 from codecarbon.cli.cli_utils import (
@@ -61,7 +62,7 @@ def _version_callback(value: bool) -> None:
 
 @codecarbon.callback()
 def version(
-    version: bool | None = typer.Option(
+    version: Optional[bool] = typer.Option(
         None,
         "--version",
         "-v",
@@ -487,8 +488,9 @@ def detect(
     previous_level = logger.level
     try:
         set_logger_level("error")
-        # allow_multiple_runs: without it, a live run makes __init__ return early
-        # and leave the tracker half-built - exactly when someone runs `detect`.
+        # allow_multiple_runs is already the default; it is passed explicitly so
+        # a config file setting it to false cannot make __init__ return early and
+        # leave the tracker half-built - exactly when someone runs `detect`.
         tracker = EmissionsTracker(
             save_to_file=False, allow_multiple_runs=True, log_level="error"
         )
--- a/codecarbon/cli/main.py
+++ b/codecarbon/cli/main.py
@@ -490,6 +490,51 @@ def detect():
             f" BUT only tracking these GPU ids : {hardware_info['gpu_ids']}"
         )
     print(f"- GPU model: {gpu_model_str}")
+    print("\nRun `codecarbon doctor` to see what is measured and what is estimated.")
+
+
+@codecarbon.command("doctor", short_help="Report measurement quality per component.")
+def doctor(
+    json_output: Annotated[
+        bool, typer.Option("--json", help="Print the report as JSON.")
+    ] = False,
+    strict: Annotated[
+        bool,
+        typer.Option(
+            "--strict", help="Exit with code 1 if any component is estimated."
+        ),
+    ] = False,
+):
+    """
+    Tell you, for each power component, whether CodeCarbon measures it or
+    estimates it, why, and how to improve it.
+    """
+    import json
+
+    from codecarbon.diagnostics import ESTIMATED, diagnose, render_text, summary
+    from codecarbon.emissions_tracker import EmissionsTracker
+
+    tracker = EmissionsTracker(save_to_file=False)
+    tracker._ensure_hardware_ready()
+    diagnostics = diagnose(tracker._hardware)
+
+    if json_output:
+        typer.echo(
+            json.dumps(
+                {
+                    "codecarbon_version": __version__,
+                    "components": [d.as_dict() for d in diagnostics],
+                    "summary": summary(diagnostics),
+                },
+                indent=2,
+            )
+        )
+    else:
+        print(f"CodeCarbon {__version__} - measurement quality report\n")
+        print(render_text(diagnostics))
+
+    if strict and any(d.status == ESTIMATED for d in diagnostics):
+        raise typer.Exit(1)
 
 
 def questionary_prompt(prompt, list_options, default):
--- a/codecarbon/cli/main.py
+++ b/codecarbon/cli/main.py
@@ -501,7 +501,8 @@ def doctor(
     strict: Annotated[
         bool,
         typer.Option(
-            "--strict", help="Exit with code 1 if any component is estimated."
+            "--strict",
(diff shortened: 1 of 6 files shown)


davidberenstein1957:
Closing. 749 lines to report which measurement fallback was chosen is the wrong shape for this. The useful part is telling users at the moment a fallback fires, which is a handful of logger.warning calls at each fallback site. I will do that as a small PR instead.

PR #1361, opened 2026-08-12
feat: add SCI report output method

### LINKED ISSUE #1353: Feature proposal: SCI (ISO/IEC 21031) report output method
Feature proposal.

## Problem

Sustainability and procurement teams ask for one number by name: the Software Carbon Intensity, as defined by the Green Software Foundation and standardised as ISO/IEC 21031:2024.

```
SCI = (E * I + M) / R
```

CodeCarbon already measures `E` (energy consumed) and already applies `I` (regional carbon intensity, including cloud region and PUE handling). What it does not do is express the result in the shape a reviewer recognises, so teams export the CSV and rebuild the arithmetic in a spreadsheet. `R`, the functional unit, is the one term that genuinely cannot be inferred — only the user knows what one unit of their software is — and it should be recorded as an explicit declaration rather than guessed.

## Proposed design

A new output method, alongside the existing ones, mirroring the BoAmps exporter that already covers exactly this shape of problem (a standards-format JSON report written once at the end of a run).

```python
from codecarbon.output_methods.sci import SCIOutput, FunctionalUnit

sci = SCIOutput(
    functional_unit=FunctionalUnit(name="inference request", count=10_000),
    output_dir="reports",
)
tracker = EmissionsTracker(output_handlers=[sci])
```

Or declaratively, matching how `BoAmpsOutput.from_file()` works today:

```python
sci = SCIOutput.from_file("sci_context.json")
```

and via config, `output_methods = csv,sci`.

The handler writes `sci_report_<run_id>.json` containing the SCI value, the four terms, and a provenance block (grid intensity source, measurement boundary, hardware, PUE, duration). `I` is derived as `emissions * 1000 / energy_consumed` rather than recomputed, so the report is consistent with the CSV by construction. `task_out()` writes per-task reports when tasks are in use.

## The `M` term

CodeCarbon has no manufacturing/embodied data and should not invent any. An uncited embodied figure in a compliance report is worse than a blank one. So `M` is user-declared only: it comes from the context file or the constructor, and when it is not supplied the report still emits, with `M_gCO2e: 0` and `M_source: "not declared"` so the reader can see the report is a partial one. Sourcing embodied figures automatically (a bundled hardware table, cloud-provider per-instance figures) is deliberately out of scope until someone owns the citations.

Similarly, when `R` is never declared the report is written with `sci: null` and a `status` explaining why, rather than dividing by zero or silently assuming `R = 1`.

## Why it fits

`OutputMethod` is a plain str enum in `codecarbon/output_methods/base_output.py` and `_init_output_methods` in `codecarbon/emissions_tracker.py` appends one handler per selected member; `BoAmpsOutput` already demonstrates the "final standards report, `live_out()` is a no-op" pattern. No new dependencies — `json` and `dataclasses` cover it, as they do for BoAmps.

## Scope boundary

In scope: the report, the enum value, config keys, docs, tests.

Explicitly out of scope: any change to what `emissions` means; automatic embodied-carbon estimation; an `EmbodiedProfile` tracker parameter; multiple simultaneous functional units; schema validation on write (no stable machine-readable SCI schema to validate against yet).

davidberenstein1957:
Adds an SCI (Software Carbon Intensity, ISO/IEC 21031:2024) output method, following the same pattern as the existing BoAmps exporter.

## What it adds

`codecarbon/output_methods/sci.py`, a single module holding the two declaration dataclasses, the pure mapping function, and the handler:

- `FunctionalUnit` / `EmbodiedDeclaration` — the `R` and `M` terms, which only the user can supply.
- `map_emissions_to_sci(data, ...)` — pure, no I/O. `I` is derived as `emissions * 1000 / energy_consumed` rather than recomputed, so the report agrees with the CSV by construction and inherits the cloud-region, PUE and country/region handling already applied upstream.
- `SCIOutput(BaseOutput)` — `out()` writes `sci_report_<run_id>.json`, `task_out()` writes `sci_report_tasks_<run_id>.json` with one entry per task, `live_out()` is a documented no-op because the report is final rather than incremental.
- `OutputMethod.SCI` registered in `_init_output_methods`, reading an optional `sci_context_file` from the usual config hierarchy.

Two honesty rules, both tested:

- No `R` declared → the report is still written, with `sci: null` and a `status` field explaining why, instead of dividing by zero or assuming `R = 1`.
- No `M` declared → `M_gCO2e: 0` with `M_source: "not declared"`, so a reader can see the report covers only the operational half.

## User-facing surface

```python
from codecarbon.output_methods.sci import FunctionalUnit, SCIOutput

sci = SCIOutput(functional_unit=FunctionalUnit(name="inference request", count=10_000))
tracker = EmissionsTracker(output_handlers=[sci])
```

`sci.set_functional_unit_count(n)` for the common case where the count is only known at the end; `SCIOutput.from_file("sci_context.json")` for the declarative path; `output_methods = csv,sci` plus `sci_context_file` in `.codecarbon.config`.

Docs: a new SCI section in `docs/reference/output.md` (including a sample report), a row in `docs/how-to/examples.md`, and a runnable `examples/sci_output.py`. No new dependencies.

## Verification

`uv run pytest tests/test_sci_output.py -q` — 14 tests covering the formula, provenance assembly, the undeclared-`R` and undeclared-`M` branches, zero-energy, the handler and its task path, and context-file loading (valid, missing, malformed). `black --check` and `ruff check` were run scoped to the touched files; the remaining ruff hits are the same `Optional`/`List`/`dict()` style the surrounding modules already use, so the new code matches its neighbours.

## Deliberately left out

- Any automatic embodied-carbon estimation: no `EmbodiedProfile` tracker parameter, no bundled hardware table, no cloud per-instance figures. CodeCarbon has no defensible manufacturing data, and an uncited `M` in a compliance report is worse than a blank one. `M` is user-declared or explicitly undeclared.
- Any change to what `emissions` means. Existing consumers are untouched.
- Multiple simultaneous functional units, and schema validation on write (there is no stable machine-readable SCI schema to validate against yet).

Closes #1353

---

## Review follow-up: embodied carbon in task reports

`task_out()` passed the same declared `M` to every task, so a run with 5 tasks reported the device's full embodied carbon 5 times, and each per-task `sci` included 100% of it — a wrong number in a document framed as ISO/IEC 21031 output.

`M` is now apportioned by each task's share of the run duration. Chosen over omitting `M` from task reports because the split is exhaustive: `R` is already the run-level count, so with an apportioned `M` the per-task figures sum back to the run-level report instead of quietly dropping half the accounting. `M_source` records the share applied, e.g. `"vendor LCA; apportioned by duration (25.0% of the run)"`.

Tested by `test_task_out_apportions_embodied_by_duration` (unequal durations: asserts the 10/30 split of a declared 40 gCO2e and that the shares sum back to 40), plus the undeclared-`M` and zero-duration branches.

Also documented in `docs/reference/output.md` that the configuration-only path (`output_methods = csv,sci`) can never produce a non-null `sci` unless `functionalUnit.count` is hardcoded in the context file, since `set_functional_unit_count()` needs a reference to a handler the tracker owns.

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -446,7 +446,7 @@ class BaseEmissionsTracker(ABC):
                                    EmissionsTracker(output_methods=[OutputMethod.CSV, OutputMethod.API])
 
                                Available values: ``CSV``, ``API``, ``LOGGER``,
-                               ``PROMETHEUS``, ``LOGFIRE``, ``BOAMPS``.
+                               ``PROMETHEUS``, ``LOGFIRE``, ``BOAMPS``, ``SCI``.
                                When provided, the individual ``save_to_*`` flags are
                                ignored. Defaults to ``[OutputMethod.CSV]``.
                                Can also be set in config as a comma-separated string:
@@ -620,6 +620,7 @@ class BaseEmissionsTracker(ABC):
         from codecarbon.output_methods.http import CodeCarbonAPIOutput, HTTPOutput
         from codecarbon.output_methods.metrics.logfire import LogfireOutput
         from codecarbon.output_methods.metrics.prometheus import PrometheusOutput
+        from codecarbon.output_methods.sci import SCIOutput
 
         methods = set(self._output_methods) if self._output_methods else set()
 
@@ -667,6 +668,23 @@ class BaseEmissionsTracker(ABC):
 
         if OutputMethod.BOAMPS in methods:
             self._output_handlers.append(BoAmpsOutput(output_dir=self._output_dir))
+
+        if OutputMethod.SCI in methods:
+            context_file = self._external_conf.get("sci_context_file")
+            handler = None
+            if context_file:
+                try:
+                    handler = SCIOutput.from_file(
+                        context_file, output_dir=self._output_dir
+                    )
+                except (OSError, TypeError, ValueError) as e:
+                    logger.error(
+                        f"Could not read sci_context_file {context_file}: {e}. "
+                        "Writing an SCI report without the declarations."
+                    )
+            self._output_handlers.append(
+                handler or SCIOutput(output_dir=self._output_dir)
+            )
 
     def get_detected_hardware(self) -> Dict[str, Any]:
         """
--- a/codecarbon/output_methods/base_output.py
+++ b/codecarbon/output_methods/base_output.py
@@ -15,7 +15,7 @@ class OutputMethod(str, Enum):
         )
 
     Available values: ``CSV``, ``API``, ``LOGGER``, ``PROMETHEUS``,
-    ``LOGFIRE``, ``BOAMPS``.
+    ``LOGFIRE``, ``BOAMPS``, ``SCI``.
 
     .. note::
         HTTP output is not configured here; it is enabled by setting the
@@ -28,6 +28,7 @@ class OutputMethod(str, Enum):
     PROMETHEUS = "prometheus"
     LOGFIRE = "logfire"
     BOAMPS = "boamps"
+    SCI = "sci"
 
 
 class BaseOutput:
--- a/codecarbon/output_methods/sci.py
+++ b/codecarbon/output_methods/sci.py
@@ -1,0 +1,291 @@
+"""
+SCI (Software Carbon Intensity) output handler for CodeCarbon.
+
+Writes a report in the shape defined by the Green Software Foundation's
+Software Carbon Intensity specification, standardized as ISO/IEC 21031:2024::
+
+    SCI = (E * I + M) / R
+
+CodeCarbon measures ``E`` (energy consumed) and applies ``I`` (carbon
+intensity). ``R``, the functional unit, and ``M``, the embodied emissions,
+are declarations only the user can make, so they are never guessed: an
+undeclared term is reported as undeclared.
+"""
+
+import json
+import os
+from dataclasses import dataclass
+from typing import List, Optional
+
+from codecarbon.external.logger import logger
+from codecarbon.output_methods.base_output import BaseOutput
+from codecarbon.output_methods.emissions_data import EmissionsData, TaskEmissionsData
+
+SCI_SPEC_VERSION = "ISO/IEC 21031:2024"
+
+
+@dataclass
+class FunctionalUnit:
+    """The ``R`` term: what one unit of the software is, and how many happened."""
+
+    name: str = ""
+    count: float = 0.0
+
+
+@dataclass
+class EmbodiedDeclaration:
+    """The ``M`` term: user-declared embodied emissions, in gCO2eq."""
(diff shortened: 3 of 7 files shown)


benoit-cty:

## Verdict: ✅ Approve after 1 doc fix and 1 robustness fix

`SCIOutput(BaseOutput)` is clean, self-contained and follows the BoAmps pattern. The formula `(E*I + M)/R` is implemented correctly. It refuses to invent R (writes `sci: null` with a status) and flags an undeclared M. 29 tests pass, and it merges cleanly with master.

**Must fix:**
1. **The sample report in `docs/reference/output.md` shows the wrong SCI value.** It shows `"sci": 0.0000127` with E=0.1007, I=417.08, M=42.5, R=10000. The formula gives (0.1007×417.08 + 42.5)/10000 = **0.00845**. This is the canonical example people will copy, so it needs to be right.
2. **A context file whose top level isn't a JSON object stops all tracker output.**
   - A `sci_context_file` holding valid JSON that isn't an object (e.g. `[1]`) raises `AttributeError` at `sci.py:183`.
   - The `except (OSError, TypeError, ValueError)` around `emissions_tracker.py:~678` doesn't catch it. `__init__` aborts under `@suppress` and leaves `_output_handlers == []`, so **no CSV or anything else is written**.
   - Fix: validate `isinstance(context, dict)` in `from_file` and raise `ValueError`, or add `AttributeError` to the except. Please add a test.

**Should fix:**
3. **M is split by the tasks' summed duration, not the run's** (`sci.py:259-262`).
   - `share = task.duration / sum(task durations)`, but the docs and docstring say "share of the run's duration".
   - Example: a 100 s run with one 10 s task assigns that task 100% of M.
   - Either divide by the run duration, or change the docs and drop the "per-task figures sum back to the run-level report" claim, which only holds when tasks cover the whole run.

**Nits:**
- Task reports always show `"pue": 1`, because `TaskEmissionsData` has no `pue` field and `sci.py:103` falls back to 1. That is misleading provenance when PUE ≠ 1. I itself is still correct because it's derived.
- `set_functional_unit_count()` mutates the caller's `FunctionalUnit` instance. Copy it, or document that.
- A relative `sci_context_file` resolves against the CWD, not the config file's location. Worth a line in the docs.
- `flush()` writes an interim report that applies the full declared M to only part of E. It is overwritten at `stop()`, but anyone reading the flushed file sees an inflated M share. Consider pro-rating M by elapsed time in interim reports.

davidberenstein1957:
Made the changes in 89d2897: non-object context file now raises ValueError (tracker degrades, test added), sample SCI fixed to 0.00845, M-split docs now say tasks' summed duration, task pue is null, FunctionalUnit is copied, relative path documented.
Skipped pro-rating M in flush() reports: they're overwritten at stop(), and pro-rating needs a design call on elapsed vs. declared run length.

PR #1362, opened 2026-08-12
feat: record token counts on tasks for energy-per-token reporting

### LINKED ISSUE #1347: Feature proposal: record token counts on tasks for energy-per-token reporting
This is a **feature proposal** for discussion before implementation.

## The user problem

When people measure a local LLM with CodeCarbon, the number they get back is energy per script run. That number is not comparable to anything: it depends entirely on how many prompts happened to be sent during the run. The number people actually want — and the one that can go in a model card, a benchmark table, or a quantization decision — is energy per output token, or CO2 per request.

CodeCarbon already measures the hard part (the power draw over a task). The only missing ingredient is the token counter, and that is something only the caller knows. Today users have to keep that bookkeeping entirely outside CodeCarbon and do the division themselves, and nothing lands in the task CSV.

## Proposed design

Small and contained, reusing the existing task machinery rather than adding a subsystem.

1. Three optional fields on `TaskEmissionsData` (defaults `0`, so existing rows and readers are unaffected): `input_tokens`, `output_tokens`, `n_requests`. Plus two derived properties, `energy_per_output_token` (kWh) and `emissions_per_request` (kgCO2eq), computed rather than stored so they cannot go stale, and returning `0.0` instead of raising when the counter is zero.

2. Counters carried on `Task` and passed through `Task.out()`, so they flow to every output handler through the existing `BaseOutput.task_out()` path and appear as three new CSV columns.

3. One recording entry point on the tracker:

```python
tracker.record_tokens(input_tokens=..., output_tokens=..., n_requests=1)
tracker.record_tokens(response=resp)  # duck-typed extraction
```

   The `response=` form reads the counts that serving stacks already return — `usage.prompt_tokens` / `usage.completion_tokens` for OpenAI-compatible clients, `prompt_eval_count` / `eval_count` for Ollama, `prompt_token_ids` / `outputs[].token_ids` for vLLM `RequestOutput`. All of it is `getattr` / `dict.get`, so CodeCarbon imports nothing new and works with any client following those shapes.

4. The same method exposed on `TaskEmissionsTracker`, so the context-manager form reads naturally:

```python
with TaskEmissionsTracker("llama3.1:8b", tracker) as task:
    for prompt in prompts:
        resp = client.chat.completions.create(...)
        task.record_tokens(response=resp)
```

Counters accumulate across the life of one task rather than opening a task per request, because a single request is typically far shorter than `measure_power_secs` and per-request tasks would be dominated by measurement noise.

`start_task` / `stop_task`, `Task`, `TaskEmissionsData` and `task_out()` already exist and already do the energy isolation and the routing to output handlers. This change adds fields to a dataclass and one method; it introduces no new module, no new dependency, and no new measurement path.

## Scope boundary

Deliberately out of scope:

- No prefill/decode split. Attributing those separately needs sub-second sampling that the current scheduler cannot deliver.
- No per-request task granularity. Accumulation only, for the noise reason above.
- No vendor-specific or model-specific machinery, and no optional dependency on any serving library — extraction stays duck-typed.
- Token counts stay task-level and do not propagate to run-level `EmissionsData`, since a run may mix inference with other work.
- Under continuous batching (e.g. vLLM), requests overlap and per-request attribution is not physically meaningful; aggregate per-token figures remain valid and the limitation is documented rather than papered over.

Happy to adjust the surface before implementing.

davidberenstein1957:
## What this adds

Energy per run is not a comparable number for an LLM: it depends entirely on how many prompts happened to be sent. Energy per output token is. CodeCarbon already measures the power draw over a task; the only missing ingredient was the token counter, which only the caller knows.

This PR lets the caller hand that counter over, and carries it through the existing task machinery to the output handlers.

## User-facing surface

Three new fields on `TaskEmissionsData`, all defaulting to `0`: `input_tokens`, `output_tokens`, `n_requests`. They are plain stored counters. **No derived properties are exposed** — an earlier version of this description claimed `energy_per_output_token` and `emissions_per_request` were added, and that was wrong. `TaskEmissionsData.values` is built from `__dict__`, so a property reaches no output method, and both figures are one division away from the columns above. See the comment at `codecarbon/output_methods/emissions_data.py:115-117`.

One recording entry point, on the tracker and mirrored on `TaskEmissionsTracker`:

```python
with TaskEmissionsTracker(task_name="llama3.1:8b", tracker=tracker) as task:
    for prompt in prompts:
        response = client.chat.completions.create(...)
        task.record_tokens(
            input_tokens=response.usage.prompt_tokens,
            output_tokens=response.usage.completion_tokens,
        )
```

That is the whole API. CodeCarbon does not inspect the response object: the caller reads its own client's usage fields, which it already knows how to do, and no inference library is imported or added as a dependency. (A `response=` argument that duck-typed OpenAI / Ollama / vLLM payloads was in an earlier revision and has been removed — vendor response shapes drift, and each drift would land here as a silent zero.)

Counts accumulate over the life of one task; recording with no active task logs a warning and is otherwise a no-op.

The three fields flow through `Task.out()` and appear as three extra columns in the per-task CSV. Note that the task CSV is rewritten in full on each run (`FileOutput.task_out` writes a fresh dataframe to a per-run-id path), so there is no header-migration concern for existing files.

## Verification

`tests/test_token_tracking.py`: accumulation across several `record_tokens` calls in one task, recording with no active task, and an end-to-end check that the columns land in the task CSV with the right values. `tests/test_tracking_inference.py` and `test_docs_examples.py` still pass.

```
uv run pytest tests/test_token_tracking.py tests/test_tracking_inference.py tests/test_docs_examples.py -q
15 passed
```

`ruff format` / `ruff check` are clean on the touched files (the repo-wide `task lint` / `task format` were not run — they currently churn ~120 unrelated files).

## Deliberately left out

- No response-object introspection for any serving stack.
- No derived per-token / per-request properties; the raw counters and the existing energy columns are enough to compute them.
- No prefill/decode split — attributing those needs sub-second sampling the current scheduler cannot deliver.
- No per-request task granularity. A single request is typically far shorter than `measure_power_secs`, so per-request tasks would be dominated by measurement noise.
- Token counts stay task-level and do not propagate to run-level `EmissionsData`, since a run may mix inference with other work.
- Continuous batching makes per-request attribution physically meaningless; that is documented as a limitation rather than papered over.

Docs updated: an "LLM inference, energy per token" section in `docs/tutorials/python-api.md` and the three task columns in `docs/reference/output.md`.

Closes #1347

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -819,6 +819,15 @@ class BaseEmissionsTracker(ABC):
             return
         if response is not None:
             extracted_input, extracted_output = extract_token_counts(response)
+            if not extracted_input and not extracted_output:
+                # Most common cause: a streamed OpenAI chunk, whose `usage` is
+                # None unless the request passed
+                # stream_options={"include_usage": True}.
+                logger.debug(
+                    "record_tokens : No token count found on the given response, "
+                    "recording 0 tokens. For a streamed response, ask your client "
+                    'for usage data (OpenAI: stream_options={"include_usage": True}).'
+                )
             input_tokens += extracted_input
             output_tokens += extracted_output
         task.record_tokens(
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -27,7 +27,7 @@ from codecarbon.external.hardware import CPU, GPU, AppleSiliconChip
 from codecarbon.external.logger import logger, set_logger_format, set_logger_level
 from codecarbon.external.ram import RAM
 from codecarbon.external.scheduler import PeriodicScheduler
-from codecarbon.external.task import Task
+from codecarbon.external.task import Task, extract_token_counts
 from codecarbon.input import DataSource
 from codecarbon.lock import Lock
 from codecarbon.output_methods.base_output import BaseOutput, OutputMethod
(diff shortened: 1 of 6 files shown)


davidberenstein1957:
Closing. This asks callers to hand CodeCarbon token counts it cannot verify, in order to compute a division the caller can already do themselves — in exchange for three permanent schema fields. Not a good trade for a measurement library, and no one has asked for it.

Issue #1356, opened 2026-08-12
Carbon-aware scheduling: `codecarbon wait` and a `@carbon_aware` decorator


davidberenstein1957:
**Feature proposal.**

## The problem

CodeCarbon answers "how much did that job cost the atmosphere?" It cannot answer "when should I have run it?"

For jobs that are not latency-sensitive — nightly retrains, hyperparameter sweeps, batch inference, backfills, cron CI — the largest reduction available is usually not making the code faster, it is moving it a few hours. Grid carbon intensity in a typical European or US zone swings by a factor of two to four within a day. Deferring a multi-hour training job from an evening peak to an overnight wind trough is a large cut for a one-line change.

This is the question users ask after their first month with the package: "fine, I know it's 4 kg — now what?"

## Proposed design

Three layers, each usable on its own.

**1. Forecast retrieval.** A new `codecarbon/core/intensity_forecast.py` exposing `IntensityPoint` / `Forecast` dataclasses and `get_forecast(geo, token=..., horizon_hours=48) -> Forecast | None`. It reuses the existing geolocation (`GeoMetadata` lat/lon or `country_2letter_iso_code`) and the already-supported `electricitymaps_api_token` config key, so existing users need no new credential. When no provider is configured it returns `None` and every caller degrades to "run now" — a job must never block on a missing credential.

**2. Window selection.** `best_window(forecast, duration, deadline) -> (start, mean_intensity)`. A sliding mean over the forecast points restricted to windows finishing before the deadline. Pure function, no I/O, trivially testable. It returns "now" whenever nothing better exists, so the run-immediately path is the default rather than a special case.

**3. User-facing surfaces.**

A `codecarbon wait` CLI command, alongside `codecarbon monitor`:

```console
# print the recommendation and exit
$ codecarbon wait --dry-run --deadline 24h --duration 90m
Best start: 2026-08-13 03:00 UTC  (112 gCO2e/kWh, now: 341)  -> saves ~67%

# block until the greenest window, then run under measurement
$ codecarbon wait --deadline 12h --duration 2h -- python train.py
```

The blocking form delegates to the existing `run_and_monitor`, so measurement, CSV output and exit-code propagation are unchanged.

And a decorator / context manager in `codecarbon/carbon_aware.py`:

```python
@carbon_aware(deadline="6h", duration="45m")
def nightly_retrain():
    ...
```

[...]

PR #1363, opened 2026-08-12
feat: carbon-aware `codecarbon wait` (forecast + greenest-window selection)

davidberenstein1957:
A deliberately narrow first slice of carbon-aware scheduling (#1356).

> Stacks on #1358: `wait` reuses that PR's cached, backed-off Electricity Maps client instead of opening a second HTTP path. Review #1358 first; this PR targets `feat/intensity-providers` and its diff here is only the `wait` feature.

## What is in it

- `codecarbon/core/intensity_forecast.py` — `IntensityPoint` / `Forecast` dataclasses, `get_forecast()` against the Electricity Maps `/carbon-intensity/forecast` endpoint (going through #1358's shared `request()`, so a failing location backs off once for the whole process rather than once per caller), and `best_window()`, a pure sliding-mean window search with no I/O.
- `codecarbon/cli/wait.py` + a `wait` command in `codecarbon/cli/main.py`:

```console
# print the recommendation and exit
$ codecarbon wait --dry-run --deadline 24h --duration 90m
🌱 Best start: 2026-08-13 03:00 UTC  (112 gCO2e/kWh, now: 341)  -> saves ~67%

# block until the greenest window, then run under measurement
$ codecarbon wait --deadline 12h --duration 2h -- python train.py

# the job must be *finished* within 8 hours, and takes about 2
$ codecarbon wait --finish-by 8h --duration 2h -- python train.py
```

`--deadline` is the latest acceptable **start**, so `--deadline 12h --duration 2h` may leave the job running 14 hours from now. `--finish-by` is the complement most people actually mean: it bounds the end, and is implemented as `finish_by - duration` feeding the same search. A `--finish-by` shorter than `--duration` is rejected.

**One API call.** `wait` fetches the forecast and nothing else. The "now" figure printed, and the value `--threshold` compares against, is the first forecast point — the current period as the forecast sees it. An earlier revision also hit `/carbon-intensity/latest` for a live reading, which was a second network round trip in service of a percentage.

The blocking form delegates to the existing `run_and_monitor`, so measurement, CSV output and exit-code propagation are unchanged. The tracker only starts after the sleep, so a waiting process holds no lock. Ctrl-C during the wait starts the job immediately rather than aborting.

`get_forecast` never raises: no token, a non-200, a malformed payload or an empty forecast all return `None`, and every caller degrades to running now. A job is never blocked on a missing credential.

## What is deliberately not in it

- **The `@carbon_aware` decorator / `wait_for_green_window` context manager.** The CLI is the smaller surface and validates whether anyone wants the sleeping behaviour before we own a second one.
- **`deferred_seconds` / `avoided_emissions` on `EmissionsData`.** Worth having, but they should ship with a proven blocking path — two always-zero CSV columns are a schema change for no reader.
- **A static fallback diurnal profile** for users without a token. It would broaden reach a lot but risks systematically wrong advice: solar-heavy zones trough at midday, wind-heavy zones at night.
- **Mid-wait re-evaluation** as the forecast updates, and region shifting.

## What is blocked on the intensity-provider work

Electricity Maps is currently the only source that can serve a forecast, and only for users holding a paid token. That caps how much of the audience this feature can reach, and it is why this is a narrow slice rather than the full design in #1356.

`get_forecast` is written to be absorbed: once pluggable intensity providers land it should become an optional `forecast()` method on the provider protocol rather than a second HTTP client, and the HTTP lives in one place (#1358's `request()`) so that move is mechanical.

## Tests

`tests/test_intensity_forecast.py` (14) and `tests/cli/test_wait.py` (22), all passing, no network — HTTP is stubbed with `responses` as in `tests/test_electricitymaps_api.py`, and geolocation, config and `time.sleep` are monkeypatched in the CLI tests. Coverage includes payload parsing, lat/lon vs `countryCode` selection, naive-vs-aware timestamps, horizon truncation, every failure mode returning `None`, the shared cooldown skipping the request, the window search (trough, flat, monotonic, deadline shorter than duration, duration longer than horizon), `--finish-by` bounding the end and its rejection when shorter than `--duration`, the duration parser, the threshold short-circuit taking no second API call, and delegation to `run_and_monitor` with the residual command.

```
uv run pytest tests/cli/test_wait.py tests/test_intensity_forecast.py -q   # 36 passed
```

Docs: a `codecarbon wait` section in `docs/reference/cli.md` and a pointer from `docs/tutorials/cli.md`.

Refs #1356

### DIFF
--- a/codecarbon/cli/main.py
+++ b/codecarbon/cli/main.py
@@ -480,6 +480,12 @@ def wait(
         str,
         typer.Option(help="Maximum delay before the job must start."),
     ] = "12h",
+    finish_by: Annotated[
+        Optional[str],
+        typer.Option(
+            help="Latest acceptable finish time, e.g. '8h'. Overrides --deadline."
+        ),
+    ] = None,
     threshold: Annotated[
         Optional[float],
         typer.Option(help="gCO2e/kWh at or below which we start immediately."),
@@ -509,6 +515,7 @@ def wait(
         ctx,
         duration=duration,
         deadline=deadline,
+        finish_by=finish_by,
         threshold=threshold,
         dry_run=dry_run,
         log_level=log_level,
--- a/codecarbon/cli/main.py
+++ b/codecarbon/cli/main.py
@@ -465,6 +465,57 @@ def monitor(
         raise e
 
 
+@codecarbon.command(
+    "wait",
+    short_help="Wait for a low-carbon window, then run a command.",
+    context_settings={"allow_extra_args": True, "ignore_unknown_options": True},
+)
+def wait(
+    ctx: typer.Context,
+    duration: Annotated[
+        str,
+        typer.Option(help="Expected job length, e.g. '90m', '2h', '1h30m'."),
+    ] = "1h",
+    deadline: Annotated[
+        str,
+        typer.Option(help="Maximum delay before the job must start."),
+    ] = "12h",
+    threshold: Annotated[
+        Optional[float],
+        typer.Option(help="gCO2e/kWh at or below which we start immediately."),
+    ] = None,
+    dry_run: Annotated[
+        bool,
+        typer.Option(help="Print the recommendation and exit without waiting."),
+    ] = False,
+    measure_power_secs: Annotated[
+        int,
+        typer.Option(help="Interval between two measures."),
+    ] = 10,
+    log_level: Annotated[
+        str,
+        typer.Option(help="Log level (critical, error, warning, info, debug)"),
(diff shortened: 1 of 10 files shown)


davidberenstein1957:
Closing. This is ~900 lines and a forecast API dependency to do what amounts to sleeping until the grid is cleaner, and there is no demand behind it beyond my own issue. A documented shell snippet covers the common case. The provider work it was stacked on (#1358) is being reduced to a small standalone caching fix and stays open.

PR #1365, opened 2026-08-12
feat(cli): live local dashboard for monitor

### LINKED ISSUE #1348: Feature proposal: live local dashboard for `codecarbon monitor --ui`
This is a feature proposal.

## The problem

Today the feedback loop for a user who just installed CodeCarbon is: run something, wait for it to finish, open `emissions.csv`. For a long training run that is hours of nothing, and for a first five-minute evaluation or a workshop it is fatal — the tool never gets to show that it works.

There is also a recurring support question: "is CodeCarbon actually reading my GPU?" A live per-component power readout answers that immediately, without asking anyone to switch on debug logging.

Neither existing visualization covers this case:

- `codecarbon/viz/carbonboard.py` (the `carbonboard` console script) is a Dash app that reads a *finished* `emissions.csv` and renders it post-hoc. It needs the optional `carbonboard` extra (`dash`, `dash_bootstrap_components`, `fire`) and is by design a historical comparison view.
- `webapp/` + `carbonserver` do show live data, but require FastAPI, PostgreSQL, authentication, an experiment id and network access. That is the right architecture for a team and the wrong one for "what is my laptop doing right now".

## Proposed design

An output handler that keeps a bounded rolling window of live samples in memory, plus a stdlib HTTP server that serves one self-contained HTML page polling a JSON endpoint.

CLI surface, on the existing `monitor` command:

```
codecarbon monitor --ui                      # http://127.0.0.1:8050
codecarbon monitor --ui --ui-port 9000 --ui-host 127.0.0.1
codecarbon monitor --ui --offline --country-iso-code FRA
```

Public API, usable like any other output handler:

```python
from codecarbon.viz.live import LiveDashboardOutput
tracker = EmissionsTracker(output_handlers=[LiveDashboardOutput(port=8050)])
```

Routes: `GET /` (the page), `GET /data` (JSON: rolling samples, tasks, static hardware metadata), `GET /health`.

`BaseOutput.live_out(total, delta)` already exists and is already called from `EmissionsTracker._measure_power_and_energy` on every `measure_power_secs` tick — it is exactly what `PrometheusOutput` uses. `BaseOutput.exit()` is already called from `stop()`, which gives clean server shutdown for free. So the feature is genuinely just "a handler that remembers the last N samples, and something that shows them"; no change to the tracker, the scheduler or the data model.

## Scope boundary

- **No new dependency, no build step, no database, no network.** `http.server`, `json`, `threading`, `collections`, `importlib.resources`. The page is one HTML file with inline SVG for the chart — no CDN, so it works on an air-gapped machine. If this starts wanting a web framework or a JS toolchain, it has overshot.
- Binds to `127.0.0.1` by default; a non-loopback bind prints a warning, since the server is unauthenticated by design. Remote access is SSH port forwarding's job.
- A busy port must never take down a measurement run: log clearly and keep monitoring without the UI.
- **Not** in scope: reading pre-existing `emissions.csv` history (that is carbonboard), auth, persistence, or any growth toward replacing either existing dashboard. The two tools answer different questions and neither is touched.

Happy to take feedback on the CLI naming and on whether the config-file keys (`ui`, `ui_port`, `ui_host`) are wanted at all — flags may well be enough for an interactive foreground command.

davidberenstein1957:
Closes #1348

## What it adds

`codecarbon monitor --live` reprints a Rich table in the terminal on every measurement: elapsed time, cumulative gCO2eq, energy, and current CPU / GPU / RAM power.

Rich is already a dependency, so this is no new dependency, no HTTP server, no port to bind, no HTML/JS asset. Historical, cross-run analysis stays the job of `carbonboard` and the web dashboard.

## User-facing surface

- `codecarbon monitor --live`, also on the wrapped form (`codecarbon monitor --live -- python train.py`).
- `codecarbon.output_methods.live.LiveTableOutput`, usable like any other output handler: `EmissionsTracker(output_handlers=[LiveTableOutput()])`.
- Docs: new section in `docs/how-to/visualize.md`, option in `docs/reference/cli.md`.

## One change outside the new file

`live_out` was only called every `api_call_interval` measures (5 minutes with the defaults), which is far too coarse for a live view. Handlers can now set `live_out_every_measure = True` (declared on `BaseOutput`, defaults to `False`) to be called on every measure with `delta=None` — the delta is deliberately not computed there, since `_compute_emissions_delta` consumes state used by the periodic call.

Those per-measure calls get a read-only snapshot: `_prepare_emissions_data(update=False)` skips `_update_emissions()`, so the measurement loop does not advance `_last_energy_covered` and does not trigger a carbon-intensity lookup. Existing handlers are unaffected.

## Verified

- `tests/test_custom_handler.py`: `_update_emissions` call count is identical with and without a live handler attached, and the live handler is called once per measure.
- `tests/cli/test_cli_main.py`: `--live` registers a `LiveTableOutput`, no handler without it.
- Full `tests/` suite passes (`test_viz_data.py` skipped locally, `dash` not installed).

## Deliberately left out

- Reading existing `emissions.csv` history — carbonboard's job.
- A browser UI, charts, config-file keys.


davidberenstein1957:
Closing. A live terminal table plus a new public LiveTableOutput class is a lot of permanent surface for a display nicety. There is one real fix buried in here — the live_out call frequency — and I will extract that into a small standalone PR rather than ship the rest.

PR #1366, opened 2026-08-12
feat: add scheduler job metadata to output

### LINKED ISSUE #1355: Attach SLURM/scheduler job metadata to the emissions record
### The problem

HPC users are among CodeCarbon's most motivated adopters — academic and national centres have sustainability reporting obligations and already account every job by user and project — but the unit of work on a cluster is a SLURM job, not a Python function. Today there is no way to say "this row came from job 1234567 on node nid001, charged to account xyz".

The workaround people actually use is smuggling the job id into the project name:

```bash
export CODECARBON_PROJECT_NAME=$SLURM_JOB_ID
```

That works, but it overloads a field that means something else, and it gives you only one of the several identifiers a centre needs. Our own material shows the gap: `docs/how-to/slurm.md` is a long hand-rolled recipe, and `examples/slurm_rocm/` ships four batch scripts that each re-solve the same tagging problem. The information CodeCarbon needs is already sitting in the environment — SLURM exports it into every job step for free — we simply do not read it.

The payoff for reading it is that the emissions CSV becomes directly joinable against `sacct -j <id>`, which is what turns a pile of per-run rows into a cluster report.

### Proposed design

**New optional fields on `EmissionsData`**, all defaulting to `""` so nothing changes for non-HPC users:

```python
scheduler: str = ""        # "slurm"
job_id: str = ""
job_name: str = ""
job_user: str = ""
job_account: str = ""
job_partition: str = ""
node_name: str = ""
```

**Populated automatically** in `EmissionsTracker._prepare_emissions_data()` from the environment. Under SLURM these come from `SLURM_JOB_ID`, `SLURM_JOB_NAME`, `SLURM_JOB_ACCOUNT`, `SLURM_JOB_PARTITION`, `SLURM_JOB_USER` and `SLURMD_NODENAME`. No user action, no new API — an existing batch script starts producing tagged rows the moment it upgrades.

**A generic escape hatch** so other schedulers do not need code: any field can be set or overridden with a `CODECARBON_JOB_*` environment variable (`CODECARBON_JOB_ID`, `CODECARBON_JOB_ACCOUNT`, …). PBS, LSF and OAR sites map their own variables onto that contract in two lines of shell, and we publish the contract as documentation rather than shipping a backend per scheduler.

### Why it fits the existing extension points

`_prepare_emissions_data()` is already the single place where run context is assembled onto the record, alongside cloud and geography metadata, so this is one more source of context in the place that collects context. The CSV writer already detects a header change and backs up the previous file, so the schema addition is handled by machinery that exists. The API output path lists its fields explicitly in `ApiClient.add_emission`, so the new columns land in the CSV without altering the API contract. Prometheus labels are an explicit allowlist and are likewise unaffected. There are no new dependencies — it is `os.environ` and a dataclass.

### Scope boundary

Deliberately out of scope for now:

- **A `codecarbon slurm` wrapper command.** `codecarbon monitor -- <cmd>` already wraps an arbitrary command, and with automatic metadata it needs nothing SLURM-specific. A dedicated command should wait until there is a job the generic one cannot do.
- **Prolog/epilog integration.** Reading energy counters at job start and end from SLURM's prolog/epilog would give a centre cluster-wide coverage with no user changes, but it means CodeCarbon code running as root on every compute node. That is a much larger blast radius and deserves its own proposal and its own security review.
- **One-tracker-per-node rank election.** Only meaningful once something is launching trackers on our behalf. Related to the same problem in distributed training, and should be solved once, generically, when there is a caller for it.
- **Per-job attribution on shared nodes.** When several jobs share a node, node-level power cannot be split between them. This needs a decision before any site-wide path ships; it does not block simply labelling the rows.
- **PBS/LSF backends**, superseded by the `CODECARBON_JOB_*` contract above.

### Testing

Entirely testable with `monkeypatch.setenv` — no cluster required. A fake SLURM environment produces populated metadata, an empty environment stays inert, and `CODECARBON_JOB_*` takes precedence.

Implementation of the scoped slice is ready; PR to follow.

davidberenstein1957:
## What this adds

One new CSV column, `scheduler_job_id`, filled from `SLURM_JOB_ID` when CodeCarbon runs inside a SLURM job step. Under SLURM this happens automatically — there is nothing to enable, no new flag, and no code change in a user's batch script. An existing `.slurm` script starts producing tagged rows as soon as it upgrades.

The point is joinability: with the job ID on the row, `emissions.csv` lines up directly against `sacct -j <id>`, which is what turns a pile of per-run rows into a cluster report. It also retires the `CODECARBON_PROJECT_NAME=$SLURM_JOB_ID` workaround, which worked but overloaded a field that means something else.

## User-facing surface

One column, defaulting to `""` so nothing changes off a cluster:

| Output field | Source |
|---|---|
| `scheduler_job_id` | `CODECARBON_SCHEDULER_JOB_ID`, else `SLURM_JOB_ID` |

`CODECARBON_SCHEDULER_JOB_ID` is the whole story for PBS, LSF and OAR: a site maps its scheduler's variable onto ours in one line of shell (`export CODECARBON_SCHEDULER_JOB_ID=$PBS_JOBID`), and we ship a documented contract rather than a backend per scheduler. It also takes precedence over the SLURM value, for sites whose SLURM configuration is unusual.

No new dependencies — this is `os.environ` and one dataclass field.

### Why one column and not seven

An earlier revision of this PR added `scheduler`, `job_id`, `job_name`, `job_user`, `job_account`, `job_partition` and `node_name`. That is seven columns of HPC-only metadata added to the header of *every* CodeCarbon user's `emissions.csv`, rotating the file on upgrade for everyone, to carry values that `sacct -j <id>` returns for free once the job ID is on the row. The job ID is the join key; the rest is a copy of the accounting database. One column.

## How it fits

`EmissionsTracker._prepare_emissions_data()` is already where run context is assembled onto the record alongside cloud and geography metadata, so this is one more context source in the place that collects context. The value is read **once at tracker init** (`self._scheduler_job_id`) rather than re-reading `os.environ` on every flush, and is passed as an explicit keyword rather than splatted into the constructor — a `**detect_job_metadata()` splat turns any future field-name drift into a runtime `TypeError` at flush time.

### Effect on the API and other output methods

- `CodeCarbonAPIOutput` goes through `ApiClient.add_emission`, which lists its fields explicitly, so the new column never reaches carbonserver and no API contract changes.
- Prometheus labels are an explicit allowlist, unaffected.
- `HTTPOutput` (the generic "guide you" endpoint, not carbonserver) does `dataclasses.asdict(total)` and therefore *does* post the extra key. Checked against the server schema anyway: `carbonserver/carbonserver/api/schemas.py` sets `extra="forbid"` only on `Empty`; `EmissionBase` / `EmissionCreate` use pydantic's default `extra="ignore"`, so an unexpected key would be dropped rather than rejected. Nothing to change on the server side.

## Verified

- `tests/test_schedulers.py`: SLURM environment detected; empty environment stays inert; the `CODECARBON_SCHEDULER_JOB_ID` contract works standalone and takes precedence over SLURM; and an end-to-end check that the field reaches `EmissionsData` and its CSV columns. All `mock.patch.dict` env fixtures — no cluster required.
- Full suite: `uv run pytest tests -q --ignore=tests/test_viz_data.py` — 632 passed, 21 skipped (`test_viz_data.py` needs `dash`, not in the dev environment).
- `ruff format` / `ruff check` clean on the touched files (the repo-wide `task format` churns ~120 unrelated files and was not run).

### Upgrade note: existing `emissions.csv` files are rotated once

`FileOutput.has_valid_headers` compares sorted header lists, so a new column means that on the first run after upgrading, an existing `emissions.csv` is backed up next to itself and a new file is started with the new header. Nothing is lost, but a pipeline reading a fixed path will find only the new rows in it. Documented in `docs/reference/output.md`. `tests/test_data/emissions_valid_headers.csv` gains the column, as it must whenever the CSV schema grows.

### Based on #1370

The schema change surfaces a bug in `FileOutput.out`'s `"update"` mode: it cast each value through the stored column's dtype, and an all-empty text column is read back by pandas as `float64`, so `float("")` raised and the write failed. #1370 fixes that at the root by dropping the coercion entirely, so this branch is stacked on `fix/csv-update-dtype-coercion` and carries no workaround of its own. Rebase onto master once #1370 merges.

### Split out: #1390, the multi-rank double-counting warning

`SLURM_NTASKS_PER_NODE > 1` with `tracking_mode="machine"` means every rank measures the whole node and the job's total is multiplied by the rank count. That warning was in this branch, is unrelated to labelling rows with a job ID, and is arguably more valuable — it now lives in **#1390** against the same base.

## Deliberately left out

- **The other six metadata columns.** See above; `sacct -j` has them.
- **A `codecarbon slurm` wrapper command.** `codecarbon monitor -- <cmd>` already wraps an arbitrary command, and with the job ID detected automatically it needs nothing SLURM-specific.
- **Prolog/epilog integration.** Snapshotting energy counters at job start and end would give a centre cluster-wide coverage with no user changes, but it means CodeCarbon running as root on every compute node. That is its own proposal and its own security review.
- **One-tracker-per-node rank election.** Only meaningful once something launches trackers on our behalf, and it is the same problem as `LOCAL_RANK == 0` in distributed training — worth solving once, generically, when there is a caller. For now the hazard is a warning (#1390) and a note in `docs/how-to/slurm.md`.
- **Per-job attribution on shared nodes.** Node-level power cannot be split between concurrent jobs; that decision blocks any site-wide path, not this labelling work.

Docs updated: `docs/reference/output.md` gains the field, and `docs/how-to/slurm.md` gains a "The job ID in the output" section covering the variable, the `sacct` join, the other-scheduler contract and the multi-node caveat.

Closes #1355


davidberenstein1957:
Closing. This adds a permanent CSV column and a new environment variable for a workflow that already has a one-line workaround (CODECARBON_PROJECT_NAME=$SLURM_JOB_ID), which the PR description itself notes. Output schema additions are forever and this one has no demand behind it beyond my own issue. The genuinely useful part of the SLURM work — the multi-rank double-counting warning — lives in #1390 and stays open.

PR #1369, opened 2026-08-12
build: point lint tasks at pre-commit

### LINKED ISSUE #1368: `task lint` and `task format` run tools the project does not enforce
Running `uv run task lint` or `uv run task format` on a clean checkout of `master` reports roughly 1300 findings and rewrites about 120 files that CI considers perfectly fine. One of the autofixes produces broken code.

## Reproduction

```
git checkout master
uv run pre-commit run --all-files   # all hooks pass
uv run task lint                    # ~1300 findings
uv run task format                  # rewrites ~120 files
```

## Root cause

The two tasks invoke linters the project does not actually use:

```toml
lint   = "black --check --diff . && ruff check . && mypy ."
format = "black . && ruff check --fix --exit-non-zero-on-fix ."
```

- There is **no ruff configuration** anywhere in the repository — no `[tool.ruff]` in `pyproject.toml`, no `ruff.toml`, no `.ruff.toml`. Ruff therefore runs against its own defaults, which have no relationship to the style the codebase is written in.
- Ruff is **not in `.pre-commit-config.yaml`**, and `.github/workflows/pre-commit.yml` is the only lint job in CI. Ruff has never gated a single commit.
- `mypy .` walks the whole monorepo, including `carbonserver/` and `webapp/`, rather than the `-m codecarbon` scope the existing `mypy-check` task uses.

The style that *is* enforced comes from pre-commit: autoflake, isort, black (pinned), and flake8 with `.flake8`.

## Impact

Contributors and coding agents are told by `CLAUDE.md` and `docs/how-to/agent-instructions.md` to run these tasks before submitting. Doing so either buries a real change under a repo-wide reformat, or requires manually reverting all of it. Worse, ruff's `SIM117` autofix currently rewrites a nested `with` in `tests/test_powermetrics.py` into syntactically invalid Python.

## Expected

`task lint` and `task format` should run what CI runs, and pass on a clean `master`.

## Also in scope

Two adjacent dead entries in the same task table:

- `precommit = "c"` — the command is the stray string `c`.
- `docs-check-drift = "python scripts/check-docs-drift.py"` — that script does not exist.

davidberenstein1957:
Fixes the `task lint` / `task format` tasks so they run what CI actually enforces.

## What changed

```diff
-precommit = "c"
-lint = "black --check --diff . && ruff check . && mypy ."
-format = "black . && ruff check --fix --exit-non-zero-on-fix ."
+precommit = "pre-commit run --all-files"
+format = "pre-commit run --all-files"
```

Also drops the now-unreferenced `ruff` dev dependency. The `docs-check-drift` task is left untouched: #1383 makes `scripts/check_docs_drift.py` exist and repoints the task at it, so there is nothing dead to remove and no conflict on that line.

## Why

There is no ruff config anywhere in the repo, ruff is not in `.pre-commit-config.yaml`, and `.github/workflows/pre-commit.yml` is the only lint job in CI. So `ruff check .` was running against its own defaults, unrelated to the style the codebase is written in and never gating anything. `mypy .` likewise walked `carbonserver/` and `webapp/` rather than the `-m codecarbon` scope the existing `mypy-check` task uses.

The practical effect was that anyone following `CLAUDE.md`'s instruction to run these before submitting got ~1300 findings and ~120 rewritten files on an otherwise clean tree — and ruff's `SIM117` autofix rewrote a nested `with` in `tests/test_powermetrics.py` into syntactically invalid Python.

## Verification

```
$ uv run task format
autoflake................................................................Passed
isort....................................................................Passed
black....................................................................Passed
flake8...................................................................Passed
```

Passes clean on `origin/master`, matching `uv run pre-commit run --all-files`.

## Notes for review

- **The `lint` task is gone, not aliased.** pre-commit has no check-only mode: `--all-files` runs the hooks, which fix in place, and there is no `--no-fix` to hold them to reporting. A `lint` task pointing at it would rewrite the working tree — the opposite of what someone typing `lint` expects. So there is one task, `format`, and `docs/how-to/agent-instructions.md` (a.k.a. `CLAUDE.md`) now points at it. `mypy-check` remains as the read-only check.
- **Dropping mypy from the lint path is deliberate.** `mypy .` walked `carbonserver/` and `webapp/` instead of the package, it is not in `.pre-commit-config.yaml` and not in CI, so it gated nothing while producing noise. `mypy-check`, scoped to `-m codecarbon`, is untouched and still available for anyone who wants a type check.
- Removing `ruff` from the dev group is deliberate: leaving an unconfigured, unenforced linter installed is what caused this. If the project wants ruff, it should land as a configured pre-commit hook with the codebase brought into line in its own PR.

Closes #1368


inimaz (approved):
Looks good thanks @davidberenstein1957 !

PR #1374, opened 2026-08-12
fix: send and store real sub-second emission durations

davidberenstein1957:
Extracted from #1203 (`feat/add-fastapi-middleware`), which bundled this with unrelated FastAPI middleware work. It stands alone and can be reviewed and merged independently; #1203 will be rebased to drop the duplicated hunks.

## What

- `duration` becomes `float` instead of `int` in `codecarbon/core/schemas.py` and `carbonserver/carbonserver/api/schemas.py` (`EmissionBase`).
- `ApiClient.add_emission` no longer refuses emissions shorter than one second, and no longer truncates the duration with `int(...)`. It does still skip a **non-positive** duration: the server declares `duration` as `Field(..., gt=0)`, so a zero-length flush would be a 422. Skipping rather than clamping, because clamping would invent a duration that was never measured.
- `ExperimentReport`, `ProjectReport` and `OrganizationReport` widen `duration` from `int` to `float`.

## Why

The DB column and the ORM were **always** `Column(Float)` (`carbonserver/carbonserver/api/infra/database/sql_models.py`) — only the pydantic models claimed `int`. So nothing about storage changes here; the schemas are simply being made honest about what the database already holds. On the client side the `< 1` guard silently dropped any measurement shorter than a second, which is every short-lived task.

The report widening is not cosmetic: those endpoints `SUM()` a Float column into a field declared `int`. Today every stored duration happens to be a whole number, so it validates by luck. The first sub-second duration in the database makes the sum non-integral and the response model raises a validation error — a latent 500 on the experiment/project/organization report endpoints. Widening the reports is therefore part of this fix, not a separate cleanup.

**No DB migration is needed.** The `emissions.duration` column is already `Column(Float)` in `carbonserver/carbonserver/api/infra/database/sql_models.py` — this PR only changes the pydantic layer, the storage type is unchanged.

## Tests

- `carbonserver/tests/api/test_schema_compatibility.py::test_millisecond_duration_survives_client_to_server` — a client payload with `duration=0.0042` validates against the server schema unchanged.
- `tests/test_api_call.py::TestApi::test_add_emission_sends_millisecond_duration_unchanged` — replaces `test_add_emission_skips_short_duration`; asserts the POST body carries `0.0042`.

- `tests/test_api_call.py::TestApi::test_add_emission_skips_zero_duration` — a `duration=0.0` flush is dropped client-side and never reaches the server's `gt=0` validator.

Both fail on `master` (`ValidationError`, and `emissions not sent because of a duration smaller than 1`) and pass with the fix.

`uv run pytest tests/ -q --ignore=tests/test_viz_data.py` → 626 passed, 21 skipped. carbonserver unit tests → 108 passed. `pre-commit run --all-files` clean.

### DIFF
--- a/carbonserver/carbonserver/api/schemas.py
+++ b/carbonserver/carbonserver/api/schemas.py
@@ -56,7 +56,7 @@ class User(UserBase):
 class EmissionBase(BaseModel):
     timestamp: datetime
     run_id: UUID
-    duration: int = Field(
+    duration: float = Field(
         ..., gt=0, description="The duration must be greater than zero"
     )
     emissions_sum: Optional[float] = Field(
@@ -283,7 +283,7 @@ class ExperimentReport(ExperimentBase):
     gpu_energy: float
     ram_energy: float
     energy_consumed: float
-    duration: int
+    duration: float
     emissions_rate: float
     emissions_count: int
     cpu_utilization_percent: Optional[float] = None
@@ -396,7 +396,7 @@ class ProjectReport(ProjectBase):
     gpu_energy: float
     ram_energy: float
     energy_consumed: float
-    duration: int
+    duration: float
     emissions_rate: float
     emissions_count: int
     cpu_utilization_percent: Optional[float] = None
@@ -443,7 +443,7 @@ class OrganizationReport(OrganizationBase):
     gpu_energy: float
     ram_energy: float
     energy_consumed: float
-    duration: int
+    duration: float
     emissions_rate: float
     emissions_count: int
     cpu_utilization_percent: Optional[float] = None
--- a/carbonserver/tests/api/test_schema_compatibility.py
+++ b/carbonserver/tests/api/test_schema_compatibility.py
@@ -118,3 +118,27 @@ def test_client_create_payloads_validate_against_server_schemas(
     client_payload, server_schema
 ):
     server_schema.model_validate(dataclasses.asdict(client_payload))
+
+
+def test_millisecond_duration_survives_client_to_server():
+    """A single FastAPI request lasts milliseconds and must not be rounded away."""
+    payload = client_schemas.EmissionCreate(
+        timestamp="2021-04-04T08:43:00+02:00",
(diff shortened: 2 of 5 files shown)


benoit-cty:

## Verdict: ✅ Approve, but deploy the server before releasing the client

Moving `duration` from int to float in the client dataclass, `EmissionBase` and the three `*Report` schemas is correct.
- **No migration is needed:** `emissions.duration` is already `Column(Float)` (`sql_models.py:14`).
- **Old client → new server:** works; ints are accepted by the float field.
- **Report widening:** needed, since the reports `func.sum()` a Float column into `duration: int`.
- **Webapp:** uses `z.number()`, so it is unaffected.
- **Tests:** 77 client tests and 56 carbonserver tests pass.

**Deploy ordering: new client → old server loses data** (medium)
- `codecarbon/core/api_client.py:201` now sends the unrounded duration (e.g. `15.0023`).
- An old server with `duration: int` rejects fractional floats: pydantic v2 `int_from_float` returns **422**. `HTTPOutput._emit` (`output_methods/http.py:66-69`) only logs that, so nearly every emission is **silently dropped**.
- This hits users if the package reaches PyPI before api.codecarbon.io is redeployed, and anyone on a self-hosted carbonserver older than this PR.
- Please mention this in the PR description, and plan to deploy the API before the package release. Also add a release note for self-hosters. Optionally, the client could retry once with `round(duration)` on a 422 to stay compatible with old servers.

**Nit:**
- The test docstrings mention "A single FastAPI request lasts milliseconds", which looks like leftover context from #1203.

davidberenstein1957:
Made the changes in b71cd53: added a deploy-order and self-hoster release note to the description, reworded the test docstrings.
Skipped the optional retry with round(duration) on 422: deploying the server first covers it, happy to add if you want old-server compat.

davidberenstein1957:
Made the changes in 0ddda1a5: client retries once with a rounded duration on a 422, so old servers keep receiving emissions.

PR #1377, opened 2026-08-12
docs: correct the estimation chain reference

davidberenstein1957:
Corrects the estimation chain in `explanation/methodology.md` and the pages that link into it.

## Factual corrections

**`force_cpu_power` is not a fixed constant.** With `force_mode_cpu_load` unset, control reaches `_setup_fallback_tracking` (`core/resource_tracker.py:272-279`), and because `psutil` is a hard dependency (`pyproject.toml:35`) it selects `cpu_load` mode with the user value as the TDP (`resource_tracker.py:179-190`). So in machine mode `force_cpu_power=65` reports `65 × (0.1 + 0.9 × load³)` — 6.5 W to 65 W — not a flat 65 W. Only without `psutil` is it flat, and then it is half the value. It is also used as given, not multiplied by the package count. Table row 1 and the bullet below it are fixed.

**`how-to/troubleshooting.md` no longer contradicts this page.** It said Apple Silicon reads power through `powermetrics` and pointed at sudoers instructions this PR deletes. On Apple Silicon `_setup_cpu_load_fast` wins at `resource_tracker.py:228-230` before `powermetrics` is tried at `:231-233`, so passwordless sudo does not change the backend. That section now says so, and keeps the sudoers recipe where it still applies: Intel Macs without Intel Power Gadget.

## Other content

- `cpu_load` has two power models, cubic with a 10% floor in machine mode and linear with no floor in process mode, selected by `tracking_mode`.
- `pue` is applied per component inside the measurement loop, so the per-component CSV columns are already inflated by it.
- Backend selection and carbon-intensity resolution are stated as ordered tables with source lines.
- `explanation/references.md` is new: the bibliography, moved out of `methodology.md` and out of `accuracy.md`.
- The four-term formula expansion and an index routing row, extracted from the closed #1382.

## Checks

`uv run task docs` passes: build clean, link check clean.


davidberenstein1957:
One thing from #1378, which I am closing in favour of this: `docs/images/cpu_fallback.png` shows a CPU fallback ladder that no longer matches `codecarbon/core/resource_tracker.py` (`cpu_load` is preferred over `constant` whenever psutil is present, and `force_cpu_power` skips the platform probe entirely).

If the ladder gets illustrated again — here or later — please generate the diagram from Mermaid or Graphviz checked into the repo, so the next correction is a diff instead of a redraw in an image editor.

PR #1073, opened 2026-02-21
docs: port documentation to Markdown with Zensical theme

davidberenstein1957:
## Summary

Port CodeCarbon documentation from reStructuredText (.rst) to Markdown (.md) with Zensical theme.

## Changes

- Convert all RST docs to Markdown format
- Add  configuration
- Use CodeCarbon teal branding (matching logo)
- Fix index.md formatting
- Add proper headers and structure matching original RST

## Branding

- Primary color: teal (CodeCarbon's brand color)
- Accent color: teal
- Dark/light mode support enabled

## Files Changed

- docs/*.md (19 files)
- docs/zensical.yml (new)
- zensical.toml (new)

## Verification

Build docs with: `uv run task docs`

PR #1383, opened 2026-08-12
build: add a docs constant drift check

davidberenstein1957:
Stacks on #1377.

Adds a check that fails when a numeric constant changes in the code while the docs keep the old value.

## Why

The methodology rewrite in #1377 found that `methodology.md` had been documenting a CPU fallback ladder the code does not implement, plus several constants that were stale, wrong, or undocumented. Those docs were correct when written. They drifted silently, and nothing caught it.

This closes the mechanical half of that gap. It cannot catch prose going stale, but it can catch a number going stale, and a number going stale is what turned `85 W` into a value users were reading off a page while the code did something else.

## What it does

`scripts/check_docs_drift.py` guards nine constants:

| Constant | Source |
|---|---|
| `POWER_CONSTANT` | `codecarbon/external/hardware.py:23` |
| `CONSUMPTION_PERCENTAGE_CONSTANT` | `codecarbon/external/hardware.py:26` |
| `DEFAULT_POWER_PER_CORE` | `codecarbon/core/cpu.py:29` |
| `RAM_SLOT_POWER_X86` + the x86 floor | `codecarbon/external/ram.py:14` |
| `world_average` | `data/private_infra/carbon_intensity_per_source.json` |
| the three equivalence divisors | `codecarbon/viz/data.py:54-96` |

Every value is **imported** from the codebase, or loaded from the actual data file (the three `carbonboard` equivalence factors were inline literals, so this PR names them `KG_CO2E_PER_MILE`, `KG_CO2_PER_TV_HOUR` and `KG_CO2_PER_HOUSEHOLD_WEEK` in `viz/data.py` and imports those), then asserted to appear in the page documenting it **inside its surrounding phrase** — `0.409 kg CO₂e per mile`, `475 gCO₂eq/kWh`, `10 W (2 DIMMs × 5 W)` — so a bare `0.409` or `85` sitting elsewhere on the page cannot false-pass the check. Nothing re-parses Python source with regexes — a checker that reads the code textually can drift from the code it is meant to guard.

## Scope, deliberately narrow

Not in scope, and the module docstring says so to stop the next contributor extending it wrongly:

- **Prose accuracy** — not mechanically checkable; attempting it produces noise.
- **Link checking** — `scripts/check_docs_links.py` already does it.
- **The order of any fallback ladder** — verifying narrative structure false-fails, and a check that false-fails gets disabled, which is worse than no check. If ladder ordering needs guarding, the honest tool is a unit test over `resource_tracker.py`.

## Verification

Passes on this branch:

```
$ uv run python scripts/check_docs_drift.py
docs drift check: 9 constants match the docs
```

And it has been seen to fail. Changing the documented value to 999 in a scratch copy:

```
Documentation drift: a constant changed in the code but the docs still show the old value.

  POWER_CONSTANT
    code value : 85  (docs must contain 'POWER_CONSTANT = 85')
    docs       : page says: ... `POWER_CONSTANT = 999` ...
    fix        : edit docs/explanation/methodology.md to match the code -- or fix the code if the docs are the correct value
```

Exit 1, and the message names the constant, both values, and the file to edit, so a failure is fixable without reading the script.

There is deliberately **no unit test for the script**: it is itself the test, and the docs CI job failing is the assertion.

## Note for review

`pyproject.toml` had a `docs-check-drift` task pointing at `scripts/check-docs-drift.py`, a file that never existed. This repoints it at the real script (underscored, matching `check_docs_links.py`) and wires it into the docs CI job.

PR #1369 previously deleted that dead task; it has been updated to leave the line alone, so the two branches now merge cleanly.


davidberenstein1957:
Closing. A drift check guarding nine hardcoded constants is a lot of machinery for a failure mode that has occurred once. A comment naming the source file next to each constant does the same job. Separately, this uncovered a real pre-existing bug worth fixing on its own: pyproject.toml defines docs-check-drift pointing at scripts/check-docs-drift.py, which does not exist on master — a dangling reference from #1073. I will open a small PR that just removes or repairs that task line.

Issue #1371, opened 2026-08-12
Flaky: tracker flush and logging-output tests fail intermittently on master


davidberenstein1957:
Three tests fail intermittently when the whole suite runs, across two files:

- `tests/test_emissions_tracker_flush.py::test_carbon_tracker_offline_flush`
- `tests/test_logging_output.py::test_carbon_tracker_offline_logging_output`

The failures are row counts (`0 == 1`, `2 == 3`), so an expected measurement did not land before the assertion. Each test passes in isolation, and it reproduces on `master`, not just on feature branches.

The most likely cause is how CI runs the suite. `.github/workflows/test-package.yml:36` is:

```
uv run codecarbon monitor -- uv run task test-coverage
```

so a live tracker wraps the entire test run. It holds `/tmp/.codecarbon.lock` and writes its own `emissions.csv` in the repo root for as long as the tests execute, while the tests under it are starting and stopping their own trackers. That is a real interaction and it is not present when you run a single test locally.

Two smaller things that make it worse:

- `tests/test_logging_output.py` builds three trackers with no `output_dir` and no `output_file` (lines 48-52, 64-69, and the `@track_emissions` decorator at 81-85), so they all default to `emissions.csv` in the current working directory. Nothing cleans that up.
- Both files use a fixed filename in `tempfile.gettempdir()` (`emissions-test-TestCarbonTrackerFlush.csv`, `emissions-test-TestCarbonLoggingOutput.log`) with no per-run suffix, and `setUp`/`tearDown` both delete it.

`PeriodicScheduler` chaining a `threading.Timer` per tick may also contribute, since a tracker a test does not stop keeps firing into the next test and `tests/conftest.py` only clears the hardware probe caches. Unverified. The re-arming timer goes away with #1339 / #1324, so it is worth re-measuring after one of those lands.

Likely fixes: give the logging test an explicit `output_dir`, make the temp filenames unique per process, and stop asserting on wall-clock-dependent row counts.

davidberenstein1957:
Additional finding from a separate investigation, which narrows this considerably.

`tests/test_emissions_tracker_flush.py` and `tests/test_logging_output.py` are order/state-sensitive **around a shared `emissions.csv` in the repository root**. They fail on `origin/master` under some orderings and **pass after `rm emissions.csv`**.

[...]

PR #1386, opened 2026-08-15
test: make the flush and logging output tests self-contained

NoiceHax:
## Description

`tests/test_logging_output.py` only checks the logger output, but its trackers kept the default CSV output on, so every run also appended a row to `emissions.csv` in the working directory. `_persist_data` calls the output handlers in a loop and the CSV one runs first, so once anything went wrong with that shared file the logger handler was never reached. `flush()` and `stop()` are wrapped in `@suppress(Exception)`, so nothing was printed and the test failed later on a row count instead. The three trackers there now pass `save_to_file=False`, and `tearDown` checks that `emissions.csv` in the working directory was not touched.

`tests/test_emissions_tracker_flush.py` pointed every test at one fixed file name under `tempfile.gettempdir()`, so two tests could end up counting rows in the same file. Each test now gets its own temporary directory.

`tests/conftest.py` gets an autouse fixture that stops `PeriodicScheduler` timers a test left armed. The issue asks for that to stay a safety net, so it stays one.

One thing I left out on purpose: `_persist_data` could wrap each `handler.out()` call in try/except so a failing output method cannot drop the ones after it. That would have turned this silent failure into a warning, but it changes runtime behaviour, so I would rather do it separately if you want it.

## Related Issue

#1371

## Motivation and Context

These two test modules shared files with each other and with the rest of the suite, which is why they pass alone and fail in a full run. This removes the shared state. It does not touch library code.

## How Has This Been Tested?

Ran the full suite five times on master and five times on this branch, on Windows 11 with Python 3.14. Both sets gave the same numbers: the three tests named in the issue passed every time, and the only failure was `test_task_energy_with_live_update_interference`, which already fails on master on this machine. So I could not reproduce the intermittent failure here and cannot claim a drop in the failure rate.

What I could reproduce is the coupling behind it. Replace `emissions.csv` in the repository root with something the CSV handler cannot write to, and all three tests in `test_logging_output.py` fail on master, even though none of them assert anything about CSV output. On this branch they pass.

## Types of changes

## AI Usage Disclosure

- [ ] ðŸŸ¥ AI-vibecoded: You cannot explain the logic. Car analogy : the car drive by itself, you are outside it and just tell it where to go.
- [x] ðŸŸ  AI-generated: Car analogy : the car drive by itself, you are inside and give instructions.
- [ ] â­ AI-assisted. Car analogy : you drive the car, AI help you find your way.
- [ ] â™»ï¸ No AI used. Car analogy : you drive the car.

An AI agent wrote this patch and this description under my direction, and I checked the reasoning and ran the tests locally before opening the PR.

## Checklist:
- [ ] All new and existing tests passed. See the testing section above: one unrelated test fails on this machine on master as well.


davidberenstein1957 (approved):
Verified this against current master rather than taking it on trust, and it holds up.

Reproduced the coupling first: the logging tests kept the default CSV output on, so all three trackers also wrote emissions.csv into the repo working directory. `_persist_data` iterates the handlers with the CSV one first and no per-handler try/except, and both flush() and stop() are wrapped in @suppress(Exception) — so anything that makes the CSV write raise silently aborts the whole persist call, the LoggerOutput handler after it never runs, and the test fails later on a row count with no traceback. Making the shared CSV unwritable and running the file reproduces all three failures on master, and this branch is unaffected.

On the changes: save_to_file=False plus the tearDown assertion that the working-dir CSV was not touched is the actual root-cause removal, and the assertion keeps it from regressing. The per-test TemporaryDirectory also removes the fixed shared filename under gettempdir(), which mattered for parallel checkouts on one machine. The _stop_leaked_schedulers fixture is correct — PeriodicScheduler arms threading.Timer(interval, self._run), so thread.function.__self__ is the scheduler and .stop() reaches it.

Merged onto current master locally: full suite 628 passed / 20 skipped, and the two files pass 3/3 on repeat runs. Confirmed they no longer write to the working directory. No sleeps, no retries, no library changes.

Worth noting for a follow-up, and not a blocker: the library-side silent failure is untouched. A user with an unwritable CSV path still loses their API and logger output too, with nothing logged. A per-handler try/except plus a warning in _persist_data would turn that class of bug into a visible message.

Thanks for the careful fix.

PR #1389, opened 2026-08-15
Log the source of every configuration key

### LINKED ISSUE #312: Add log of config and where they come from
* CodeCarbon version: 2.1
* Operating System: All

### Description

I'd like to know the origin of my codecarbon behavior. For example where did he get my experiment_id.

### What I Did

Use code carbon in any way.

### What I Expect

The source of the configuration is shown in the log.

### What I Get

Nothing

NoiceHax:
## Description

`get_hierarchical_config` only logged which `.codecarbon.config` files it had read, so there was still no way to tell which layer set a given key. It now reads the global file, the local file and the `CODECARBON_` environment variables into separate dicts, merges them in the same order as before, and keeps a key to source map next to the result. The two file level messages are replaced by one line naming the source of each resolved key, with environment variables shown as `CODECARBON_<KEY>`.

Only key names and their source are logged, never the values, because `api_key` and `experiment_id` live in this config space.

Example output:

```
[codecarbon INFO @ 09:30:00] Codecarbon configuration: api_key from /home/victor/.codecarbon.config, experiment_id from CODECARBON_EXPERIMENT_ID, measure_power_secs from /home/victor/.codecarbon.config
```

## Related Issue

Closes #312

## Motivation and Context

Config comes from three places that silently override each other. When a value was not what you expected, nothing in the log pointed at the file or the variable responsible. That is what the issue asks for, for example knowing where `experiment_id` was picked up from.

## How Has This Been Tested?

Two tests added to `tests/test_config.py`. The first checks the key to source map for a key set only in the global file, a key overridden by the local file, and a key overridden by an environment variable. The second checks the log line names the sources and does not contain the value of `api_key`.

Ran `CODECARBON_ALLOW_MULTIPLE_RUNS=True pytest tests/test_config.py tests/test_resource_tracker.py tests/cli`, 114 passed. The full package suite gives 627 passed with one failure in `test_task_energy_with_live_update_interference`, which fails the same way on master on my machine, so it does not come from this change.

## Types of changes

## AI Usage Disclosure

- [ ] ðŸŸ¥ AI-vibecoded: You cannot explain the logic. Car analogy : the car drive by itself, you are outside it and just tell it where to go.
- [x] ðŸŸ  AI-generated: Car analogy : the car drive by itself, you are inside and give instructions.
- [ ] â­ AI-assisted. Car analogy : you drive the car, AI help you find your way.
- [ ] â™»ï¸ No AI used. Car analogy : you drive the car.

I used an AI coding agent to write this patch and this description from an approach I decided on, and I reviewed the code and ran the tests locally before opening the PR.

## Checklist:

- [x] I have updated the documentation accordingly.

### DIFF
--- a/codecarbon/core/config.py
+++ b/codecarbon/core/config.py
@@ -1,9 +1,12 @@
 import configparser
 import os
 from pathlib import Path
-from typing import List, Optional, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 from codecarbon.external.logger import logger
+
+CONFIG_FILE_NAME = ".codecarbon.config"
+CONFIG_SECTION = "codecarbon"
 
 
 def clean_env_key(k: str) -> str:
@@ -36,7 +39,7 @@ def parse_env_config() -> dict:
         dict: existing relevant environment variables mapped into a dict
     """
     return {
-        "codecarbon": {
+        CONFIG_SECTION: {
             clean_env_key(k): v
             for k, v in os.environ.items()
             if k.lower().startswith("codecarbon_")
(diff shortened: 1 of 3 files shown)


benoit-cty (approved):

benoit-cty:
Thanks, it's a good feature :+1:

PR #1390, opened 2026-08-16
feat: warn when SLURM ranks double count a node

davidberenstein1957:
Split out of #1366, which mixed it in with the job-metadata columns. It is unrelated scope and stands on its own.

## The hazard

On SLURM, `srun --ntasks-per-node=4` starts four ranks on one node. With `tracking_mode="machine"` — the default — each of those four trackers measures the *whole node*, so the job's total is silently 4x the truth. Nothing in the output says so; the numbers just look wrong in a way nobody notices.

## What this does

`codecarbon/core/slurm.py` reads `SLURM_NTASKS_PER_NODE` at tracker init and, in `machine` mode with more than one rank per node, logs a warning naming the two ways out: elect one tracker per node (`SLURM_LOCALID == 0`) or switch to `tracking_mode="process"`.

That is the whole change: ~20 lines, `os.environ` and a regex, no dependency, no config key, no new output field. `4(x2)` (heterogeneous allocations) parses as 4.

## Not in this PR

Automatic rank election. Suppressing the extra trackers is the real fix, but it is the same problem as `LOCAL_RANK == 0` in distributed training and deserves one generic solution with a caller asking for it. Warning first.

## Tests

`tests/test_slurm.py`: warns for `4` and for `4(x2)`; silent for `1`, for no SLURM variable at all, and for `tracking_mode="process"`.

```
uv run pytest tests/test_slurm.py tests/test_emissions_tracker.py -q
33 passed
```

Based on #1370, same base as #1366.

### DIFF
--- a/codecarbon/core/slurm.py
+++ b/codecarbon/core/slurm.py
@@ -1,0 +1,26 @@
+import os
+import re
+
+from codecarbon.external.logger import logger
+
+
+def warn_on_multi_rank_double_counting(tracking_mode: str) -> None:
+    """
+    Warn when several ranks share a node and each one measures the whole node.
+
+    In ``machine`` mode every rank reports the node's power, so the job's
+    reported footprint is silently multiplied by the number of ranks per node.
+    """
+    if tracking_mode != "machine":
+        return
+    # SLURM writes this as "4", or as "4(x2)" for a heterogeneous allocation.
+    ntasks_per_node = os.environ.get("SLURM_NTASKS_PER_NODE", "")
+    match = re.match(r"\d+", ntasks_per_node)
+    if match and int(match.group()) > 1:
+        logger.warning(
+            f"SLURM_NTASKS_PER_NODE is {ntasks_per_node} and tracking_mode is "
+            "'machine': every rank measures the whole node, so the job's total "
+            "will be multiplied by the number of ranks per node. Start the "
+            "tracker on one rank per node (SLURM_LOCALID == 0), or use "
+            "tracking_mode='process'."
+        )
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -21,6 +21,7 @@ import psutil
 
 from codecarbon._version import __version__
 from codecarbon.core.config import get_hierarchical_config, normalize_gpu_ids
+from codecarbon.core.slurm import warn_on_multi_rank_double_counting
 from codecarbon.core.units import Energy, Power, Time, Water
 from codecarbon.core.util import count_cpus, count_physical_cpus, suppress
 from codecarbon.external.hardware import CPU, GPU, AppleSiliconChip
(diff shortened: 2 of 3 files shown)


benoit-cty:

## Verdict: 🔧 Request changes

The hazard is real. `allow_multiple_runs` defaults to True (`emissions_tracker.py:163`) and the lock file is in per-node `/tmp`, so nothing stops N trackers per node from each counting the whole machine. The detection needs work, though.

**Must fix:**
1. **Wrong variable, so the common layouts are missed** (`codecarbon/core/slurm.py:16`).
   - `SLURM_NTASKS_PER_NODE` is only set when `--ntasks-per-node` is passed explicitly. `sbatch -N2 -n8` or `--ntasks=8` alone leaves it unset, so the warning never fires.
   - The `4(x2)` format that the code and `tests/test_slurm.py:17` parse belongs to `SLURM_TASKS_PER_NODE` / `SLURM_STEP_TASKS_PER_NODE`, which are always set.
   - Please read `SLURM_STEP_TASKS_PER_NODE`, falling back to `SLURM_TASKS_PER_NODE`.
2. **The warning fires even when the user follows its advice** (`slurm.py:14-19`).
   - The check is env-only, and every task inherits the env. So a user who starts the tracker only on `SLURM_LOCALID == 0`, as the message recommends, still gets the warning.
   - The only way to silence it is `tracking_mode="process"`, which measures something different.
   - Suggested fix: warn only when `SLURM_LOCALID` is set and non-zero. A machine-mode tracker on a non-zero local rank is almost certainly a duplicate. The recommended pattern then never warns, and a naive "tracker on every rank" setup still warns.
   - Also reword the message to "if more than one rank per node starts a tracker…", and consider an opt-out (config key or env var).
3. **False positive in the batch step.** `SLURM_NTASKS_PER_NODE` is also exported into the sbatch batch script. A single tracker in the batch step (e.g. wrapping `torchrun` or an inner `srun`) warns although nothing is double counted. The step-scoped variables from point 1 avoid this.

**Low:**
4. **Log spam.** The warning is emitted from `__init__` (`emissions_tracker.py:604`) once per tracker per rank, so an N-rank job logs it N times. The `SLURM_LOCALID != 0` approach from point 2 reduces that to at most the duplicate ranks.

**Tests:**
- The tests only cover the regex and the mode gate. Please add cases for `SLURM_TASKS_PER_NODE` / `SLURM_STEP_TASKS_PER_NODE`, `SLURM_LOCALID` 0 vs non-zero, and one test that goes through `EmissionsTracker.__init__`.
- The description says "Based on #1370", but the diff contains only the SLURM commit. Please remove that line.

davidberenstein1957:
Made the changes in bceb640: warn only on non-zero SLURM_LOCALID, read step tasks-per-node, reworded message, added tracker-init test. Dropped the #1370 line from the description.
Skipped the opt-out key: with the LOCALID check, following the advice already silences it.

PR #1393, opened 2026-08-17
chore: pr size labeler

inimaz:
## Description
Given the increasing number of PRs, this is a CI helper that labels PRs by size. Like this the maintainer can prioritize and review smaller PRs and merge PRs faster.

## AI Usage Disclosure

- [x] 🟠 AI-generated: Car analogy : the car drive by itself, you are inside and give instructions.

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -1,0 +1,36 @@
+name: PR Size Labeler
+
+# Automatically labels pull requests based on the number of lines changed
+# (additions + deletions), making it easier to prioritize reviews when
+# there are many open PRs.
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+
+permissions:
+  pull-requests: write
+  contents: read
+  issues: write
+
+jobs:
+  size-label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Label PR by size
+        uses: codelytv/pr-size-labeler@095a41fca88b8764fd9e008ad269bcdb82bb38b9 # v1.10.4
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          xs_label: "size/XS"
+          xs_max_size: "10"
+          s_label: "size/S"
+          s_max_size: "50"
+          m_label: "size/M"
+          m_max_size: "200"
+          l_label: "size/L"
+          l_max_size: "500"
+          xl_label: "size/XL"
+          fail_if_xl: "false"
+          message_if_xl: >
+            This PR exceeds the recommended size of 500 lines changed. Consider
+            splitting it into smaller PRs to ease the review process.


.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,36 @@
...
+# there are many open PRs.
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+
+permissions:
+  pull-requests: write
+  contents: read
+  issues: write
```
SaboniAmine, line 14: do we need to enable issue write operation ? I understood that it enables comments writing

.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  size-label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Label PR by size
+        uses: codelytv/pr-size-labeler@095a41fca88b8764fd9e008ad269bcdb82bb38b9 # v1.10.4
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
SaboniAmine, line 23: Would it be possible to use a github app instead of passing a github token ?
Not sure it applies here, but Sebastian Ramirez gave a great feedback about his use of it there : https://bsky.app/profile/tiangolo.com/post/3msy3sbkgxc23

SaboniAmine (commented):
That seems really useful, thanks Inigo! A few comments but we can merge this quickly

.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,36 @@
...
+          xs_max_size: "10"
+          s_label: "size/S"
+          s_max_size: "50"
+          m_label: "size/M"
+          m_max_size: "200"
+          l_label: "size/L"
+          l_max_size: "500"
+          xl_label: "size/XL"
+          fail_if_xl: "false"
+          message_if_xl: >
```
inimaz, line 34: It will post this comment if flagged as XL. Also see that the sizes are customizable, we can start like this and see if they are too small later on?

.github/workflows/pr-labeler.yml
inimaz: This is optional, it is only needed to post a comment to the PR when the PR is flagged as XL. See the other comment.

.github/workflows/pr-labeler.yml
inimaz: Do you mind if we create an issue for this? I completely agree, but we are using `secrets.GITHUB_TOKEN` in other actions also. We can migrate all of them in a different PR?

SaboniAmine: Ok for me!

.github/workflows/pr-labeler.yml
SaboniAmine: Im ok with PR comments, but are issues ones also needed ? Anyway this is not blocking, its not a big deal

SaboniAmine (approved):

.github/workflows/pr-labeler.yml
inimaz: Ah true, I was not getting it. Apparently the comment itself is done through `POST /repos/{owner}/{repo}/issues/{id}/comments`, which requires `issues: write` :( It looks like `pull_request: write` by its own is not enough.

PR #1396, opened 2026-08-19
perf: probe cloud metadata providers concurrently

davidberenstein1957:
Split out of #1359, which bundled this with a `cloud_detection` switch. This is only the probing change, so it can be judged on its own.

`get_env_cloud_details()` walks AWS, Azure and GCP metadata endpoints in order, each with a 1 second `requests` timeout. On a machine that is not on any of the three, all three time out one after the other, so tracker startup pays roughly 3 seconds of link-local requests that were never going to answer. It runs once, in `CloudMetadata.from_utils()` during tracker init.

This submits all three probes to a `ThreadPoolExecutor` and then resolves the futures in `CLOUD_METADATA_MAPPING` order, so if more than one endpoint answers, the provider picked is the same one master would have picked. The per-provider request body moves to a `_probe_provider` helper that returns `None` instead of raising.

The tradeoff a reviewer should weigh, because this is not a pure win in both directions:

- Off cloud: three concurrent 1 second timeouts instead of three sequential ones, so ~1 second instead of ~3. That is the case this targets, and it is the common one.
- On a cloud: master returned as soon as the first matching provider answered, typically in milliseconds, and never probed the rest. Here the `with ThreadPoolExecutor(...)` block calls `shutdown(wait=True)` on exit, and the early `return` is inside that block, so returning still waits for the other two probes to hit their 1 second timeout. Cloud users therefore go from near-instant detection to about 1 second.

Both are once per tracker start, not per measurement, so the on-cloud regression is a one-off second at startup rather than anything recurring. If that is still unacceptable, the alternative is to keep the executor alive past the return (an explicit `shutdown(wait=False)`) and leak the pending threads, which trades the wait for daemon threads outliving the call. I went with the blocking form because it leaves no threads behind; say the word if you prefer the other side of that trade.

`tests/test_cloud.py` gains a case that stubs all three endpoints as answering and asserts both that the first mapping entry wins and that all three were actually called. `uv run pytest tests/test_cloud.py` passes (6 tests).

### DIFF
--- a/codecarbon/core/cloud.py
+++ b/codecarbon/core/cloud.py
@@ -1,3 +1,4 @@
+from concurrent.futures import ThreadPoolExecutor
 from typing import Any, Dict, Optional
 
 import requests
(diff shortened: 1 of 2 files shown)


davidberenstein1957:
Closing this one myself. Splitting it out of #1359 was the right call, but on its own the tradeoff does not resolve: off cloud it collapses three one-second timeouts into one, while on a cloud it waits about a second for the other two providers to time out where the old code returned on the first that answered. That is inherent to leaving the ThreadPoolExecutor context, not something a tweak fixes. A PR whose own description has to argue both directions is not one I would want a maintainer to spend review time on. If provider probing ever shows up as a real startup-cost complaint, this is easy to bring back with a measurement attached.

PR #1397, opened 2026-08-19
fix(powermetrics): time out the powermetrics subprocess

davidberenstein1957:
Split out of #1333, which also rejects powermetrics on non-Apple-Silicon Macs. This is just the subprocess timeout, separated so it can land independently.

`ApplePowermetrics._log_values()` shells out with `subprocess.call(cmd, universal_newlines=True)` and no timeout. The command is `sudo powermetrics ...`, so if sudo has no cached credential and no tty to prompt on, or if powermetrics itself wedges, the call blocks forever. It is invoked from `get_details()` on every measurement, which means the tracker's measurement path hangs with no output and no way out short of killing the process.

This passes a timeout and, on `subprocess.TimeoutExpired`, logs a warning and returns `None` so that measure is skipped rather than blocking. The existing non-zero-returncode path is unchanged.

The timeout value is a heuristic, not a measured bound:

    timeout = self._n_points * self._interval / 1000 * 2 + 5

`_n_points` samples at `_interval` milliseconds each is the nominal runtime of the command, so `n_points * interval / 1000` is that runtime in seconds. It is doubled to absorb sampler overhead and scheduling jitter, and 5 seconds is added as a floor so that short configurations still get a usable margin. With the defaults (`n_points=10`, `interval=100`) that is 1 second of expected work and a 7 second limit.

Nothing measured picks the 2x and the +5; they are chosen to be comfortably loose so a healthy run never trips them, and the timeout only exists to bound a hang. If a reviewer has real numbers for how long powermetrics overruns its nominal duration under load, that is the input that should replace this formula.

`tests/test_powermetrics.py` gains a case asserting `_log_values()` returns `None` and warns once when `subprocess.call` raises `TimeoutExpired`. `uv run pytest tests/test_powermetrics.py` passes (17 tests).

### DIFF
--- a/codecarbon/core/powermetrics.py
+++ b/codecarbon/core/powermetrics.py
@@ -30,6 +30,26 @@ def clear_powermetrics_cache() -> None:
     is_powermetrics_available.cache_clear()
 
 
+def _kill_process(process: subprocess.Popen) -> None:
+    """
+    Kill a Powermetrics subprocess, tolerating a failure to do so.
+
+    The command runs through `sudo`, so the child may be owned by root and
+    refuse our signal. Letting that raise would replace the timeout we are
+    handling with a PermissionError, and waiting on it without a bound would
+    hang for as long as the child lives.
+    """
+    try:
+        process.kill()
+        process.wait(timeout=1)
+    except (OSError, subprocess.SubprocessError):
+        logger.debug("Could not kill the Powermetrics process.")
+    finally:
+        for stream in (process.stdout, process.stderr):
+            if stream is not None:
+                stream.close()
+
+
 def _has_powermetrics_sudo() -> bool:
     if shutil.which("sudo") is None:
         logger.debug("sudo not available, we won't use Apple PowerMetrics.")
@@ -40,7 +60,9 @@ def _has_powermetrics_sudo() -> bool:
         )
         return False
 
-    with subprocess.Popen(
+    # No context manager here: Popen.__exit__ would wait() without a timeout,
+    # which is exactly the hang we are trying to avoid when the kill fails.
+    process = subprocess.Popen(
         [
             "sudo",
             "powermetrics",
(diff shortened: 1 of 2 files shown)


benoit-cty:
Some change made:

1. The timeout now actually bounds the hang. Replaced subprocess.call(..., timeout=) with an explicit Popen + wait(timeout=), deliberately not under a context manager, since Popen.__exit__ calls wait() with no timeout and would re-hang when the kill is refused. Added stdin=subprocess.DEVNULL so sudo gets EOF and exits non-zero instead of blocking on a password prompt — that removes the main hang at the source rather than relying on being able to kill a root-owned child.

2. New [_kill_process()](vscode-webview://13qe2srtq7rfa1e0j47ttranh8f06vadkqqo1rgicmqsrna8gimk/codecarbon/core/powermetrics.py#L33) helper that swallows OSError/SubprocessError from kill()/wait(), so a refused signal can't turn the handled TimeoutExpired into an escaping PermissionError, and closes the pipes the context manager used to close.

3. Applied the same fix to [_has_powermetrics_sudo()](vscode-webview://13qe2srtq7rfa1e0j47ttranh8f06vadkqqo1rgicmqsrna8gimk/codecarbon/core/powermetrics.py#L63). It had the identical bug — a return False from inside with Popen(...) after a possibly-failing process.kill(), i.e. an unbounded wait() in __exit__ during startup probing. It now uses a plain Popen, stdin=DEVNULL, and _kill_process(). The sudo-prompt detection still works: with DEVNULL, sudo writes "a terminal is required to read the password" to stderr, which the existing regex matches — just immediately instead of after the 3 s deadline.

4. The skip is now a real skip. _log_values() returns bool, and get_details() returns {} when it's False instead of re-reading a log file that still holds the previous measure. Also {timeout:g} so the warning reads "7 seconds", not "7.0 seconds".

I deliberately left the returncode != 0 path returning True (warn, then read the file) — that's today's behaviour, and flipping it would silently turn real readings into zeros for anyone whose powermetrics exits non-zero but still writes data. Say the word if you'd rather that path also skip.

**One thing this does not fix**, and should be treat as a separate issue: when `get_details()` returns `{}`, `AppleSiliconChip._get_power()` records `0 W` for that interval rather than omitting the sample. There's no "no measurement" concept at that layer, so a genuine skip needs a change in `hardware.py`.

benoit-cty (approved):
Thanks !

PR #1408, opened 2026-09-06
fix: make tracker.stop() idempotent

### LINKED ISSUE #1307: tracker.stop() is not idempotent: a second call writes a duplicate emissions row

sohammishra864-wq:
Body:

## Description
Make `tracker.stop()` idempotent so calling it multiple times does not produce duplicate emissions rows or trigger duplicate handler cleanup.

## Related Issue
Closes #1307

## Motivation and Context
When using the tracker as a context manager and also calling `stop()` explicitly to capture the return value, `__exit__` calls `stop()` a second time. The second call ran the full measurement-persist-exit cycle again, writing a duplicate CSV row and double-calling every output handler's `exit()

The root cause: `_start_time` was never cleared after stopping, so the existing guard never triggered. Additionally, the lock release ran before the guard, causing a redundant
release on re-entry.

## How Has This Been Tested?
- Verified calling `stop()` twice returns the same emissions value without writing a
duplicate row
- Verified `with` block + explicit `stop()` produces a single CSV row

## Types of changes
- [ ] Breaking change (fix or ting functionality to change)

## AI Usage Disclosure
- [ ]  AI-vibecoded
- [ ]  AI-generated
- [ ]  AI-assisted
- [x]  No AI used

## Checklist:
- [x] I have read the **docs/hent.

### DIFF
--- a/codecarbon/emissions_tracker.py
+++ b/codecarbon/emissions_tracker.py
@@ -924,12 +924,12 @@ class BaseEmissionsTracker(ABC):
                 "Another instance of codecarbon is already running. Exiting."
             )
             return
-        if not self._allow_multiple_runs:
-            # Release the lock
-            self._lock.release()
         if self._start_time is None:
-            logger.error("You first need to start the tracker.")
-            return None
+            logger.warning("Tracker already stopped or never started.")
+            return getattr(self, "final_emissions", None)
+
+        if not self._allow_multiple_runs:
+            self._lock.release()
 
         if self._scheduler:
             self._scheduler.stop()
@@ -961,6 +961,8 @@ class BaseEmissionsTracker(ABC):
 
         for handler in self._output_handlers:
             handler.exit()
+
+        self._start_time = None
 
         return emissions_data.emissions
 


SaboniAmine (approved):
Thanks for your contribution!

